PR #6, opened 2019-04-25
Add top-level README, and bazel and kaniko Tasks

ImJasonH:
This migrates some initial contents from https://github.com/knative/build-templates

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,0 +1,101 @@
+# Tekton Task Catalog
+
+This repository contains a catalog of `Task` resoruces, which are designed to be
+reusable in many pipelines.
+
+Each `Task` is provided in a separate directory along with a README.md and a
+Kubernetes manifest, so you can choose which `Task`s to install on your
+cluster.
+
+## `Task` Kinds
+
+There are two kinds of `Task`s:
+
+ 1. `ClusterTask` with a Cluster scope, which can be installed by a cluster
+    operator and made available to users in all namespaces
+ 2. `Task` with a Namespace scope, which is designed to be installed and used
+    only within that namespace.
+
+`Task`s in this repo are namespace-scoped `Task`s, but can be installed as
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -0,0 +1,101 @@
+# Tekton Task Catalog
```
dlorenc: I was thinking this would eventually contain both tasks and pipelines. WDYT?

kaniko/kaniko.yaml
```diff
@@ -0,0 +1,46 @@
+apiVersion: tekton.dev/v1alpha1
```
dlorenc: Starting to think through the testing story here, WDYT about also including a valid TaskRun? You can inline the resources. At a minimum (you don't have to do it in this PR), we could create all of the supplied Runs and make sure they succeed.

ImJasonH: I had thought about it. That gets tricky when the Task requires auth to successfully push the image. It's the same reason the build-templates repo never included actual running builds... 😒

README.md
ImJasonH: Yeah definitely. I only called this the "Task Catalog" because it only has Tasks for right now.

kaniko/kaniko.yaml
dlorenc: Hmm, maybe we'll have to come up with some conventions on what auth we'll provide as part of the test. Thanks for the explanation!

README.md
bobcatfish: and resources one day too (once we decouple them from the controller... WATCH THIS SPACE...)

README.md
```diff
@@ -0,0 +1,101 @@
...
+
+This repository contains a catalog of `Task` resoruces, which are designed to be
```
vdemeester: `s/resoruces/resources`

PR #8, opened 2019-04-26
Add examples for using oc

piyush-garg:
This will add examples of using oc binary in task
It will add a dockerfile of oc binary and two
examples

1. Using oc on same cluster using serviceAccount
2. Using oc on different cluster using clusterResource

Thanks

### DIFF
--- a/openshift-cli/README.MD
+++ b/openshift-cli/README.MD
@@ -1,14 +1,14 @@
 ## Openshift-Client
 
-`oc` is a tool which is used to interact with OpenShift from the command line. The following section provides two examples to interact with Openshift Cluster from Tekton Pipelines.
+`oc` is a tool which is used to interact with OpenShift from the command line. The following section provides two examples to interact with OpenShift Cluster from Tekton Pipelines.
 
-Dockerfile of the oc binary has also been added and the image is available [here]( https://quay.io/repository/openshift-pipeline/openshift-cli).
+The `Dockerfile` of the `oc` binary has been added and the image is available [here]( https://quay.io/repository/openshift-pipeline/openshift-cli).
 
 ## Examples:
 
-- Using oc on the same cluster using serviceAccount
+- Using `oc` on the same cluster using serviceAccount
 
-- Using oc on a different cluster using clusterResource
+- Using `oc` on a different cluster using clusterResource
 
 ## Prerequisite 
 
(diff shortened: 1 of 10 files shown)


openshift-cli/taskUsingClusterResource/octaskresource.yaml
```diff
@@ -0,0 +1,30 @@
...
+  inputs:
+    resources:
+      - name: clusterdetails
+        type: cluster
+  steps:
+    - name: version
+      image: quay.io/openshift-pipeline/openshift-cli
+      command:
+        - /usr/local/bin/oc
+      args: ["--kubeconfig", "/workspace/${inputs.resources.clusterdetails.name}/kubeconfig", "--context", "${inputs.resources.clusterdetails.name}", "version"]
```
vdemeester: With `v0.3.0` release of `tektoncd/pipeline`, you can now "factor" those into `containerTemplate` :angel:

vdemeester: (if you use `KUBECONFIG` env for example, with `args`, it's a bit trickier :sweat_smile: )

ImJasonH:
Can you add a README to this directory describing what it does?

In general, I'm not sure where we want to draw the line on "samples" in this repo, versus the repo being a repository for usable Tasks/Pipelines/Resources, out-of-the-box.

The bazel/kaniko/buildpacks Tasks are intended to be used directly, but I understand we also want a sample Pipeline for build-test-scan and staged rollout, which is probably unlikely to be directly usable without modification.

openshift-cli/taskUsingClusterResource/octaskresource.yaml
piyush-garg: Tried, not working properly, created an issue on pipeline https://github.com/tektoncd/pipeline/issues/811

openshift-cli/README.MD
```diff
@@ -0,0 +1,91 @@
+## Openshift-Client
+
+`oc` is a tool which is used to interact with OpenShift from the command line. The following section provides two examples to interact with Openshift Cluster from Tekton Pipelines.
+
+Dockerfile of the oc binary has also been added and the image is available [here]( https://quay.io/repository/openshift-pipeline/openshift-cli).
```
vdemeester: s/Dockerfile/The `Dockerfile`/g

PR #9, opened 2019-04-26
Add Task for Cloud Native Buildpacks

ImJasonH:
This migrates the build template at https://github.com/knative/build-templates/blob/master/buildpacks/cnb.yaml

Notable changes:
* Passes `-app=.` with `workingdir: /workspace/source` instead of `-app=/workspace` since Tekton puts source in the `./source` subdir
* Uses `heroku/buildpacks` image, since it seemed to give better output while I was debugging; if buildpacks folks prefer the other image let me know, or we can change it in a future PR.

I tested this running on source at https://github.com/buildpack/sample-java-app and the task was able to detect it as Java and build it.

/cc @matthewmcnew @sclevine @ekcasey

### DIFF
--- a/buildpacks/buildpacks-v3.yaml
+++ b/buildpacks/buildpacks-v3.yaml
@@ -1,0 +1,102 @@
+apiVersion: tekton.dev/v1alpha1
+kind: Task
+metadata:
+  name: buildpacks-v3
+spec:
+  inputs:
+    params:
+    - name: IMAGE
+      description: The image you wish to create. For example, "repo/example", or "example.com/repo/image".
+    - name: RUN_IMAGE
+      description: The run image buildpacks will use as the base for IMAGE.
+      default: packs/run:rc
+    - name: BUILDER_IMAGE
+      description: The builder image (must include v3 lifecycle and compatible buildpacks).
+      default: heroku/buildpacks
+    - name: USE_CRED_HELPERS
+      description: Use Docker credential helpers for Google's GCR, Amazon's ECR, or Microsoft's ACR.
+      default: 'true'
+    - name: CACHE
+      description: The name of the persistent app cache volume
+      default: empty-dir
+    - name: USER_ID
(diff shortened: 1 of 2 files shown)


buildpacks/buildpacks-v3.yaml
```diff
@@ -0,0 +1,102 @@
...
+  inputs:
+    params:
+    - name: IMAGE
+      description: The image you wish to create. For example, "repo/example", or "example.com/repo/image".
+    - name: RUN_IMAGE
+      description: The run image buildpacks will use as the base for IMAGE.
+      default: packs/run:rc
+    - name: BUILDER_IMAGE
+      description: The builder image (must include v3 lifecycle and compatible buildpacks).
+      default: heroku/buildpacks
```
matthewmcnew:
```suggestion
      default: packs/samples:v3alpha2
```

Align the task yaml with readme or remove the default altogether?

The most recent release of the [pack cli](https://github.com/buildpack/pack) does not include a default builder.

buildpacks/buildpacks-v3.yaml
```diff
@@ -0,0 +1,102 @@
...
+metadata:
+  name: buildpacks-v3
+spec:
+  inputs:
+    params:
+    - name: IMAGE
+      description: The image you wish to create. For example, "repo/example", or "example.com/repo/image".
+    - name: RUN_IMAGE
+      description: The run image buildpacks will use as the base for IMAGE.
+      default: packs/run:rc
```
matthewmcnew:
```suggestion
      default: packs/run:v3alpha2
```

The packs/run:rc is not a production ready run image. 
Align with the README or remove the default altogether?

ImJasonH:
Thanks for the review @matthewmcnew !

I'd like to keep as many of the defaults in place as possible, so that users don't have to dive deep into buildpacks and understand all the options to be able to use the Task.

PR #11, opened 2019-04-30
Add buildah Task

ImJasonH:
# Changes

Adds a Task that builds an image described by a Dockerfile using Buildah. This is migrated from https://github.com/knative/build-templates/tree/master/buildah

# Submitter Checklist

review them:

for more details._

@vbatts @vdemeester


buildah/buildah.yaml
```diff
@@ -0,0 +1,42 @@
+apiVersion: tekton.dev/v1alpha1
+kind: Task
+metadata:
+  name: buildah
+spec:
+  inputs:
+    params:
+    - name: BUILDER_IMAGE
```
vdemeester: So we have now a `buildah` image that is continuously update here : https://quay.io/repository/openshift-pipeline/buildah — coming from upstream images recipes (https://github.com/containers/buildah/tree/master/buildahimage). We should be able to use this one as default (and removing the need to have a dockerfile in there)

buildah/buildah.yaml
```diff
@@ -0,0 +1,42 @@
...
+    resources:
+    - name: source
+      type: git
+
+  steps:
+  - name: build
+    image: ${inputs.params.BUILDER_IMAGE}
+    workingdir: /workspace/source
+    command: ['buildah', 'bud', '--tls-verify=${inputs.params.TLSVERIFY}', '--layers', '-f', '${inputs.params.DOCKERFILE}', '-t', '${inputs.params.IMAGE}', '.']
+    volumeMounts:
```
vdemeester: We may need `privileged` (at least on openshift)

```
    securityContext:
      privileged: true
```

cc @vbatts

buildah/buildah.yaml
ImJasonH: Fantastic!

buildah/README.md
```diff
@@ -0,0 +1,63 @@
...
+
+```
+kubectl apply -f https://raw.githubusercontent.com/tektoncd/catalog/master/buildah/buildah.yaml
+```
+
+## Inputs
+
+### Parameters
+
+* **IMAGE**: The Docker image name to apply to the newly built image.
```
dlorenc: Do you think it would make sense to switch this to an Output of type image instead of an input parameter? See https://github.com/tektoncd/catalog/pull/15 as well.

ImJasonH: Sure. I wasn't exactly sure how output images worked while they don't yet collect the pushed image digest (https://github.com/tektoncd/pipeline/pull/721), or whether I should use info about the Task's output image in the configuration of the steps. I guess that makes sense though, I'll give it a shot.

PR #14, opened 2019-05-02
Add more documentation on development/contributing new submissions.

dlorenc:
# Changes

The goal of this commit is to add a set of "requirements" and guidelines
to serve as a starting point for maintainers reviewing new submissions.
I expect this will evolve over time as we get more experience maintaining
the catalog.

# Submitter Checklist

review them:

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -1,3 +1,36 @@
 # Developing additions to the Catalog
 
-TBD, for now see [CONTRIBUTING.md](CONTRIBUTING.md).
+The Catalog repository is intended to serve as a location where users can find
+`Task`s and `Pipeline`s that are maintained, useful and follow established
+best practices.
+
+When reviewing PRs that add new `Task`s or `Pipeline`s, maintainers will follow
+the following guidelines:
+
+* Submissions should be useful in real-world applications.
+While this repository is meant to be educational, its primary goal is to serve
+as a place users can find, share and discover useful components.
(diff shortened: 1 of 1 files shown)


DEVELOPMENT.md
```diff
@@ -1,3 +1,35 @@
...
+The Catalog repository is intended to serve as a location where users can find
+`Task`s and `Pipeline`s that are maintained, useful and follow established
+best practices.
+
+When reviewing PRs that add new `Task`s or `Pipeline`s, maintainers will follow
+the following guidelines:
+
+* Submissions should be independently usable.
+While this repository is meant to be educational, its primary goal is to serve
+as a place users can find, share and discover useful components.
```
kimsterv: I think an example here would be helpful.

DEVELOPMENT.md
```diff
@@ -1,3 +1,35 @@
...
+* Submissions should follow established best practices.
+Tekton is still young so this is going to be a shifting goalpost, but here are
+some examples:
+    * `Task`s should expose parameters and declare input/output resources, and
+    document them.
+    * Submissions should be as *portable* as possible.
+    They should not be hardcoded to specific repositories, clusters,
+    environments etc.
+    * Images should either be pinned by digest or point to tags that
```
sbwsg: s/tags that documented/tags with documented/

dlorenc: Thanks!

DEVELOPMENT.md
dlorenc: How's this?

PR #19, opened 2019-05-08
Make the oc task reusable

piyush-garg:
This will add a script to Dockerfile which will
receive command as a string and then parse them
to execute with OC

This way we can make the task reusable

Changes to use the Ubi image as the base image

Update both the examples to use reusables task

Update README


openshift-cli/Dockerfile
```diff
@@ -1,13 +1,16 @@
-FROM docker.io/centos:7
+FROM registry.access.redhat.com/ubi8-dev-preview/ubi:latest
```
siamaksade: The UBI 8 GA should be used instead: `registry.access.redhat.com/ubi8/ubi`

openshift-cli/Dockerfile
```diff
@@ -1,13 +1,16 @@
...
 
 RUN cd /tmp \
   && yum -y update \
   && yum install -y wget \
   && wget https://mirror.openshift.com/pub/openshift-v4/clients/oc/latest/linux/oc.tar.gz \
   && tar -xvzf oc.tar.gz \
-  && mv oc /usr/local/bin/ \
+  && mv oc oc-origin \
```
vdemeester: This could be done in one shot : `mv oc /usr/local/bin/oc-origin` :wink:

siamaksade:
Is there a reason for using `openshift-cli` for the task name and directory but `openshift-client` in the YAML file names?

piyush-garg:
@siamaksade There was no reason for that, I have updated all to `openshift-client` for consistent naming

PR #21, opened 2019-05-09
Remove bazel from the catalog.

dlorenc:
# Changes

I spent a long time trying to get bazel to work with auth, but was unable
to. I think we should remove it for now until we have a test harness and
can verify it actaully works.

# Submitter Checklist

review them:


zhouhaibing089:
Thanks for the amazing work of maintaining this awesome catalog!

When I think of bazel as a Task, it isn't just about image pushing at all. It's more about specifying the targets for `build`/`test`/`run`. And image pushing just happened to be one target for `run`. It is common to just run a Task for build or test as well as running complicated jobs.

May I ask what's the challenge of configuring auth for bazel so we can have a followup and solve the issue, and also can we abstract the bazel task to accept parameters which specify the bazel targets?

```yaml
kind: TaskRun
metadata:
  name: bazel
spec:
  taskRef:
    name: bazel
  inputs:
    params:
    - name: BUILD_TARGET
      value: //path/to/build
    - name: TEST_TARGET
      value: //path/to/test
    - name: RUN_TARGET
      value: //path/to/run
    resources:
    - name: source
      resourceSpec:
        type: git
        params:
        - name: url
          value: https://github.com/my-user/my-repo
```

PR #22, opened 2019-05-10
kaniko: set DOCKER_CONFIG

AkihiroSuda:
Discussed in https://github.com/tektoncd/pipeline/pull/706
Fix https://github.com/tektoncd/pipeline/issues/848

Signed-off-by: Akihiro Suda <akihiro.suda.cz@hco.ntt.co.jp>

# Changes

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/kaniko/kaniko.yaml
+++ b/kaniko/kaniko.yaml
@@ -22,6 +22,11 @@ spec:
   - name: build-and-push
     workingdir: /workspace/source
     image: gcr.io/kaniko-project/executor
+    # specifying DOCKER_CONFIG is required to allow kaniko to detect docker credential
+    # https://github.com/tektoncd/pipeline/pull/706
+    env:
+    - name: DOCKER_CONFIG
+      value: /builder/home/.docker
     command:
     - /kaniko/executor
     - --dockerfile=${inputs.params.DOCKERFILE}


AkihiroSuda:
I confirmed this works with `gcr.io`

vdemeester:
@bobcatfish I'm still not entirely sure why this is needed, but if it fixes most people, let's get it merge as is and figure things out later :sweat_smile:

dlorenc:
> @bobcatfish I'm still not entirely sure why this is needed, but if it fixes most people, let's get it merge as is and figure things out later 😅

I remember debugging this for knative/build here: https://github.com/knative/build-templates/pull/87

This looks like the right fix to me!

vtereso:
Perhaps someone can make better sense of this than myself. The core of what is going "wrong" should be the following: https://github.com/google/go-containerregistry/blob/master/pkg/authn/keychain.go#L41, which is used by Kaniko here https://github.com/GoogleContainerTools/kaniko/blob/master/pkg/executor/push.go#L110. I have run some sanity tests and by adding steps before executing Kaniko in a build task to check the file system and I get ```/builder
/builder/home
/builder/home/.docker
/builder/home/.docker/config.json``` and additionally, I have run another step that prints $HOME that would be referenced in the above authn and I get `/builder/home`, which would be exactly the same as the ```env:
    - name: DOCKER_CONFIG
      value: /builder/home/.docker``` proposed fix?

vtereso:
Not familiar with the wrapping container logic of Tekton, but https://github.com/GoogleContainerTools/kaniko/blob/master/deploy/Dockerfile#L36 seems to make sense.

dlorenc:
Yeah, so to summarize and make sure I understand the issue:

1.  github.com/google/go-containerregistry respects the env var DOCKER_CONFIG, then $HOME/.docker/config.json in that order https://github.com/google/go-containerregistry/blob/master/pkg/authn/keychain.go#L41
1. kaniko sets DOCKER_CONFIG to /kaniko/.docker https://github.com/GoogleContainerTools/kaniko/blob/master/deploy/Dockerfile#L40
1. tekton/pipeline overrides $HOME to /builder/home, and injects credentials at /builder/home/.docker.

Because of 2), we always check /kaniko/.docker instead of the $HOME version.

vtereso:
@dlorenc Yeah good point. I wasn't looking for it so I didn't realize `$DOCKER_CONFIG` is also specified by Kaniko and since the first (`$DOCKER_CONFIG`) config directory will never fail, it must be the correct location or will yield invalid credentials. I was under the impression that `$HOME` was had been overridden by the Kaniko container, but Kubernetes env should have priority? I am not familiar with precedence for this.

PR #23, opened 2019-05-13
Add the required boilerplate to run presubmit tests in prow.

dlorenc:
# Changes

This adds a single dummy test (required to make dep fetch the dependencies),
and then the Gopkg.toml configuration so we grab the presubmit helper scripts.

This will eventually use the code in plumbing instead of knative, but that isn't
ready yet.

# Submitter Checklist

review them:


dlorenc:
Ref https://github.com/tektoncd/catalog/pull/23

PR #28, opened 2019-05-17
Add buildah dockerfiles

chmouel:
# Changes

Add DockerFile for Buildah task

# Submitter Checklist

review them:

- [ X ] Includes [docs](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if user facing)
- [ X ] Commit messages follow [commit message best practices](https://github.com/tektoncd/community/blob/master/standards.md#commit-messages)

for more details._

### DIFF
--- a/buildah/Dockerfiles/Dockerfile
+++ b/buildah/Dockerfiles/Dockerfile
@@ -1,0 +1,8 @@
+FROM centos:7
+
+RUN yum update -y && \
+    yum install --setopt=tsflags=nodocs -y \
+    buildah && \
+    yum clean all && rm -rf /var/cache/yum/
+
+CMD /usr/bin/buildah
--- a/buildah/Dockerfiles/Makefile
+++ b/buildah/Dockerfiles/Makefile
@@ -1,0 +1,6 @@
+REGISTRY=quay.io/openshift-pipeline/buildah
+
+all: build
+
+build:
+	docker build . -t $(REGISTRY) && docker push $(REGISTRY)
--- a/buildah/Dockerfiles/README.md
+++ b/buildah/Dockerfiles/README.md
@@ -1,0 +1,1 @@
+Simple Dockerfile to install buildah 


buildah/Dockerfiles/Dockerfile
```diff
@@ -0,0 +1,18 @@
+FROM centos:7
```
siamaksade: Should be based on UBI 8

chmouel: i can't find buildah and podman package on ubi 8  :(

buildah/Dockerfiles/Makefile
```diff
@@ -0,0 +1,6 @@
+REGISTRY=quay.io/openshift-pipeline/buildah
```
dlorenc: We won't have permission to push here right? Maybe we should figure out a home for all the images in the catalog.

chmouel: @vdemeester what do you think ?

vdemeester: agreeing with @dlorenc, in the future we may want to have some common place where the image defined here in the catalog are pushed.

dlorenc:
Ref https://github.com/tektoncd/catalog/issues/29

vbatts:
is the goal of the catalog to have yaml, and possibly Dockerfiles, ready for folks to use?
I feel this PR is not quite re-using the Dockerfile that upstream buildah provides, nor the fresh builds of it that live at quay.io/openshift-pipeline/buildah.
I feel that it might only need to be the yaml to make these things available in tekton as resources.

dlorenc:
> I feel that it might only need to be the yaml to make these things available in tekton as resources.

Yeah I think we want to prefer to use existing images, where possible. If for some reason Tekton requires a differently configured image we could consider hosting the Dockerfile here, but I don't think that should be a common case.

chmouel:
The idea at first was to have the s2i binary as well, altought we don't need to have an image containing all so let's use `quay.io/openshift-pipeline/buildah`

PR #31, opened 2019-05-22
Add s2i task 

khrm:
# Changes
Adds task that builds an image described by a using s2i.
Also provides dockerfile for generating s2i image needed for building.
Taskrun in ./examples is also 

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/s2i/examples/nodejs-s2i.yaml
+++ b/s2i/examples/nodejs-s2i.yaml
@@ -1,0 +1,25 @@
+apiVersion: tekton.dev/v1alpha1
+kind: TaskRun
+metadata:
+  name: s2i-nodejs-taskrun
+spec:
+  # Use service account with git and image repo credentials
+  serviceAccount: saacount
+  taskRef:
+    name: s2i
+  inputs:
+    resources:
+    - name: git-source
+      resourceSpec:
+        type: git
+        params:
+        - name: url
+          value: https://github.com/sclorg/nodejs-ex
+  outputs:
+    resources:
+    - name: image
+      resourceSpec:
+        type: image
+        params:
+        - name: url
+          value: quay.io/openshift-pipeline/nodejs-s2i-ex
--- a/s2i/s2i.yaml
+++ b/s2i/s2i.yaml
@@ -1,0 +1,62 @@
+apiVersion: tekton.dev/v1alpha1
+kind: Task
+metadata:
+  name: s2i
+spec:
+  inputs:
+    resources:
+     - name: git-source
+       type: git
+    params:
+    - name: BUILDER_IMAGE
+      description: The location of the builder image.
+      default: quay.io/openshift-pipeline/s2i
+    - name: S2I_IMAGE
+      description: The location of the s2i builder image.
+      default: docker.io/centos/nodejs-8-centos7
+    - name: PATH_CONTEXT
+      description: The location of the path to run s2i from.
+      default: .
+    - name: TLSVERIFY
+      description: Verify the TLS on the registry endpoint (for push/pull to a non-TLS registry)
+      default: "true"
+
+  outputs:
+    resources:
+    - name: image
+      type: image
+
+  steps:
+  - name: generate
+    image: ${inputs.params.BUILDER_IMAGE}
+    workingdir: /workspace/git-source
+    command: ['s2i', 'build', '${inputs.params.PATH_CONTEXT}', '${inputs.params.S2I_IMAGE}', '--as-dockerfile', '/gen-source/Dockerfile.gen']
+    volumeMounts:
+     - name: gen-source
+       mountPath: /gen-source
+  - name: build
+    image: quay.io/openshift-pipeline/buildah
+    workingdir: /gen-source
+    command: ['buildah', 'bud', '--tls-verify=${inputs.params.TLSVERIFY}', '--layers', '-f', '/gen-source/Dockerfile.gen', '-t', '${outputs.resources.image.url}', '.']
+    volumeMounts:
+    - name: varlibcontainers
+      mountPath: /var/lib/containers
+    - name: gen-source
+      mountPath: /gen-source
+    securityContext:
+      privileged: true
+
+  - name: push
+    image: quay.io/openshift-pipeline/buildah
+    command: ['buildah', 'push', '--tls-verify=${inputs.params.TLSVERIFY}', '${outputs.resources.image.url}', 'docker://${outputs.resources.image.url}']
+    volumeMounts:
+    - name: varlibcontainers
+      mountPath: /var/lib/containers
+    securityContext:
+      privileged: true
+
+  volumes:
+  - name: varlibcontainers
+    emptyDir: {}
+  - name: gen-source
+    emptyDir: {}
--- a/s2i/Dockerfile
+++ b/s2i/Dockerfile
@@ -1,0 +1,9 @@
+FROM centos:7
+
+# Get latest S2I release  from github with some curl+rest+python magic (which is provided by default 
+# in the image so wedon't have to install extra packages)
+RUN mkdir -p /usr/local/bin && \
+    curl -L $(curl -L -s "https://api.github.com/repos/openshift/source-to-image/releases/latest"| python -c "import sys, json;x=json.load(sys.stdin);print([ r['browser_download_url'] for r in x['assets'] if 'linux-amd64' in r['name']][0])") -o /tmp/s2i.tgz && \
+    tar xz -f/tmp/s2i.tgz -C /usr/local/bin/
+
+ENTRYPOINT ["/usr/local/bin/s2i"]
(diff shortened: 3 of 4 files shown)


s2i/examples/nodejs-s2i.yaml
```diff
@@ -0,0 +1,24 @@
+apiVersion: tekton.dev/v1alpha1
+kind: TaskRun
+metadata:
+  name: s2i-nodejs-1
+spec:
+  serviceAccount: kbaig
```
hrishin: could we use a better service account name here? :sweat_smile: 

Would it be nice if we can put a sample service account file as well?

s2i/s2i.yaml
```diff
@@ -0,0 +1,45 @@
...
+  outputs:
+    resources:
+    - name: image
+      type: image
+
+  steps:
+  - name: generate
+    image: ${inputs.params.BUILDER_IMAGE}
+    workingdir: /workspace/git-source
+    command: ['s2i', 'build', '.', '${inputs.params.S2I_IMAGE}', '--as-dockerfile', '/gen-source/Dockerfile.gen']
```
hrishin: should we accept the thitd parameter as an `input.parameter`? where the default parameter value would be `.` (current dir)?

So the user could use this task without making a change in task definition?

openshift-s2i/Dockerfile
```diff
@@ -0,0 +1,11 @@
+From golang:alpine as builder
```
vdemeester: We should use `centos:7` here for now

openshift-s2i/s2i.yaml
```diff
@@ -0,0 +1,48 @@
...
+    resources:
+     - name: git-source
+       type: git
+    params:
+    - name: BUILDER_IMAGE
+      description: The location of the builder image.
+      default: quay.io/openshift-pipeline/kbaig_s2i
+    - name: S2I_IMAGE
+      description: The location of the s2i builder image.
+      default: docker.io/centos/nodejs-8-centos7
```
vdemeester: Let's provide a `s2i` tasks that has no default, and additional tasks (`s2i-nodejs`, `s2i-python`, …) that uses those images (`centos/nodejs-…`) — That will allow better configuration for each (as an example : I could want nodejs-10 with centos7, or nodejs-8 with something else, so for node, we could have a version parameter or something like that)

openshift-s2i/s2i.yaml
```diff
@@ -0,0 +1,48 @@
...
+  name: s2i
+spec:
+  inputs:
+    resources:
+     - name: git-source
+       type: git
+    params:
+    - name: BUILDER_IMAGE
+      description: The location of the builder image.
+      default: quay.io/openshift-pipeline/kbaig_s2i
```
vdemeester: This needs to be `quay.io/openshift-pipeline/s2i`.

openshift-s2i/s2i.yaml
```diff
@@ -0,0 +1,48 @@
...
+  steps:
+  - name: generate
+    image: ${inputs.params.BUILDER_IMAGE}
+    workingdir: /workspace/git-source
+    command: ['s2i', 'build', '${inputs.params.PATH_CONTEXT}', '${inputs.params.S2I_IMAGE}', '--as-dockerfile', '/gen-source/Dockerfile.gen']
+    volumeMounts:
+     - name: gen-source
+       mountPath: /gen-source
+  - name: build-and-push
+    image: gcr.io/kaniko-project/executor
```
vdemeester: Let's use `buildah` here

piyush-garg:
@khrm I see two taskruns and both almost the same. Any reason for that. Am I missing something?

nikhil-thomas:
:+1:

openshift-s2i/examples/taskrun-nodejs.yaml
```diff
@@ -0,0 +1,24 @@
+apiVersion: tekton.dev/v1alpha1
+kind: TaskRun
+metadata:
+  name: s2i-nodejs-1
```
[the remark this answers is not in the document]
khrm: This is a wrong file which got pushed. Deleting it.

openshift-s2i/examples/taskrun-nodejs.yaml
```diff
@@ -0,0 +1,24 @@
...
+spec:
+  serviceAccount: kbaig
```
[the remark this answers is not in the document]
khrm: This is a wrong file which got pushed. Deleting it.

openshift-s2i/examples/nodejs-s2i.yaml
```diff
@@ -0,0 +1,25 @@
+apiVersion: tekton.dev/v1alpha1
+kind: TaskRun
+metadata:
+  name: s2i-nodejs-1
```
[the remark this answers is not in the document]
khrm: Rename it to s2i-nodejs-taskrun

openshift-s2i/README.md
```diff
@@ -0,0 +1,35 @@
...
+```
+kubectl apply -f https://raw.githubusercontent.com/tektoncd/catalog/master/s2i/s2i-nodejs.yaml
+```
+
+## Inputs
+
+### Parameters
+
+* **BUILDER_IMAGE:**: The name of the image containing the s2i command. See
+  note below.  (_default:_ docker.io/kbaig/s2i)
```
hrishin: `docker.io/kbaig/s2i` -> `quay.io/redhat-pipelines/s2i`?

openshift-s2i/s2i.yaml
khrm: OK. I didn't use buildah because it requires privilege.

openshift-s2i/README.md
vdemeester: @hrishin it's `quay.io/openshift-pipelines/s2i` :wink:

openshift-s2i/examples/nodejs-s2i.yaml
```diff
@@ -0,0 +1,25 @@
...
+        - name: url
+          value: https://github.com/sclorg/nodejs-ex
+  outputs:
+    resources:
+    - name: image
+      resourceSpec:
+        type: image
+        params:
+        - name: url
+          value: docker.io/kbaig/nodejs-ex
```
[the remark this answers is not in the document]
vdemeester: question: isn't there an official s2i nodejs image somewhere ?

openshift-s2i/s2i.yaml
khrm: Changed to buildah

openshift-s2i/examples/nodejs-s2i.yaml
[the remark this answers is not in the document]
khrm: I didn't find any. I have changed to quay.io/openshift-pipeline for example.

openshift-s2i/Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM centos:7
+
+# Get latest S2I release  from github with some curl+rest+python magic (which is provided by default 
+# in the image so wedon't have to install extra packages)
+RUN mkdir -p /usr/local/bin && \
+    curl -L $(curl -L -s "https://api.github.com/repos/openshift/source-to-image/releases/latest"| python -c "import sys, json;x=json.load(sys.stdin);print([ r['browser_download_url'] for r in x['assets'] if 'linux-amd64' in r['name']][0])") -o /tmp/s2i.tgz && \
+    tar xz -f/tmp/s2i.tgz -C /usr/local/bin/
+
+CMD /usr/local/bin/s2i
```
vdemeester: nit: a bit late but, shouldn't this be `ENTRYPOINT` even ? (that way `docker run s2i --help` would work the same as `s2i` does)

openshift-s2i/s2i.yaml
khrm: Will send another pr with ruby, python and nodejs tasks.

khrm:
@vdemeester Done.

openshift-s2i/Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM centos:7
+
+# Get latest S2I release  from github with some curl+rest+python magic (which is provided by default 
+# in the image so wedon't have to install extra packages)
+RUN mkdir -p /usr/local/bin && \
+    curl -L $(curl -L -s "https://api.github.com/repos/openshift/source-to-image/releases/latest"| python -c "import sys, json;x=json.load(sys.stdin);print([ r['browser_download_url'] for r in x['assets'] if 'linux-amd64' in r['name']][0])") -o /tmp/s2i.tgz && \
+    tar xz -f/tmp/s2i.tgz -C /usr/local/bin/
+
+ENTRYPOINT /usr/local/bin/s2i
```
vdemeester: ah needs to be the json syntax (otherwise the entrypoint is `/bin/sh -c /usr/local/bin/s2i`)

```
ENTRYPOINT ["/usr/local/bin/s2i"]
```

khrm: Done.

khrm:
@vdemeester Done.

PR #32, opened 2019-05-22
Add a specific version and context support for kaniko  

houshengbo:
We need to pin to specific version rather than go with the dynamic latest commit. The version v0.9.0 seems to be a stable one.

Add the parameter context support for kaniko.

# Changes

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/kaniko/kaniko.yaml
+++ b/kaniko/kaniko.yaml
@@ -8,7 +8,9 @@ spec:
     - name: DOCKERFILE
       description: Path to the Dockerfile to build.
       default: ./Dockerfile
-
+    - name: CONTEXT
+      description: The build context used by Kaniko.
+      default: ./
     resources:
     - name: source
       type: git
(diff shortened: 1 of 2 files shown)


houshengbo:
The same PR is submitted ti Tektoncd pipeline as well: https://github.com/tektoncd/pipeline/pull/878.

kaniko/kaniko.yaml
```diff
@@ -30,4 +32,5 @@ spec:
     command:
     - /kaniko/executor
     - --dockerfile=${inputs.params.DOCKERFILE}
+    - --context=/workspace/source/${inputs.params.CONTEXT} # The user does not need to care the workspace and the source.
```
bobcatfish: nice!! :D p.s. with [path templating](https://github.com/tektoncd/pipeline/blob/master/docs/tasks.md#templating) (i.e. in this case `${inputs.resouces.source.path}/${inputs.params.CONTEXT}`) the author of the Task wouldnt have to worry about it either :D

PR #38, opened 2019-05-29
Update s2i tasks documentation

vdemeester:
# Changes

- Small fix on the parameter name (and comments), `BUILDER_IMAGE` and
  `S2I_IMAGE` were inverted.
- Remove the `examples` folder and add an example inline.
- Add a non-exhaustive list of s2i builder images that can be used
  with this task.

cc @siamaksade @khrm 

# Submitter Checklist

review them:

for more details._


s2i/README.md
```diff
@@ -13,15 +13,18 @@ kubectl apply -f https://raw.githubusercontent.com/tektoncd/catalog/master/s2i/s
 
 ### Parameters
 
-* **BUILDER_IMAGE**: The name of the image containing the s2i command. See
+* **BUILDER_IMAGE**: The location of the s2i builder image.
```
siamaksade: Don't think this param is needed. It has to always be the same image

vdemeester: you mean the `S2I_IMAGE` image right (the one that has the `s2i` binary in) ?

s2i/README.md
```diff
@@ -1,27 +1,33 @@
-# S2I
+# Source-to-Image
 
-This Task builds source into a container image using Project Atomic's
-[source-to-image/s2i](https://github.com/openshift/source-to-image). 
+[Source-to-Image (S2I)](https://github.com/openshift/source-to-image)
+is a toolkit and workflow for building reproducible container images
+from source code. S2I produces images by injecting source code into a
```
chmouel: nice one!

s2i/s2i.yaml
```diff
@@ -28,9 +24,9 @@ spec:
 
   steps:
   - name: generate
-    image: ${inputs.params.BUILDER_IMAGE}
-    workingdir: /workspace/git-source
-    command: ['s2i', 'build', '${inputs.params.PATH_CONTEXT}', '${inputs.params.S2I_IMAGE}', '--as-dockerfile', '/gen-source/Dockerfile.gen']
+    image: quay.io/openshift-pipeline/s2i
```
chmouel: So we don't want to allow the user to override the s2i image?

vdemeester: Do you think we should do ? looking around catalog, sometimes we allow it, sometimes we don't, so I wasn't sure :joy_cat:

chmouel: If I was an advanced user using this I would probably would override it with my own image that would do fancy stuff, but i guess we are not to that point yet,

PR #41, opened 2019-06-04
new Task: curl-and-grep

olegchorny:
# Changes

new Task: curl-and-grep

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/curl-and-grep/curl-and-grep.yaml
+++ b/curl-and-grep/curl-and-grep.yaml
@@ -1,0 +1,30 @@
+apiVersion: tekton.dev/v1alpha1
+kind: Task
+metadata:
+  name: curl-and-grep
+spec:
+  inputs:
+    params:
+    - name: url
+      description: URL to be curl'ed.
+    - name: text
+      description: text to grep
+      default: "HTTP"
+    - name: buffer
+      description: file to save output
+      default: '/workspace/buffer.txt'
+  steps:
+  - name: grep
+    image: plerionio/curl
+    command: 
+    - 'curl'
+    - '-iL'
+    - ${inputs.params.url} 
+    - '-o' 
+    - ${inputs.params.buffer}
+  - name: curl
+    image: plerionio/curl
+    command:
+    - 'grep'
+    - ${inputs.params.text}
+    - ${inputs.params.buffer}
--- a/curl-and-grep/pipelinerun.yaml
+++ b/curl-and-grep/pipelinerun.yaml
@@ -1,0 +1,88 @@
+apiVersion: tekton.dev/v1alpha1
+kind: PipelineResource
+metadata:
+  name: productpage-git
+spec:
+  type: git
+  params:
+    - name: revision
+      value: master
+    - name: url
+      value: https://github.com/olegchorny/bookinfo-productpage
+---
+apiVersion: tekton.dev/v1alpha1
+kind: PipelineResource
+metadata:
+  name: productpage-image
+spec:
+  type: image
+  params:
+    - name: url
+      value: docker.io/ochorny/bookinfo-productpage
+---
+apiVersion: tekton.dev/v1alpha1
+kind: Pipeline
+metadata:
+  name: productpage-pipeline
+spec:
+  resources:
+    - name: source-repo
+      type: git
+    - name: web-image
+      type: image
+  tasks:
+    - name: build-kaniko
+      taskRef:
+        name: kaniko
+      resources:
+        inputs:
+          - name: source
+            resource: source-repo
+        outputs:
+          - name: image
+            resource: web-image
+    - name: deploy-web
+      taskRef:
+        name: deploy-using-kubectl
+      resources:
+        inputs:
+          - name: source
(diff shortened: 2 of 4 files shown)


curl-and-grep/curl-and-grep.yaml
```diff
@@ -0,0 +1,30 @@
...
+metadata:
+  name: curl-and-grep
+spec:
+  inputs:
+    params:
+    - name: url
+      description: URL to be curl'ed.
+    - name: text
+      description: text to grep
+      default: "HTTP"
```
vdemeester: Should this be a default or should we make this parameter required (aka no default value)

curl-and-grep/pipelinerun.yaml
```diff
@@ -0,0 +1,88 @@
+apiVersion: tekton.dev/v1alpha1
```
vdemeester: We tend to have those only in the `README.md` (especially as the CI will try to apply those, which we may want or not) 
cc @dlorenc

curl-and-grep/curl-and-grep.yaml
olegchorny: The motivation is to allow this task to be usable even in the case if there is no expected text in the body. So that, we can parse headers, where is "HTTP" should exist anyway. 
Initially, I was thinking to create separate tasks for curl and for grep, but one combined task could be better, to decrease the number of steps in pipelines and make them easier to use.

curl-and-grep/pipelinerun.yaml
olegchorny: got it, i will submit another PR is you find the idea as usable and implementation as viable.

vdemeester: let's wait for @dlorenc. You'll be able to amend your commit anyway, no need for a new PR :angel:

dlorenc: Yep, sorry for missing this. +1 on moving to the readme :)

olegchorny: thanks for the comment, I need some time to refresh my mind to make changes :)

curl-and-grep/curl-and-grep.yaml
```diff
@@ -0,0 +1,30 @@
...
+    - name: buffer
+      description: file to save output
+      default: '/workspace/buffer.txt'
+  steps:
+  - name: grep
```
afrittoli: step names are mixed up? the grep step does a curl, the curl step does a grep

curl-and-grep/curl-and-grep.yaml
```diff
@@ -0,0 +1,30 @@
...
+    image: plerionio/curl
```
afrittoli: This is alpine + curl - however do you know the image maintainer / if the image is maintained?

I wonder if we should add to the catalog repo machinery to include custom docker images and build them on a regular basis. This way we would be in control of the images we include in Tasks. 
For images that are well maintained like alpine, python, ubuntu or so it's fine to get them from dockerhub.

curl-and-grep/curl-and-grep.yaml
```diff
@@ -0,0 +1,30 @@
...
+  name: curl-and-grep
+spec:
+  inputs:
+    params:
+    - name: url
+      description: URL to be curl'ed.
+    - name: text
+      description: text to grep
+      default: "HTTP"
+    - name: buffer
```
afrittoli: What is the reason for the buffer? You could `curl -O- URL | grep TEXT`.

Since the image is the same there is not much overhead in using two steps, but even with the buffer this could be easily done as a single step `curl -O[BUFFER] [URL]; grep [TEXT] [BUFFER]`.

curl-and-grep/curl-and-grep.yaml
afrittoli: NIT: you may want to add a comment about this in the docs

curl-and-grep/curl-and-grep.yaml
olegchorny: you're right

curl-and-grep/curl-and-grep.yaml
olegchorny: Agreed, I need to find more trusted image.

curl-and-grep/curl-and-grep.yaml
olegchorny: good idea, let me give it a try.

curl-and-grep/curl-and-grep.yaml
olegchorny: @afrittoli It seems like before we start to create such custom tasks, we really need to think about the trusted source of any kind of tools. For example, in case of curl, it's really difficult to find the trusted image. So, I'm with you on the catalog for custom images under the Tekton team control.

sbwsg:
@olegchorny do you plan to continue work on this PR?

olegchorny:
@sbwsg Let me close it, since the context of this PR is little bit outdated.
Thanks.

PR #43, opened 2019-06-05
Initial golang tasks

vdemeester:
# Changes

Related to https://github.com/tektoncd/pipeline/pull/954 and https://github.com/tektoncd/pipeline/pull/949, this provides initial Tasks for Golang.

This adds initial golang tasks : build (`golang-build`), tests (`golang-tests`) and linting (`golangci-lint`, using golangci-lint).

Improvements (todos):
- [ ] Define `Output` Resources: either of type `git` *or* `storage` ?
- [ ] Once tektoncd/pipeline#207 is fixed, remove the *bashisms*.
- [ ] … your feedback here :angel: 

Closes tektoncd/pipeline#532
Closes tektoncd/pipeline#558

/cc @chmouel @bobcatfish 

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/golang/build.yaml
+++ b/golang/build.yaml
@@ -1,0 +1,44 @@
+apiVersion: tekton.dev/v1alpha1
+kind: Task
+metadata:
+  name: golang-build
+spec:
+  inputs:
+    params:
+    - name: package
+      description: base package to build in
+    - name: packages
+      description: "packages to build (default: ./cmd/...)"
+      default: "./cmd/..."
+    - name: version
+      description: golang version to use for builds
+      default: "1.12"
+    - name: flags
+      description: flags to use for the test command
+      default: -v
+    - name: GOOS
+      description: "running program's operating system target"
+      default: linux
+    - name: GOARCH
+      description: "running program's architecture target"
+      default: amd64
+    resources:
+    - name: source
+      type: git
+      targetPath: src/${inputs.params.package}
+  steps:
+  - name: build
+    image: golang:${inputs.params.version}
+    workingdir: /workspace/src/${inputs.params.package}
+    command:
+    - /bin/bash
+    args:
+    - -c
+    - "go build ${inputs.params.flags} ${inputs.params.packages}"
+    env:
+    - name: GOPATH
+      value: /workspace
+    - name: GOOS
+      value: "${inputs.params.GOOS}"
+    - name: GOARCH
+      value: "${inputs.params.GOARCH}"
--- a/golang/lint.yaml
+++ b/golang/lint.yaml
@@ -1,0 +1,41 @@
+apiVersion: tekton.dev/v1alpha1
+kind: Task
+metadata:
+  name: golangci-lint
+spec:
+  inputs:
+    params:
+    - name: package
+      description: base package (and its children) under validation
+    - name: flags
+      description: flags to use for the test command
+      default: --verbose
+    - name: version
+      default: golangci-lint version to use
+      default: "v1.16"
+    - name: GOOS
+      description: "operating system target (default: node operating system)"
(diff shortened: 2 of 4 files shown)


golang/lint.yaml
```diff
@@ -0,0 +1,41 @@
...
+    - name: GOARCH
+      description: "architecture target (default: node architecture)"
+      default:
+    resources:
+    - name: source
+      type: git
+      targetPath: src/${inputs.params.package}
+  steps:
+  - name: lint
+    image: golangci/golangci-lint:${inputs.params.version}
```
ImJasonH: Do you think there's any benefit in having the default include the `:`, which would allow users to request a version by digest (`version: @sha256:...`) or specify `""` to get `:latest` (less useful)

ImJasonH: I'm a bit uneasy about recommending this image without knowing much about its provenance. Who is responsible for building and maintaining this image? What prevents them from injecting malicious code?

golang/build.yaml
```diff
@@ -0,0 +1,44 @@
...
+kind: Task
+metadata:
+  name: golang-build
+spec:
+  inputs:
+    params:
+    - name: package
+      description: base package to build in
+    - name: packages
+      description: "packages to build (default: ./cmd/...)"
```
ImJasonH: I don't think we should document the default in the description, users should be able to get information about the default from (hopefully someday generated) documentation. Having to duplicate it makes it easy to have them fall out of sync.

vdemeester: oh yeah :sweat_smile:

golang/lint.yaml
vdemeester: > Who is responsible for building and maintaining this image? What prevents them from injecting malicious code?

@ImJasonH this is true for other tasks like `buildah`, using `quay.io/buildah/stable` (maintained by `containers/buildah` maintainers) or `quay.io/openshift-pipelines/buildah:stable` (maintained by my team). 

This is a good question, and I think it deserves an issue on this repository. Should all the task have the image(s) build recipes with them ? Is `tektoncd/catalog` responsible to build and publish those images ?

ImJasonH: Yeah I definitely think it's worth documenting at least, as a best practice for reusable Tasks.

In the absence of that, I'm much less worried about referencing an image that's built by some team that maintains the tool. Having a `buildah` image built by Red Hat is fine, having a `gcloud` or `golang` image built by Google is fine, having a `golang`(-adjacent) image built by some Dockerhub user who happens to have the name `golangci` feels...dicey. If it was `golang/lint`, I'd have no real concerns.

vdemeester: The `golangci/golangci-lint` image is maintained by the `golangci` team : https://github.com/golangci. In a sense, it's the same as `buildah` built by Red Hat :angel:

vdemeester:
(I've removed `wip` as I think we can merge as is more or less, create issues for follow-ups and iterate over it :angel:)

golang/lint.yaml
chmouel: in a sense this is the same OSS trust we have when adding a golang depdence in the code,

ImJasonH: I guess that's true. I didn't realize `golanci` was a real company with a real domain. 

This seems fine then. :man_shrugging:

golang/build.yaml
ImJasonH: Ping, I still see documented defaults in the latest commit

golang/lint.yaml
ImJasonH: ping, still see `version`'s default without the `:`

This'll also need an update to _default:_s in README.md

PR #55, opened 2019-07-25
Add a new Task that uses `kn` to create services

ImJasonH:
Also define a sample ServiceAccount, ClusterRole and ClusterRoleBinding
that can be used to authorize Knative Service deployments.

# Changes

Adds a new Task that uses [`kn`](https://github.com/knative/client) to create Knative services.

# Submitter Checklist

review them:

- [y] Includes [docs](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if user facing)
- [y] Commit messages follow [commit message best practices](https://github.com/tektoncd/community/blob/master/standards.md#commit-messages)

for more details._

/assign @vdemeester 

-----

I think we'd probably like to rely on `kn` over `knctl` going forward since I believe that's expected to become the "standard" CLI for interacting with Knative. If that's okay, we should consider deleting the `knctl` task, or otherwise documenting that it's not the preferred solution.

When `kn` supports it, we should also add tasks to modify traffic splits, etc.

### DIFF
--- a/kn/kn-create.yaml
+++ b/kn/kn-create.yaml
@@ -1,0 +1,23 @@
+apiVersion: tekton.dev/v1alpha1
+kind: Task
+metadata:
+  name: kn-create
+spec:
+  inputs:
+    params:
+    - name: service
+      description: Name of the service to create
+    resources:
+    - name: image
+      type: image
+  steps:
+  - name: create
+    image: gcr.io/knative-releases/github.com/knative/client/cmd/kn
+    command: ["/ko-app/kn"]
+    args:
+    - service
+    - create
+    - ${inputs.params.service}
+    - --image
+    - ${inputs.resources.image.url}
+    # TODO(jasonhall): Support limits/resources, env vars, etc.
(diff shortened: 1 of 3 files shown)


vdemeester:
> I think we'd probably like to rely on `kn` over `knctl` going forward since I believe that's expected to become the "standard" CLI for interacting with Knative. If that's okay, we should consider deleting the `knctl` task, or otherwise documenting that it's not the preferred solution.
> 
> When `kn` supports it, we should also add tasks to modify traffic splits, etc.

Yes ! I would really prefer having `kn` than `knctl` so I am more than fine to remove `knctl` once this is merge :wink:

kn/kn-create.yaml
```diff
@@ -0,0 +1,23 @@
+apiVersion: tekton.dev/v1alpha1
+kind: Task
+metadata:
+  name: kn-create
```
vdemeester: Should we also support `-f` on create – to be able to "update" the hard way a service ?

PR #60, opened 2019-08-17
Added task for Conftest

garethr:
# Changes

Conftest is a tool used to validate configuration files using Open
Policy Agent. This commit adds a `Task` spec for using Conftest from
Tekton, along with documentation of the parameters and usage.

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/conftest/conftest.yaml
+++ b/conftest/conftest.yaml
@@ -1,0 +1,33 @@
+apiVersion: tekton.dev/v1alpha1
+kind: Task
+metadata:
+  name: conftest
+spec:
+  inputs:
+    resources:
+    - name: source
+      type: git
+      targetPath: source
+    params:
+    - name: files
+    - name: policy
+      default: "policy"
+    - name: output
+      default: "stdout"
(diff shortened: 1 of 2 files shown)


conftest/conftest.yaml
```diff
@@ -0,0 +1,33 @@
...
+    - name: output
+      default: "stdout"
+    - name: args
+      type: array
+      default: []
+
+  steps:
+  - name: conftest
+    workingdir: /workspace/source
+    image: instrumenta/conftest:latest
```
vdemeester: @garethr I think we tend to prefer to have "fixed" version image instead of `latest`, does that make sense ?

garethr: Makes sense. Making this configurable would be good in that case I presume. Do we have any conventions yet around parameter names doing the same?

vdemeester: not really :thinking: It really depends on the images… on the go tasks we are using `version`, `buildah` uses a `stable` tags..

PR #67, opened 2019-09-03
s2i and openshift-client Dockerfiles improvements

chmouel:
* openshift-client/Dockerfile Make sure we cleanup after ourselves
* s2i/Dockerfile run yum update in docker image


openshift-client/Dockerfile
```diff
@@ -3,6 +3,8 @@ FROM registry.access.redhat.com/ubi8/ubi:latest
 RUN cd /tmp \
   && yum -y update \
   && yum install -y wget \
+  && yum clean all && \
```
piyush-garg: nit: `&&` double time on both lines

PR #68, opened 2019-09-05
Added Task `update-image-digests`.

chhsia0:
# Changes

The `update-image-digests` task works as follows:
1. After a predecessor task (specified via the `pipelineTask` parameter) finishes, retrieve its `resourcesResult` and construct a data file consists of (image-resource-name, digest) pairs.
2. Use `kubectl` to patch each image resource with its digest for each pair in the data file.
3. Then, all successor task could get the digests with string interpolation.

This is intended to be a workaround for https://github.com/tektoncd/pipeline/issues/216.
The `resourcesResoult` field is proposed to be removed in pipeline 0.9, where this Task will no
longer work.

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/update-image-digests/README.md
+++ b/update-image-digests/README.md
@@ -1,0 +1,88 @@
+# Update Image Digests
+
+This Task updates all image resources that has digests exported built by a predecessor task in the same pipeline run.
+
+## Install the Task
+
+```
+kubectl apply -f https://raw.githubusercontent.com/tektoncd/catalog/master/update-image-digests/update-image-digests.yaml
+```
+
+## Inputs
+
+### Parameters
+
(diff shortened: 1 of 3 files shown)


sbwsg:
@chhsia0 do you still need this?  We have Task results now that allow a Task to emit a digest as a result in a Pipeline and for another Task to accept that digest through params.

If you still want to add this to the catalog could you rebase it against the `v1beta1` branch?  Cheers!

vdemeester:
@chhsia0 if that's fine with you I am gonna go ahead and close this (mainly because it stale and we have different ways to do it). But feel free to re-open or create a new one that would do the same (using results)
/close

PR #74, opened 2019-09-13
Remove script from openshift client task

piyush-garg:
This will remove the script used in image for/
openshift client task.

As we are able to provide arguments as string
with tekton pipeline 0.6.0
we dont need script any more, also updated the
Dockerfile, tasks and README accordingly

# Submitter Checklist

review them:

for more details._


openshift-client/Dockerfile
```diff
@@ -3,15 +3,12 @@ FROM registry.access.redhat.com/ubi8/ubi:latest
...
-  && yum clean all && \
-  &&  rm -rf /var/cache/yum && \
+  && yum clean all \
   && wget https://mirror.openshift.com/pub/openshift-v4/clients/oc/latest/linux/oc.tar.gz \
   && tar -xvzf oc.tar.gz \
-  && mv oc /usr/local/bin/oc-origin \
+  && mv oc /usr/local/bin/oc \
   && rm -rf oc.tar.gz
 
-ADD script.sh /usr/local/bin/oc
```
chmouel: I wonder why this has been done like this, i am not even sure if it's me who has done this.

chmouel: ah nm you are explaining it already

piyush-garg: It is me who have done this, This we have done as previously tekton was not supporting multiple arguments as param. Now as we can provide that, we are removing the script and the task.

PR #75, opened 2019-09-13
Fix kn-create Task to pass args in accepted manner

### LINKED ISSUE #73: kn-create task not receiving the service name parameter
# Expected Behavior

Successful service creation using kn task at https://github.com/tektoncd/catalog/tree/master/kn

# Actual Behavior

Service creation failing with error
```
[create] 'service create' requires the service name given as single argument
```

1. Follow instructions at https://github.com/tektoncd/catalog/tree/master/kn
2. Reference a built image and a service name in taskrun.yaml
3. Create taskrun

```
17:38 ➜  catalog git:(master) ✗  kubectl create -f kn/kn-create.yaml 
task.tekton.dev/kn-create created

17:38 ➜  catalog git:(master) ✗  cat taskrun.yaml 
kind: TaskRun
metadata:
  generateName: kn-create-
spec:
  serviceAccount: kn-deployer-account  # <-- run as the authorized SA

  taskRef:
    name: kn-create
  inputs:
    params:
    - name: service
      value: hello
    resources:
    - name: image
      resourceSpec:
        type: image
        params:
        - name: url
          value: gcr.io/knative-samples/helloworld-go

17:38 ➜  catalog git:(master) ✗  kubectl create -f taskrun.yaml 
taskrun.tekton.dev/kn-create-ncwqk created

17:38 ➜  catalog git:(master) ✗  tkn taskrun list
NAME              STARTED         DURATION   STATUS             
kn-create-ncwqk   4 seconds ago   ---        Running(Pending)   

17:39 ➜  catalog git:(master) ✗  tkn taskrun logs kn-create-ncwqk
[create] 'service create' requires the service name given as single argument

container step-create has failed  : Error
```

ImJasonH:
Fixes #73

# Changes

`kn` apparently doesn't like flags passed with spaces, and prefers `--foo=bar` style (with `=`). The error message when you pass flags with a space is...confusing.

I've confirmed this change makes the kn-create Task successfully attempt to create the Service.

# Submitter Checklist

review them:

- [ n/a ] Includes [docs](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if user facing)
- [ y ] Commit messages follow [commit message best practices](https://github.com/tektoncd/community/blob/master/standards.md#commit-messages)

_See [the contribution guide(https://github.com/tektoncd/catalog/blob/master/CONTRIBUTING.md)
for more details._

### DIFF
--- a/kn/kn-create.yaml
+++ b/kn/kn-create.yaml
@@ -18,11 +18,9 @@ spec:
     image: gcr.io/knative-releases/github.com/knative/client/cmd/kn
     command: ["/ko-app/kn"]
     args:
-    - service
-    - create
-    - ${inputs.params.service}
-    - --image
-    - ${inputs.resources.image.url}
-    - --force
-    - ${inputs.params.force}
+    - 'service'
+    - 'create'
+    - '$(inputs.params.service)'
+    - '--image=$(inputs.resources.image.url)'
+    - '--force=$(inputs.params.force)'
     # TODO(jasonhall): Support limits/resources, env vars, etc.


kn/kn-create.yaml
```diff
@@ -18,11 +18,9 @@ spec:
     image: gcr.io/knative-releases/github.com/knative/client/cmd/kn
     command: ["/ko-app/kn"]
     args:
-    - service
-    - create
-    - ${inputs.params.service}
-    - --image
-    - ${inputs.resources.image.url}
-    - --force
```
navidshaikh: The only space separated flag kn didnt like was this bool flag `--force`. This worked for me
```
✗  git --no-pager diff                           
diff --git a/kn/kn-create.yaml b/kn/kn-create.yaml
index 209fc47..69e79ab 100644
--- a/kn/kn-create.yaml
+++ b/kn/kn-create.yaml
@@ -23,6 +23,5 @@ spec:
     - ${inputs.params.service}
     - --image
     - ${inputs.resources.image.url}
-    - --force
-    - ${inputs.params.force}
+    - --force=${inputs.params.force}
     # TODO(jasonhall): Support limits/resources, env vars, etc.
```

ImJasonH: Ah interesting, good to know. I think I'd prefer to keep it like this now, so they're consistently using `=` where possible.

The error message is not terribly helpful in this case, would you like me to file a bug for knative/client?

navidshaikh: > The error message is not terribly helpful in this case, would you like me to file a bug for knative/client ?

That'd be helpful, thanks! :-)

navidshaikh: @ImJasonH : On the same note, IMO we don't need single quotes. WDYT ?

Issue #78, opened 2019-09-14
Tag Images for openshift-client Task


danielhelfand:
Currently, the `openshift-client` task uses a tag of [latest](https://github.com/tektoncd/catalog/blob/master/openshift-client/openshift-client-task.yaml#L15) for its image. It would be a better practice to tag the image based on the version of Tekton pipelines the task would work with and explicitly call out what the version tag is for the task as opposed to using the `latest` tag.

There should be documentation on the available tags and how those tags correspond to certain versions of Tekton pipelines/the task. The previous versions of the task should also be archived and use the appropriate image tag.

This issue is in response to #74 where the task was updated for a new version of the OpenShift Tekton operator that is actually not available yet. I would like to ask that this task not be updated until the new version of an operator is available in the future.

piyush-garg:
@danielhelfand We have reverted the change for openshift client task that broke for the operator, also we have tagged the image this time, soon we will do the archiving of task like releasing a branch or something else. This is in discussions now.

PR #80, opened 2019-09-17
Openshift client task changes to work with 0.5

piyush-garg:
This will revert the changes that we made to make the task
working with 0.6 pipelines, but it breaks the backward
compatibility. So kind of reverting the change but started
tagging the image with tekton version

and in future to maintain the release cycle and then
update the task according to new version.

Address #78

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/openshift-client/openshift-client-task.yaml
+++ b/openshift-client/openshift-client-task.yaml
@@ -7,12 +7,10 @@ spec:
     params:
       - name: ARGS
         description: The OpenShift CLI arguments to run
-        type: array
-        default:
-          - "help"
+        default: help
   steps:
     - name: oc
-      image: quay.io/openshift-pipeline/openshift-cli:latest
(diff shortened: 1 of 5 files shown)


openshift-client/script.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
```
khrm: Why not use array params and avoid this script?

openshift-client/openshift-client-task.yaml
```diff
@@ -7,12 +7,10 @@ spec:
     params:
       - name: ARGS
         description: The OpenShift CLI arguments to run
-        type: array
```
khrm: Aren't argument array, why remove it?

openshift-client/script.sh
khrm: From the discussion, this is to ensure backward compatibility. So alright with me.

PR #81, opened 2019-09-17
Revert 'remove script' and fix the Dockerfile

### LINKED ISSUE #78: Tag Images for openshift-client Task

piyush-garg:
This is breaking the backward compatibility, better release
a branch for 0.5 and then update the tasks

This change will fix the issue of client
task giving error on tekton 0.5.0

Also, tag the image with the pipeline version
in spite of always latest

Fix #78

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/openshift-client/script.sh
+++ b/openshift-client/script.sh
@@ -1,0 +1,15 @@
+#!/bin/bash
+
+set -e
+
+command="/usr/local/bin/oc-origin"
+
+for args in "$@"; do
+    for arg in $args; do
+        command+=" $arg"
+    done
+done
+
+exec $command
+
+exit 0
--- a/openshift-client/Dockerfile
+++ b/openshift-client/Dockerfile
@@ -1,16 +1,17 @@
 FROM registry.access.redhat.com/ubi8/ubi:latest
 
-RUN cd /tmp \
-  && yum -y update \
+RUN yum -y update \
   && yum install -y wget \
-  && yum clean all && \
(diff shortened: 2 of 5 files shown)


openshift-client/script.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+
+set -e
+
+command="/usr/local/bin/oc-origin"
+
+for args in "$@"; do
```
ImJasonH: Is this necessary due to a deficiency/bug in array-type param handling? This is pretty gross, and I'd love to not need it.

piyush-garg: this we are doing for the previous tekton version, we may tag this to use with previous tekton pipeline, and update this without a script for current tekton version i.e. 0.6 as we have provided the feature of array type params.

PR #100, opened 2019-10-17
Allow passing extra flags to kaniko

chmouel:
When for example we want to add --skip-tls-verify or other flags

### DIFF
--- a/kaniko/kaniko.yaml
+++ b/kaniko/kaniko.yaml
@@ -11,6 +11,8 @@ spec:
     - name: CONTEXT
       description: The build context used by Kaniko.
       default: ./
+    - name: EXTRA_ARGS
+      default: ""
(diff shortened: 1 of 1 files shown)


bioball:
This doesn't seem to work for using multiple flags. I tried to set these flags:

```
--cache=true --cache-ttl=10h
```

I'm seeing this error:

```
Error: invalid argument "true --cache-ttl=10h" for "--cache" flag: strconv.ParseBool: parsing "true --cache-ttl=10h": invalid syntax
```

I don't really see how this could be correctly implemented, though. Is there some Tekton way to implement multiple flags?

vdemeester:
@bioball indeed, I think it was done before the introduction of `type: array` parameters, that would work in that case. We need to fix that.

PR #126, opened 2019-11-07
Add gke-deploy catalog Task and Pipeline, with a PipelineRun example.

joonlim:
Demonstrates how to set up a pipeline to build, push, deploy to GKE.


joonlim:
/assign @bobcatfish 
@dorbin

gke-deploy/gke-deploy.yaml
```diff
@@ -0,0 +1,19 @@
+apiVersion: tekton.dev/v1alpha1
+kind: Task
+metadata:
+  name: gke-deploy
+spec:
+  inputs:
+    # resources:
+    # - name: source-repo
```
sbwsg: is this right? the docs for the task mention the git resource but it's commented out here.

sbwsg:
Thanks for the extensive docs on usage.  One suggestion I have is to move the example app + markdown file under a folder called "example".  Otherwise this looks good to me!

dorbin:
@spomorski

gke-deploy/gke-deploy.yaml
joonlim: Thanks for catching that. That was leftover from some experimenting. Fixing now.

gke-deploy/build-push-gke-deploy-example.md
```diff
@@ -0,0 +1,220 @@
...
+1. Create the target cluster that gke-deploy will deploy your application to.
+
+  ```bash
+  DEPLOY_CLUSTER_NAME=[NAME]
+  DEPLOY_CLUSTER_LOCATION=[REGION/ZONE]
+  DEPLOY_CLUSTER_PROJECT=[PROJECT_ID]
+  gcloud container clusters create $DEPLOY_CLUSTER_NAME --zone=$DEPLOY_CLUSTER_LOCATION --project=$DEPLOY_CLUSTER_PROJECT
+  ```
+
+2. Add the `roles/storage.admin` role to the Google service account set up above in the project of the image will be pushed to (this project may be in the same project as the one with the cluster that is running Tekton Pipelines). This will allow the Pipeline to push an image to your project, as defined [here](https://cloud.google.com/container-registry/docs/access-control).
```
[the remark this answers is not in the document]
dorbin: Also, "...project may be in the same project..."
Do you mean just "...this may be the same project as..." ?

gke-deploy/build-push-gke-deploy-example.md
```diff
@@ -0,0 +1,220 @@
...
+
+  ```bash
+  IMAGE_REGISTRY_PROJECT=[PROJECT_ID]
+  GOOGLE_SA_NAME=[GOOGLE_SERVICE_ACCOUNT_NAME]
+  TEKTON_CLUSTER_PROJECT=[PROJECT_ID]
+  gcloud projects add-iam-policy-binding $IMAGE_REGISTRY_PROJECT --role roles/storage.admin --member "serviceAccount:$GOOGLE_SA_NAME@$TEKTON_CLUSTER_PROJECT.iam.gserviceaccount.com" --project=$IMAGE_REGISTRY_PROJECT
+  ```
+
+3. Add the `roles/container.developer` role to the Google service account set up above in the project of the target cluster (this project may be in the same project as the one with the cluster that is running Tekton Pipelines). This will allow the Pipeline to deploy your application to your cluster.
```
dorbin: Similar comment here ("project may be in the same project"

gke-deploy/build-push-gke-deploy-example.md
```diff
@@ -0,0 +1,220 @@
...
+  gcloud projects add-iam-policy-binding $DEPLOY_CLUSTER_PROJECT --role roles/container.developer --member "serviceAccount:$GOOGLE_SA_NAME@$TEKTON_CLUSTER_PROJECT.iam.gserviceaccount.com" --project=$DEPLOY_CLUSTER_PROJECT
+  ```
+
+4. Install the `build-push-gke-deploy` Pipeline.
+
+  ```bash
+  kubectl apply -f https://raw.githubusercontent.com/tektoncd/catalog/master/gke-deploy/build-push-gke-deploy.yaml
+  ```
+
+5. Create the PipelineRun config to run your Pipeline.
```
dorbin:
```suggestion
5. Create the `PipelineRun` config to run your Pipeline.
```

gke-deploy/build-push-gke-deploy-example.md
```diff
@@ -0,0 +1,220 @@
...
+
+6. Watch the logs.
+
+  ```bash
+  tkn pipelinerun logs build-push-gke-deploy-run --follow
+  ```
+
+  See other ways of logging your Tekton PipelineRun [here](https://github.com/tektoncd/pipeline/blob/master/docs/logs.md).
+
+  If your deployment is successful, `gke-deploy` will print a table displaying deployed reouces. You can visit the IP address printed the Service row.
```
dorbin:
```suggestion
  If your deployment is successful, `gke-deploy` will print a table displaying deployed resources. You can visit the IP address printed on the Service row.
```

gke-deploy/README.md
```diff
@@ -0,0 +1,205 @@
...
+
+## Inputs
+
+### Resources
+
+* **source-repo**: The Git source repository that contains your application's Kubernetes configs.
+
+### Parameters
+
+* **ARGS**: The arguments to pass to `gke-deploy` CLI.
```
demobox: From what I recall, a Kubernetes cluster was also going to be a resource type, that basically would give you a kubeconfig file.

If so, I suspect we should support that as an optional feature (i.e. you can either pass the target cluster via a pipeline resource, or via arguments to the task)?

/cc @sbwsg on the state of Kubernetes resources

gke-deploy/README.md
```diff
@@ -0,0 +1,205 @@
...
+the service account that the Task runs as must have the
+[`roles/container.developer`](https://cloud.google.com/iam/docs/understanding-roles#kubernetes-engine-roles) role.
+
+```bash
+gcloud iam service-accounts [DEPLOY_CLUSTER_PROJECT] add-iam-policy-binding \
+  --role roles/container.developer \
+  --member "serviceAccount:[SERVICE_ACCOUNT_EMAIL]"
+```
+
+### Running the Pipeline
```
demobox: Can we also demo a setup in which one pipeline docs `build-push`, and another triggers on push of the GCR image and does `deploy`? That would scale more readily to a multi-env setup, and is also closer to what we recommend to users.

joonlim: Yes, we can do that in a future PR. I have one ready but it won't work until the [script](https://github.com/tektoncd/pipeline/blob/master/docs/tasks.md#step-script) feature is released.

joonlim: I'll work on this in a future PR.

gke-deploy/README.md
sbwsg: this resource type exists but unfortunately in the current implementation resources can't be marked as optional.  this means the user would need to provide a cluster resource even if they wanted to use the args-only approach. :(

joonlim: Thanks Scott, I'll add a reminder to myself to update the pipeline with this once resources can be marked optional.

sbwsg: This is the feature request in tekton for optional resources: https://github.com/tektoncd/pipeline/issues/812

gke-deploy/example/build-push-gke-deploy-example.md
```diff
@@ -0,0 +1,220 @@
...
+    --member "serviceAccount:$TEKTON_CLUSTER_PROJECT.svc.id.goog[default/$K8S_SA_NAME]" $GOOGLE_SA_NAME@$TEKTON_CLUSTER_PROJECT.iam.gserviceaccount.com
+  kubectl annotate serviceaccount $K8S_SA_NAME iam.gke.io/gcp-service-account=$GOOGLE_SA_NAME@$TEKTON_CLUSTER_PROJECT.iam.gserviceaccount.com
+
+  # Install Tekton Pipelines
+  kubectl apply --filename https://storage.googleapis.com/tekton-releases/pipeline/latest/release.yaml
+  ```
+
+## Install Tekton Pipelines CLI
+
+1. Install the Tekton Pipelines CLI to view your logs by following the instructions for your machine at https://github.com/tektoncd/cli.
```
dorbin: Doesn't need the `1. ` if this is the only step under this heading.

gke-deploy/README.md
demobox: > this resource type exists but unfortunately in the current implementation resources can't be marked as optional

Ah, OK, thanks. Would you recommend leaving things as-is for now, or recommending that users create a modified the task definition for their own use if the want to use a cluster resource as input?

sbwsg: Hmmm.  How would you feel about adding two tasks in this PR?  One for those users wanting to deploy using a cluster resource and one for those preferring to use args?

joonlim: Sure, I can add that in another PR, since this one has already been reviewed by docs

sbwsg: sounds good.  sorry this is an admittedly lame solution. we need better support for this.

demobox: Especially if we have a way to do this "properly" coming soon, I'd prefer to avoid creating two "official" tasks and then trying to explain to users which to choose.

For the time being, I'd rather either explain that we will add support for a cluster input when optional resources become available, or that users could use cluster inputs by making their own copy of the task, or both.

sbwsg: :+1: makes sense

PR #134, opened 2019-11-15
Add default value of LOCAL_CI_RUN flag

piyush-garg:
This will add the default value of LOCAL_CI_RUN
flag and also move install tekton pipeline
out of not LOCAL_CI_RUN as locally also it needs
to be installed to run tests

# Submitter Checklist

review them:

for more details._


chmouel:
/hold

we don't want to install the pipeline in local_ci_run if that's for our downstream we need to do it in another way,

PR #149, opened 2019-11-27
Add a task for terraform-cli command execution Closes #124

### LINKED ISSUE #124: terraform task(s)
We should provide Tasks that are able to use `terraform` to deploy things through it.

/kind feature

raffamendes:
# Changes

Creation of a task  to run terraform-cli commands

This is fairly complex task to setup, if someone can test this with azure or AWS would be great, I tested with OpenStack.
# Submitter Checklist

review them:

for more details._

### DIFF
--- a/terraform-cli/Dockerfile
+++ b/terraform-cli/Dockerfile
@@ -1,0 +1,18 @@
+FROM registry.access.redhat.com/ubi8/ubi
+
+RUN yum update -y && \
+    yum install zip -y && \
+    curl -O https://releases.hashicorp.com/terraform/0.12.13/terraform_0.12.13_linux_amd64.zip && \
+    unzip terraform_0.12.13_linux_amd64.zip -d /usr/local/bin && \
+    yum clean all && \
+    rm -rf /var/cache/yum
+
+
+USER 1001
+
+ENTRYPOINT ["/usr/local/bin/terraform"]
+
+
+
+
+
--- a/terraform-cli/terraform-cli-task.yaml
+++ b/terraform-cli/terraform-cli-task.yaml
@@ -1,0 +1,36 @@
+---
+apiVersion: tekton.dev/v1alpha1
+kind: Task
+metadata:
+  name: terraform-cli
+spec:
+  inputs:
+    resources:
+      - name: source
+        type: git
+    params:
+      - name: ARGS
+        description: The terraform cli commands to tun
+        type: array
+        default:
+          - "--help"
+      - name: terraform-secret
+        description: The terraform secret with credentials
+        type: string
+        default: "terraform-creds"
+  steps:
+    - name: init
+      image: quay.io/rcmendes/terraform-cli:latest
+      workingDir: /workspace/source
+      command: ["/usr/local/bin/terraform"]
+      args:
+        - "init"
+    - name: terraform-cli
+      image: quay.io/rcmendes/terraform-cli:latest
+      workingDir: /workspace/source
+      command: ["/usr/local/bin/terraform"]
+      args:
+        - "$(inputs.params.ARGS)"
+      envFrom:
+        - secretRef:
+            name: $(inputs.params.terraform-secret)
--- a/terraform-cli/OWNERS
+++ b/terraform-cli/OWNERS
@@ -1,0 +1,8 @@
+approvers:
+- chmouel
+- vdemeester
+- raffamendes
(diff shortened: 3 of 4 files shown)


raffamendes:
my bad :sweat_smile: 
ctrl+c ctrl+v detected kkkk

terraform-cli/terraform-cli-task.yaml
```diff
@@ -0,0 +1,36 @@
...
+  inputs:
+    resources:
+      - name: source
+        type: git
+    params:
+      - name: ARGS
+        description: The terraform cli commands to tun
+        type: array
+        default:
+          - "--help"
```
chmouel: Why would be the default as --help ? Do you want to make sure it fails if the user don't specify the ARGS?

terraform-cli/terraform-cli-task.yaml
```diff
@@ -0,0 +1,36 @@
...
+      - name: terraform-secret
+        description: The terraform secret with credentials
+        type: string
+        default: "terraform-creds"
+  steps:
+    - name: init
+      image: quay.io/rcmendes/terraform-cli:latest
```
chmouel: Wed rather use official image, if it need to use a custom user image, it would be best if we have the Dockerfile in the repository (i.e: like s2i or openshift-cli task). Thanks!

terraform-cli/terraform-cli-task.yaml
raffamendes: I'm using this as default just like the OC client task and the tower CLI task that i previously submitted, so the command doesn't run indefinitely

terraform-cli/terraform-cli-task.yaml
raffamendes: I think there's a official image from hashicorp, but I created one using an UBI image, how can I submit this image to the official repository? meanwhile I'll test the hashicorp official image for this

terraform-cli/terraform-cli-task.yaml
chmouel: ah i didnt' know about this pattern 🤔  cc @vdemeester

terraform-cli/terraform-cli-task.yaml
chmouel: You just need to add the Dockerfile in this repo i.e: https://github.com/tektoncd/catalog/blob/master/s2i/Dockerfile

raffamendes: Understood, will add that, and will add the Dockerfile for the ansible-tower-cli task as well in another PR.

raffamendes: Done, Dockerfile added :+1:

terraform-cli/Dockerfile
```diff
@@ -0,0 +1,15 @@
+FROM registry.access.redhat.com/ubi8/ubi
+
```
chmouel: so a couple of things : 

1) you may want to use ubi-minimal, unless you need more?

2) you don't need to install wget, curl is available (so curl -O would behave the same)

3) Make sure you do some yum update before i.e: https://github.com/tektoncd/catalog/blob/master/s2i/Dockerfile#L4 so you don't get show as security issue in quay.

4) Make sure you do yum clean all after any yum install so your images is slighty smaller,

5) it would be great if by default we are able to install latest stable, instead of having a static version number getting outdated quickly?

Sorry for nitpicking but those are good habits in general to take when writing Dockerfile,

other than that things looks fine, thanks 👍 

have the `yum install wget zip` before the `yum clean all`

(so it would not show as security issue in quay and make the image smaller)

raffamendes: I appreciate all the remarks, will use this to improve my Dockerfile writing, just to point out two things

1. ubi-minimal doesn't have yum :sweat_smile: that's why I use this version
5. I couldn't find a link to a stable version on hashicorp's site, they only provide the binary in that naming format, I agree that this is bad for maintaining

chmouel:
/lgtm

Looking Good, Thanks! 🤙🏽😻

raffamendes:
I accidentally synced my branch to the upstream and it added somethings to the PR, I already fixed that, but the lgtm label got lost, can someone tag it back? :sweat_smile: 
cc @chmouel @vdemeester

chmouel:
Thanks! 👊

Issue #151, opened 2019-11-27
Problems getting the cloud native buildpacks example to run because of missing builder image


kinkerl:
# Expected Behavior

Follow the example in the `buildpacks`directory and get it to run. 

# Actual Behavior

Looks like the used builder docker image does not exist anymore. It is referenced in the task run as "gcr.io/cncf-buildpacks-ci/tekton-cnb-test:bionic" but it looks like it got removed. Is there a new home for it somewhere?

# Steps to Reproduce the Problem

1. docker pull gcr.io/cncf-buildpacks-ci/tekton-cnb-test:bionic

# Additional Info

sbose78:
So, this works:
`docker pull cloudfoundry/cnb:bionic`

ameyer-pivotal:
@kinkerl That builder was just an example, not a real builder. We're updating this task to conform to a later builder. Try out the task in https://github.com/tektoncd/catalog/pull/175.

jromero:
@vdemeester This should be resolved.

@kinkerl let us know if that is incorrect.

vdemeester:
Ah I see #175 refered this issue :angel: 
/close

PR #175, opened 2020-02-04
Update CNB task to conform to Platform API 0.2

natalieparellano:
- Also update buildpacks README to have more easily consumable example

Signed-off-by: Natalie Arellano <narellano@vmware.com>

# Changes

* Update CNB task to conform to Platform API 0.2
  * The latest lifecycle (v0.6.0) implements the new platform API
* Also update buildpacks README to have more easily consumable example
  * This is to address the concern raised in https://github.com/tektoncd/catalog/issues/151

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/buildpacks/README.md
+++ b/buildpacks/README.md
@@ -18,30 +18,36 @@ The Cloud Native Buildpacks website describes v3 buildpacks as:
 kubectl apply -f https://raw.githubusercontent.com/tektoncd/catalog/master/buildpacks/buildpacks-v3.yaml
 ```
 
-> **NOTE:** This task is currently only compatible with Tekton **v0.6.0** and above.
+> **NOTE:** This task is currently only compatible with Tekton **v0.6.0** and above, and CNB Platform API 0.2 (lifecycle v0.6.0 and above). For previous Platform API versions, [see below](#previous-platform-api-versions).
 
 ## Inputs
 
 ### Parameters
 
-* **BUILDER_IMAGE** The image on which builds will run (must include v3 lifecycle and compatible buildpacks; _required_)
-* **USE_CRED_HELPERS:** Use Docker credential helpers. Set to `"true"` or
+* **`BUILDER_IMAGE`**: The image on which builds will run (must include v3 lifecycle and compatible buildpacks; _required_)
+
+* **`USE_CRED_HELPERS`**: Use Docker credential helpers. Set to `"true"` or
   `"false"` as string a value. (_default:_ `"false"`)
-* **CACHE** The name of the persistent app cache volume (_default:_ an empty
+
+* **`CACHE`**: The name of the persistent app cache volume (_default:_ an empty
   directory -- effectively no cache)
-* **USER_ID** The user ID of the builder image user, as a string value (_default:_ `"1000"`)
-* **GROUP_ID** The group ID of the builder image user, as a string value (_default:_ `"1000"`)
+
+* **`USER_ID`**: The user ID of the builder image user, as a string value (_default:_ `"1000"`)
+
+* **`GROUP_ID`**: The group ID of the builder image user, as a string value (_default:_ `"1000"`)
+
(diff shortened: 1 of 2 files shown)


natalieparellano:
@googlebot I signed it!

(corporate CLA)

micahyoung:
@googlebot I signed it!

(corporate CLA)

buildpacks/README.md
```diff
@@ -56,30 +62,55 @@ metadata:
...
         - name: url
-          value: gcr.io/my-repo/my-image
+          value: <your output image tag, e.g. "gcr.io/app-repo/app-image:app-tag">
+```
+
+### Example builders
+
+Cloud Foundry:
+ - `cloudfoundry/cnb:bionic`
+ - `cloudfoundry/cnb:cflinuxfs3`
```
jromero: Any reason why `tiny` is missing here? Also, is it possible to link them to any resource applicable such as the docker hub page itself: 

- https://hub.docker.com/r/cloudfoundry/cnb
- https://hub.docker.com/r/heroku/buildpacks/

natalieparellano: Originally we had just the output of `pack suggest-builders` which includes `tiny` - looks like that was changed to be a more streamlined output, but not sure why.

micahyoung: To play devil's advocate, maybe we should just list a one CF builder and one Heroku builder here, perhaps`cloudfoundry/cnb:cflinuxfs3`, since it includes all our buildpacks and potentially decreases the number of support issues we'll see of people just not having their apps detected because the accidentally used the wrong builder. Since tekton may be a less common and harder to debug use-case for lifecycle, we might want to minimize our support load. Thoughts?

micahyoung: But we should definitely also link to those builder pages, though I still don't see a good explanation on which builder to use for first-time users.

micahyoung:
@googlebot I signed it!

buildpacks/buildpacks-v3.yaml
```diff
@@ -1,134 +1,132 @@
...
+          mountPath: /layers
+        - name: $(inputs.params.CACHE)
+          mountPath: /cache
 
   volumes:
-  - name: empty-dir
-    emptyDir: {}
-  - name: layers-dir
-    emptyDir: {}
+    - name: empty-dir
```
otaviof: @natalieparellano should `empty-dir` also use `$(inputs.params.CACHE)` placeholder?

buildpacks/README.md
ameyer-pivotal: This was meant to give a list of a few builders users can get started with, but not necessarily show exactly the same output as `suggest-builders` (otherwise, we should just direct them to run that command).

ameyer-pivotal: +1 on linking to a resource though

ameyer-pivotal: > To play devil's advocate, maybe we should just list a one CF builder and one Heroku builder here, perhaps`cloudfoundry/cnb:cflinuxfs3`, since it includes all our buildpacks and potentially decreases the number of support issues we'll see of people just not having their apps detected because the accidentally used the wrong builder (since there's nothing describing which to pick). Since Tekton may be a less common and harder to debug use-case for lifecycle, we might want to minimize our support load. Thoughts?

I like that idea. We just need something that users can play with. Perhaps we can direct them to pack or the docs for more details on builders

ameyer-pivotal:
@ImJasonH Is there anything more we need to do with CLA? We've all signed.

natalieparellano:
@googlebot I signed it!

natalieparellano:
@googlebot I fixed it.

buildpacks/buildpacks-v3.yaml
natalieparellano: I don't believe so - `CACHE` will default to the `empty-dir` if no value is provided.

PR #178, opened 2020-02-07
Update openshift client task to use script mode

piyush-garg:
This will update openshift-client to be used in script mode

This will help to pass multiple commands or parametrized
commands in a single taskrun

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/openshift-client/README.md
+++ b/openshift-client/README.md
@@ -24,11 +24,15 @@ kubectl apply -f https://raw.githubusercontent.com/tektoncd/catalog/master/opens
 
 * **ARGS:** args to execute which are appended to `oc` e.g. `start-build myapp` (_default_: `help`)
 
+* **SCRIPT:** script of oc commands to execute  e.g. `oc get pod $1 -0 yaml` This will take the first value of ARGS as pod name (_default_: `oc $@`)
+
 ## Inputs `openshift-client-kubecfg`
 
 ### Parameters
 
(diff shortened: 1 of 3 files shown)


piyush-garg:
@vdemeester I think we should keep the old one also, so that it may not affect the users who are using this.

piyush-garg:
We can do that, but that is adding no value.

Do you see any value with adding a new param `SCRIPT`?

We are adding the SCRIPT task to support multiple commands in task like

```
oc rollout latest myapp
oc rollout status myapp
```

vdemeester:
> We can do that, but that is adding no value.
> 
> Do you see any value with adding a new param `SCRIPT`?
> 
> We are adding the SCRIPT task to support multiple commands in a task like
> 
> ```
> oc rollout latest myapp
> oc rollout status myapp
> ```

well, actually it can add some making a Pipeline a little bit more re-usable. (This is a simple case)

```yaml
kind: Pipeline
metadata:
  name: demo-pipeline
spec:
  params:
  - name: app
    description: app of the pipeline
  tasks:
  - name:foo
    taskRef:
      name: unit-tests
    resources:
      params:
      - name: ARGS
        value: $(params.app)
      - name: SCRIPT
        value: |
          oc rollout latest $1
          oc rollout status $1
```

piyush-garg:
That way, we can find many cases like, a user needs two args or a user need some resource and so on. I am still not sure about changing the current tasks

piyush-garg:
@vdemeester I have updated the PR. Please have a look.

PR #181, opened 2020-02-13
Add Jib Gradle catalog / fix Docker config issue for Jib Maven

chanseokoh:
# Changes

Add Jib Gradle catalog / fix Docker config issue for Jib Maven

- Add Jib Gradle catalog: counterpart of Jib Maven.
- Change the user home inside JVM from `/builder/home` to `/tekton/home` in Jib Maven
   - See: https://github.com/tektoncd/pipeline/issues/2013#issuecomment-585908031
- Run Jib Maven in batch mode (`mvn -B ...`)

(@GoogleContainerTools/java-tools-build FYI)

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/jib-gradle/jib-gradle.yaml
+++ b/jib-gradle/jib-gradle.yaml
@@ -1,0 +1,67 @@
+apiVersion: tekton.dev/v1alpha1
+kind: Task
+metadata:
+  name: jib-gradle
+spec:
+  inputs:
+    params:
+    - name: DIRECTORY
+      description: The directory containing the app, relative to the source repository root
+      default: .
+    - name: CACHE
+      description: The name of the volume for caching Gradle artifacts, local Maven repository, and base image layers
+      default: empty-dir-volume
+    - name: INSECUREREGISTRY
+      description: Whether to allow insecure registry
+      default: "false"
+    resources:
+    - name: source
+      type: git
+  outputs:
+    resources:
+    - name: image
+      type: image
+  steps:
+  - name: build-and-push
+    image: gcr.io/cloud-builders/gradle
+    command: ["sh", "-c"]
+    args:
+      - |
+        # Adds Gradle init script that applies the Jib Gradle plugin.
+        echo "initscript {
+                repositories { maven { url 'https://plugins.gradle.org/m2' } }
+                dependencies { classpath 'gradle.plugin.com.google.cloud.tools:jib-gradle-plugin:+' }
+              }
+              rootProject {
+                afterEvaluate {
+                  if (!project.plugins.hasPlugin('com.google.cloud.tools.jib')) {
+                    project.apply plugin: com.google.cloud.tools.jib.gradle.JibPlugin
+                  }
+                }
+              }" > /tekton/home/init-script.gradle
+        # Runs the Gradle Jib build.
+        gradle jib \
+          --stacktrace --console=plain \
+          --init-script=/tekton/home/init-script.gradle \
+          -Duser.home=/tekton/home \
+          -Dgradle.user.home=/tekton/home/.gradle \
+          -Djib.allowInsecureRegistries=$(inputs.params.INSECUREREGISTRY) \
+          -Djib.to.image=$(outputs.resources.image.url)
+    workingDir: /workspace/source/$(inputs.params.DIRECTORY)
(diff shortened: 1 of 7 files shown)


jib-gradle/jib-gradle.yaml
```diff
@@ -0,0 +1,67 @@
...
+      type: git
+  outputs:
+    resources:
+    - name: image
+      type: image
+  steps:
+  - name: build-and-push
+    image: gcr.io/cloud-builders/gradle
+    command: ["sh", "-c"]
+    args:
```
chmouel: I believe you can just simply replace this with `script` in here and removing the command part

chanseokoh: Ah, that sounds better. Thanks!

chanseokoh:
Any chance to get this reviewed?

jib-gradle/jib-gradle.yaml
```diff
@@ -0,0 +1,67 @@
...
+              afterEvaluate {
+                if (!project.plugins.hasPlugin('com.google.cloud.tools.jib')) {
+                  project.apply plugin: com.google.cloud.tools.jib.gradle.JibPlugin
+                }
+              }
+            }" > /tekton/home/init-script.gradle
+      # Runs the Gradle Jib build.
+      gradle jib \
+        --stacktrace --console=plain \
+        --init-script=/tekton/home/init-script.gradle \
```
chmouel: I don't think we should use /tekton/home directly in case if that change and with the work @sbwsg is planning to do, what do you think Scott?

chmouel:
Fine by me, just need to make sure about /tekton/home if there is a better way to do this,

jib-gradle/jib-gradle.yaml
sbwsg: Yeah, unless I'm misunderstanding the purpose I think the safest thing will be to use `$HOME`.  It should be `/tekton/home` until we flip the default behaviour to be whatever the image expects to be HOME and then it should follow that as well.

chanseokoh: @sbwsg @chmouel I don't think I can use `$HOME` in the following locations. Knowing that, I'd rather use the `/tekton/home` literal, since at least things won't break even if `$HOME` changes under the hood. It will continue to work; it will be just that everyone will assume `/tekton/home` is a user home and use it.

```yaml
    - name: $(inputs.params.CACHE)
      mountPath: /tekton/home/.gradle/caches
      subPath: gradle-caches
    - name: $(inputs.params.CACHE)
      mountPath: /tekton/home/.gradle/wrapper
      subPath: gradle-wrapper
    - name: $(inputs.params.CACHE)
      mountPath: /tekton/home/.m2
      subPath: m2-cache
    - name: $(inputs.params.CACHE)
      mountPath: /tekton/home/.cache
      subPath: jib-cache
```

sbwsg: Could this be an opportunity to use [workspaces](https://github.com/tektoncd/pipeline/blob/master/docs/tasks.md#workspaces) instead of raw volumes?  You wouldn't need to pass a string volume name around as a param, users of your task would have the opportunity to pass PVCs or empty dirs as they see fit, and they could specify subPaths of those volumes if they wanted to share the caches on a single volume as is done here.

`/tekton` is generally intended as a reserved area for Tekton internals.  Having it include a home directory seems, in hindsight, to be a mistake.  I totally understand if you need a literal directory name to use but I don't recommend anything under /tekton.

chanseokoh: Unfortunately, the user home must be `/tekton/home` in JVM, because Tekton auth creates a Docker config at `/tekton/home/.docker/config.json`. Unless Tekton fixes this or has an option to create that file on `<actual home by image>/.docker/config.json`, I have no choice but force `/tekton/home` as the home directory. Otherwise, Jib just cannot find this file.

Yeah, this is very unfortunate and has been pretty annoying as I said in https://github.com/tektoncd/pipeline/issues/2013#issuecomment-585908031 I hope Tekton fixes the home issue quickly.

sbwsg: Oh dear.  OK then.  :+1:

PR #123, opened 2019-11-06
Add a git task to clone a repository without PipelineResource

vdemeester:
# Changes

This is part of documenting and providing Tasks in the catalog that
would help user not using PipelineResource.

Related to [tektoncd/pipeline#1369](https://github.com/tektoncd/pipeline/issues/1369).
Part of #95 

Signed-off-by: Vincent Demeester <vdemeest@redhat.com>

# Submitter Checklist

These are the criteria that every PR should meet, please check them off as you
review them:

- [ ] Includes [docs](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if user facing)
- [x] Commit messages follow [commit message best practices](https://github.com/tektoncd/community/blob/master/standards.md#commit-messages)

_See [the contribution guide](https://github.com/tektoncd/catalog/blob/master/CONTRIBUTING.md)
for more details._

PR #182, opened 2020-02-14
Update the git fetch task to include all params used in the Git PipelineResource

sbwsg:
# Changes

This commit makes several changes to the fetch-git task in the catalog:

1. Update the set of params to match all those of the Git PipelineResource. This
has been done because the Git PipelineResource is remaining alpha when the rest
of Tekton Pipelines goes beta. This task should provide an easy transition for
users currently relying on the pipeline resource.

2. Replace the volume mount with a workspace. This is done to promote workspaces
as the way to weave volumes through a pipeline.

3. Emit the fetched SHA as a Task Result which [will soon be consumable](https://github.com/tektoncd/pipeline/pull/2042) by other
tasks in a pipeline.

4. Rename it to git-clone [per feedback on the PR for the original fetch-git](https://github.com/tektoncd/catalog/pull/123#discussion_r371861807).

This PR continues to use the git-init image that is built as part of Tekton Pipelines.
I think that this is fine for now (?) but that we should eventually rewrite the
git-init binary as a script directly inside the git-clone task so that users can
more easily see exactly what the task does. On the bright side this does have the
benefit that the behaviour of the task will exactly mirror the pipeline reosurce
for compatibility's sake.

# Submitter Checklist

review them:

for more details._


sbwsg:
> Question (for the future), shouldn't we use a simple image (ubuntu + git or something) and directly use git in script ? thinking

:100: agree.

PR #188, opened 2020-02-19
Rewrite git-clone as shell script

chmouel:
We are rewritting git-clone as a 'pure' bourne shell script with no dependence
on the git-init image.

I have rewritten almost word by word the go code from here https://git.io/JvBCr

Error handling are handled by shell's set -e so if any command fails it will
just exit 1,

There is some bugs in go code submodules which is done always so i have handled
this differently to properly be conditional to the parameter.

# Changes

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/git/git-clone.yaml
+++ b/git/git-clone.yaml
@@ -40,42 +40,78 @@ spec:
     description: The precise commit SHA that was fetched by this Task
   steps:
   - name: clone
-    image: gcr.io/tekton-releases/github.com/tektoncd/pipeline/cmd/git-init:latest
+    image: alpine/git:latest
     script: |
+      #!/usr/bin/env sh
+      # Most errors handling are handled by the set -eux
+      set -eux
       CHECKOUT_DIR="$(workspaces.output.path)/$(inputs.params.subdirectory)"
+      URL="$(inputs.params.url)"
+      REVISION="$(inputs.params.revision)"
+      SUBMODULES="$(inputs.params.submodules)"
+      SSLVERIFY="$(inputs.params.sslVerify)"
+      DEPTH="$(inputs.params.depth)"
 
-      cleandir() {
-        # Delete any existing contents of the repo directory if it exists.
-        #
-        # We don't just "rm -rf $CHECKOUT_DIR" because $CHECKOUT_DIR might be "/"
-        # or the root of a mounted volume.
-        if [[ -d "$CHECKOUT_DIR" ]] ; then
-          # Delete non-hidden files and directories
-          rm -rf "$CHECKOUT_DIR"/*
-          # Delete files and directories starting with . but excluding ..
-          rm -rf "$CHECKOUT_DIR"/.[!.]*
-          # Delete files and directories starting with .. plus any other character
-          rm -rf "$CHECKOUT_DIR"/..?*
-        fi
+      [[ "$(inputs.params.deleteExisting)" == "true" && -d ${CHECKOUT_DIR} ]] && {
+        find ${CHECKOUT_DIR}/|grep -v "^${CHECKOUT_DIR}/$"|xargs  rm -rf
       }
 
-      if [[ "$(inputs.params.deleteExisting)" == "true" ]] ; then
-        cleandir
-      fi
+      # https://git.io/JvBCV
+      ensureHomeEnv() {
+          local realhomedir=$(getent passwd $(whoami)|cut -f6 -d:)
+          if [[ ! -d ${realhomedir} ]];then
+              mkdir -p ${realhomedir}
+          fi
 
-      /ko-app/git-init \
-        -url "$(inputs.params.url)" \
-        -revision "$(inputs.params.revision)" \
-        -path "$CHECKOUT_DIR" \
-        -sslVerify "$(inputs.params.sslVerify)" \
-        -submodules "$(inputs.params.submodules)" \
-        -depth "$(inputs.params.depth)"
-      cd "$CHECKOUT_DIR"
-      RESULT_SHA="$(git rev-parse HEAD | tr -d '\n')"
-      EXIT_CODE="$?"
-      if [ "$EXIT_CODE" != 0 ]
-      then
-        exit $EXIT_CODE
-      fi
-      # Make sure we don't add a trailing newline to the result!
-      echo -n "$RESULT_SHA" > $(results.commit.path)
+          if [[ ${realhomedir} == ${HOME} ]];then
+              return
+          fi
+
+          if [[ ! -e ${realhomedir}/.ssh && -w ${realhomedir}/.ssh ]];then
+              ln -svf ${HOME}/.ssh ${realhomedir}/.ssh
+          fi
+      }
+
+      # https://git.io/JvBCr
+      fetch() {
+          local revision=${REVISION:-master}
+          [[ -n ${CHECKOUT_DIR} ]] && {
+            mkdir -p ${CHECKOUT_DIR}
+            cd ${CHECKOUT_DIR}
+          }
+
+          git init .
+
+          trimmedURL=$(echo ${URL}|sed -e 's/^[ ]*//' -e 's/[ ]*$//')
+          git remote add origin ${trimmedURL}
+
+          git config http.sslVerify ${SSLVERIFY}
+
+          fetchArgs="fetch"
+          if [[ ${DEPTH} > 0 ]];then
+              fetchArgs="${fetchArgs} --tags  --depth=${DEPTH}"
+          fi
+
+          # Translating original go code from https://git.io/JvEtG
+          # original comment: Fetch can fail if an old commitid was used so try
+          # git pull, performing regardless of error as no guarantee that the
+          # same error is returned by all git servers gitlab, github etc...
+          if git ${fetchArgs} && git reset --hard ${revision};then
+             true
+          else
+              git reset --hard FETCH_HEAD
+          fi
+
+          if [[ ${SUBMODULES} == "true" || ${SUBMODULES} == "yes" ]];then
+              git submodule init
+              git submodule update --recursive
+              git submodule status
+          fi
+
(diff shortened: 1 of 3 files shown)


git/git-clone.yaml
```diff
@@ -36,22 +36,66 @@ spec:
...
+          trimmedURL=$(echo ${URL}|sed -e 's/^[ ]*//' -e 's/[ ]*$//')
+          git remote add origin ${trimmedURL}
+
+          git config --global http.sslVerify ${SSLVERIFY}
+
+          fetchArgs="fetch"
+          if [[ ${DEPTH} > 0 ]];then
+              fetchArgs="${fetchArgs} --depth=${DEPTH}"
+          fi
+          if git ${fetchArgs};then
```
sbwsg: Missing `[[` `]]`?

sbwsg: Actually maybe I misunderstand what this is doing.  What does this do?

git/git-clone.yaml
```diff
@@ -36,22 +36,66 @@ spec:
...
+              git submodule init
+              git submodule update --recursive
+          fi
+
+          echo "Successfully cloned ${trimmedURL} @ ${REVISION} in path ${PATH}"
+      }
+
+      ensureHomeEnv
+      fetch
+      echo -n "$(git rev-parse HEAD | tr -d '\n')" > $(results.commit.path)
```
sbwsg: nit: Maybe wrap `$(results.commit.path)` in quotes?

git/git-clone.yaml
```diff
@@ -36,22 +36,66 @@ spec:
...
+          git remote add origin ${trimmedURL}
+
+          git config --global http.sslVerify ${SSLVERIFY}
+
+          fetchArgs="fetch"
+          if [[ ${DEPTH} > 0 ]];then
+              fetchArgs="${fetchArgs} --depth=${DEPTH}"
+          fi
+          if git ${fetchArgs};then
+              git pull origin master
```
sbwsg: I think it's possible that there is no "master" branch on the remote.

What is the purpose of this pull? We fetch in the previous line and then we checkout the specific revision on the next line.

git/git-clone.yaml
chmouel: It will check if the command exit properly  and if it doesn't it will fallback, I have simply converted the GO code, and I assume that fallback was written to cover a user case... i.e: 

https://github.com/tektoncd/pipeline/blob/4e9380fe0d0e7ce862ff23781f614ad8e6de9faf/pkg/git/git.go#L91-L92

git/git-clone.yaml
```diff
@@ -36,22 +36,66 @@ spec:
     description: The precise commit SHA that was fetched by this Task
   steps:
   - name: clone
-    image: gcr.io/tekton-releases/github.com/tektoncd/pipeline/cmd/git-init:latest
+    image: alpine/git:latest
     script: |
+      # Most errors handling are handled by the set -eux
```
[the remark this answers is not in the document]
chmouel: OKay! bash is not available in that alpine image tho!

git/git-clone.yaml
chmouel: yeah i wondered it too to be honest, i think we can just git reset --hard anyway...

chmouel:
Adding @sthaha for advice since he is the git guru 😀

git/git-clone.yaml
```diff
@@ -36,22 +36,71 @@ spec:
...
+          fi
+
+          if [[ ! -e ${realhomedir}/.ssh ]];then
+              ln -svf ${HOME}/.ssh ${realhomedir}/.ssh
+          fi
+      }
+
+      # https://git.io/JvBCr
+      fetch() {
+          [[ ${REVISION} == "" ]] && REVISION=master
```
sthaha: How about `REVISION=${REVISION:-master}`

git/git-clone.yaml
```diff
@@ -36,22 +36,71 @@ spec:
...
+          if [[ ${CHECKOUT_DIR} != "" ]];then
+              git init ${CHECKOUT_DIR}
+              cd ${CHECKOUT_DIR}
+          else
+              git init .
+          fi
+          trimmedURL=$(echo ${URL}|sed -e 's/^[ ]*//' -e 's/[ ]*$//')
```
sthaha: would $( ... ) get substituted to "" ?

git/git-clone.yaml
```diff
@@ -36,22 +36,71 @@ spec:
...
+          git remote add origin ${trimmedURL}
+
+          git config --global http.sslVerify ${SSLVERIFY}
```
sthaha: Do we really need `--global`?

git/git-clone.yaml
```diff
@@ -36,22 +36,71 @@ spec:
...
+              git submodule init
+              git submodule update --recursive
+          fi
+
+          echo "Successfully cloned ${trimmedURL} @ ${REVISION} in path ${PATH}"
+      }
+
+      ensureHomeEnv
+      fetch
+      echo -n "$(git rev-parse HEAD | tr -d '\n')" > "$(results.commit.path)"
```
sthaha: echo isn't needed - right? i.e. `git rev-parse HEAD | tr -d '\n' > "$( ...)` should work - right?

git/git-clone.yaml
```diff
@@ -36,22 +36,71 @@ spec:
...
+
+          if [[ ! -e ${realhomedir}/.ssh ]];then
+              ln -svf ${HOME}/.ssh ${realhomedir}/.ssh
+          fi
+      }
+
+      # https://git.io/JvBCr
+      fetch() {
+          [[ ${REVISION} == "" ]] && REVISION=master
+          if [[ ${CHECKOUT_DIR} != "" ]];then
```
sthaha: nit: this can  reduced to 
```
test -z "$CHECKOUT_DIR" || cd "$CHECKOUT_DIR"
git init .
```

git/git-clone.yaml
sthaha: also just curious  ... wouldn't `trimmedURL=$(echo $URL)` work :D ?

chmouel:
/hold 

will do this rewrite when i get a bit of time..

sthaha:
> /hold
> 
> will do this rewrite when i get a bit of time..

IMHO, we can merge this while you work on the improved version.

bobcatfish:
I'm gonna contradict https://github.com/tektoncd/pipeline/issues/1961:

> The disadvantage of this is that git-type PipelineResources become dependent on (presumably) Bash scripts, which can grow to unmaintainable sizes. Unit testing is hard in Bash, but git-init has no unit tests today anyway.

I'm not sure that a script is the most maintainable way forward here? As soon as we get beyond a few lines in a script, whether we're talking about python or git, and when we're talking about a piece of infra that is so core to so many Pipelines (i.e. pulling from git), I feel like we want to start having testing, linting, all that good stuff. Moving this logic into a script feels like a couple steps backward for me?

It definitely has the advantage of being able to see exactly what the script is doing, but I'd rather that we start adding tests and debug info to the git-init logic we already have - even move it into its own repo (or into this one).

/hold

git/git-clone.yaml
chmouel: Pipeline is smart enough to handle properly sub-shells constructs 👼

chmouel: that doesn't trim the spaces: 

image

chmouel:
@bobcatfish that's some fair points here is some of the advantages I see using a pure shell based git clone task is :

- It's  easier for user to debug what's going on instead of having to dig deep into the code, figuring out which git command has been run.

- It's easier for the user to modify, if they need some esoteric way of doing git-clonning they could just modify this task.

- It's easier for distributors to maintain and having one image less.

- There is no dependency between the go code and the task and having to make sure task version X depends on git-init version Y.

- The current git-init image is mostly launching shell commands and not doing much more what a shell script is doing. It's arguably as difficult to read the git-init code than the shell script (or at least it was for me when translating the go code in shell).

And to add this, I am  adding tests for this and we can have a job doing bash unit tests.

vdemeester:
> As soon as we get beyond a few lines in a script, whether we're talking about python or git, and when we're talking about a piece of infra that is so core to so many Pipelines (i.e. pulling from git), I feel like we want to start having testing, linting, all that good stuff. Moving this logic into a script feels like a couple steps backward for me?

I am not entirely sure how "so core" it is to Pipelines. It can be considered as core if it is in `tektoncd/pipeline` as the git `PipelineResource` git. But cloning a repository in a task, is at the same level as any other task that Pipelines are supposed to support, it shouldn't be consider as core as it is just *a way* to populate a `workspace` to be consumed by later `Task`s — at least *this* task shouldn't (but whatever is used by the git `PipelineResource` now or in the future, should).

I do agree (and adding :+1:) with @chmouel on his points, the `git-init` image is more opaque than a git script, and both do the same in the end (aka executing `git` commands).

chmouel:
I have added some tests and resynced with master...

git/git-clone.yaml
sthaha: image
🤔

sthaha: @chmouel are you using sh? or zsh?

sthaha: @chmouel you don't have to change the implementation, I was only curious ...

sbwsg:
@bobcatfish @chmouel @vdemeester can we find a venue to discuss this further and get closure for this PR?  Perhaps the Pipelines WG?  I think there are sound arguments on both sides of this debate and we should probably resolve them so that we can set the policy for other catalog entries going forward.

chmouel:
Sounds good to me, let's have this discussion on next week agenda.

chmouel:
/close

let's try in another way maybe complicated shell scripts is a bit too cryptic, will send a new PR trying it in Python+Tests.

PR #204, opened 2020-02-28
Add a delete param to the git-clone task for reusing PVCs

sbwsg:
# Changes

An implementation detail of the git-clone task in the catalog: it fails
if the directory it's cloning into is already a git repo. This behaviour
has been carried over from the GitResource. Prior to this commit the
advice for users was to manually delete the directory themselves before
running this Task.

This PR introduces a param that, when set to "true", will empty the
directory of any files and folders before performing the clone.

# Submitter Checklist

review them:

### DIFF
--- a/git/git-clone.yaml
+++ b/git/git-clone.yaml
@@ -31,6 +31,10 @@ spec:
       description: subdirectory inside the "output" workspace to clone the git repo into
       type: string
       default: "src"
+    - name: deleteExisting
+      description: clean out the contents of the repo's destination directory (if it already exists) before trying to clone the repo there
+      type: string
+      default: "false"
   results:
   - name: commit
     description: The precise commit SHA that was fetched by this Task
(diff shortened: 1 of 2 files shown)


ekarlso:
lgtm, tested on my pipeline.

git/git-clone.yaml
```diff
@@ -39,14 +43,35 @@ spec:
...
+      cleandir() {
+        # Delete any existing contents of the repo directory if it exists.
+        #
+        # We don't just "rm -rf $CHECKOUT_DIR" because $CHECKOUT_DIR might be "/"
+        # or the root of a mounted volume.
+        if [[ -d "$CHECKOUT_DIR" ]] ; then
+          # Delete non-hidden files and directories
+          rm -rf "$CHECKOUT_DIR"/*
+          # Delete files and directories starting with . but excluding ..
+          rm -rf "$CHECKOUT_DIR"/.[!.]*
```
chmouel: what about just simply : `rm -rf ${CHECKOUT_DIR} && mkdir -p ${CHECKOUT_DIR}` ?

sbwsg: I am not sure what would happen if CHECKOUT_DIR is `/` or if it is a mount point of a volume.  I figured it was safer to wipe out the contents rather than the directory itself.  But I'm still a relative linux noob so if you think it's ok to do that then sounds good to me.

sbwsg:
I've just updated this param to `deleteExisting` instead of `delete-existing` so that it's the same camel case style as the `sslVerify` param

PR #215, opened 2020-03-09
Migrate GitResources to workspace

vdemeester:
# Changes

This migrates use of `GitResources` to workspace in the catalog. 

I know the integration tests will fail because the PVC will be empty… we need to setup something to populate the PVC before running the task :stuck_out_tongue_closed_eyes: 

/cc @chmouel @bobcatfish @sbwsg 

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,24 +42,20 @@ With the `Task` installed, you can define a `TaskRun` that runs that `Task`,
 being sure to provide values for required input parameters and resources:
 
 ```
-apiVersion: tekton.dev/v1alpha1
+apiVersion: tekton.dev/v1beta1
 kind: TaskRun
 metadata:
   name: example-run
 spec:
   taskRef:
     name: golang-build
-  inputs:
-    params:
-    - name: package
-      value: github.com/tektoncd/pipeline
-    resources:
-    - name: source
-      resourceSpec:
-        type: git
-        params:
-        - name: url
-          value: https://github.com/tektoncd/pipeline
+  params:
+  - name: package
+    value: github.com/tektoncd/pipeline
+  workspaces:
+  - name: source
+    persistentVolumeClaim:
+      claimName: my-source
 ```
 
 Next, create the `TaskRun` you defined:
@@ -73,7 +69,7 @@ You can check the status of the `TaskRun` using `kubectl`:
 
 ```
 $ kubectl get taskrun example-run -oyaml
-apiVersion: tekton.dev/v1alpha1
+apiVersion: tekton.dev/v1beta1
 kind: TaskRun
 metadata:
   name: example-run
--- a/buildah/tests/resources.yaml
+++ b/buildah/tests/resources.yaml
@@ -2,21 +2,20 @@
 apiVersion: tekton.dev/v1alpha1
 kind: PipelineResource
 metadata:
-  name: nocode
-spec:
-  type: git
-  params:
-    - name: revision
-      value: master
-    - name: url
-      value: https://github.com/kelseyhightower/nocode
----
-apiVersion: tekton.dev/v1alpha1
-kind: PipelineResource
-metadata:
   name: image
 spec:
   type: image
   params:
     - name: url
       value: localhost:5000/nocode
+---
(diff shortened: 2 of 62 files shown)


buildah/tests/resources.yaml
```diff
@@ -1,22 +1,21 @@
...
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: buildah-source-pvc
+spec:
+  accessModes:
+    - ReadWriteOnce
+  resources:
+    requests:
+      storage: 500Mi
```
chmouel: is it working everywhere? what about CI that doesn't have PV available ? (ie: our downstream CI :D)

test/e2e-common.sh
```diff
@@ -79,6 +79,48 @@ function show_failure() {
     exit 1
 
 }
+
+function pvc_git_clone() {
+    local maxloop=30 # 5 minutes max
+    local pvc_name=$1
+    local remote=$2
+
+    pwd
```
chmouel: i think that's some debugging to remove,

test/e2e-common.sh
```diff
@@ -79,6 +79,48 @@ function show_failure() {
...
+    kubectl -n ${tns} apply -f git/git-clone.yaml
```
chmouel: where is that tns come from? is that a global? Probably not a good idea...

test/e2e-common.sh
```diff
@@ -79,6 +79,48 @@ function show_failure() {
...
+    local cnt=0
+    while true;do
+        [[ ${cnt} == ${maxloop} ]] && show_failure ${testname} ${tns}
+
+        status=$(kubectl get -n ${tns} tr --output=jsonpath='{.items[*].status.conditions[*].status}')
+        reason=$(kubectl get -n ${tns} tr --output=jsonpath='{.items[*].status.conditions[*].reason}')
+        [[ ${status} == *ERROR || ${reason} == *Failed || ${reason} == CouldntGetTask ]] && show_failure ${testname} ${tns}
+        [[ ${status} == True ]] && {
+            echo -n "SUCCESS: ${testname} taskrun has successfully executed: " ;
+            kubectl get pod -o name -n ${tns}|xargs kubectl logs --all-containers -n ${tns}|tail -1
```
chmouel: I know I wrote that but i think it's useless... it show just the same string like "Task Completed"

test/e2e-common.sh
```diff
@@ -79,6 +79,48 @@ function show_failure() {
...
+  - name: url
+    value: ${remote}
+  workspaces:
+  - name: output
+    persistentVolumeClaim:
+      claimName: ${pvc_name}
+EOF
+
+    local cnt=0
+    while true;do
```
chmouel: we can commonalize this loops with `158` as this seems the same code.

README.md
```diff
@@ -49,17 +49,13 @@ metadata:
...
-    - name: package
-      value: github.com/tektoncd/pipeline
-    resources:
-    - name: source
-      resourceSpec:
-        type: git
-        params:
-        - name: url
-          value: https://github.com/tektoncd/pipeline
+  params:
```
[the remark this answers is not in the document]
vdemeester: oh I thought I did :man_facepalming:

buildah/tests/resources.yaml
vdemeester: We are already doing that for `buildpacks` :thinking: So my guess is yes, otherwise, we will need to think of something…

vdemeester:
Finaly green :heart_eyes_cat: 
There is a few task that will be a bit trickier to migrate, one example is the `golang` as they *may* need more involved changes.

vdemeester:
The only remaining is `OpenWhisk` but I want to do that later :sweat_smile:

buildah/tests/resources.yaml
chmouel: Bah i let you fix downstreams ci then 😝

chmouel:
/lgtm

nice work

PR #216, opened 2020-03-11
Port ImageResource to params and results

vdemeester:
# Changes

This starts the work on using `params` for specifying the image reference, and result (when supported) to store the digest.

The `kaniko` is an example with the `results`.

Depends on tektoncd/pipeline#2212

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/kaniko/kaniko.yaml
+++ b/kaniko/kaniko.yaml
@@ -4,6 +4,8 @@ metadata:
   name: kaniko
 spec:
   params:
+  - name: IMAGE
+    description: Name (reference) of the image to build.
   - name: DOCKERFILE
     description: Path to the Dockerfile to build.
     default: ./Dockerfile
@@ -14,13 +16,12 @@ spec:
     default: ""
   - name: BUILDER_IMAGE
     description: The image on which builds will run
-    default: gcr.io/kaniko-project/executor:v0.13.0
+    default: gcr.io/kaniko-project/executor:latest
   workspaces:
   - name: source
-  resources:
-    outputs:
-    - name: image
-      type: image
+  results:
+  - name: IMAGE_DIGEST
+    description: Digest of the image just built.
 
   steps:
   - name: build-and-push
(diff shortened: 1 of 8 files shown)


kaniko/kaniko.yaml
```diff
@@ -35,7 +36,17 @@ spec:
...
       runAsUser: 0
+  - name: write-digest
+    workingDir: $(workspaces.sources.path)
+    image: imagedigestexport-and-jq # FIXME(vdemeester) change with updated image
+    script: |
+      # output of imagedigestexport [{"name":"image","digest":"sha256:eed29..660"}]
+      /ko-app/imagedigestexporter \
+        -images [{"name":"$(params.IMAGE)","type":"image","url":"$(params.IMAGE)","digest":"","OutputImageDir":"$(workspaces.source.path)/image-digest"}] \
+        -terminationMessagePath=/tmp/digest
+      cat  | jq '.[0].digest' | tee /tekton/results/IMAGE_DIGEST
```
sbwsg: Does `jq` add any newlines after the digest string when it's piped like this?  Suggest adding a `tr -d '\n'` if so to make sure the result the controller receives is just the digest value minus whitespace.

sbwsg: also, should this be `cat /tmp/digest`?

kaniko/kaniko.yaml
```diff
@@ -35,7 +36,17 @@ spec:
...
       runAsUser: 0
+  - name: write-digest
+    workingDir: $(workspaces.sources.path)
+    image: imagedigestexport-and-jq # FIXME(vdemeester) change with updated image
+    script: |
+      # output of imagedigestexport [{"name":"image","digest":"sha256:eed29..660"}]
+      /ko-app/imagedigestexporter \
+        -images [{"name":"$(params.IMAGE)","type":"image","url":"$(params.IMAGE)","digest":"","OutputImageDir":"$(workspaces.source.path)/image-digest"}] \
+        -terminationMessagePath=/tmp/digest
+      cat /tmp/digest | jq '.[0].digest' | tee /tekton/results/IMAGE_DIGEST
```
afrittoli: The write-digest step could write the output to a file and the next task that runs alpine:python could publish the digest from the JSON to results.

vdemeester: A task or a step, indeed :wink:

afrittoli: Step, indeed, that's what I meant

kaniko/kaniko.yaml
afrittoli: > Does `jq` add any newlines after the digest string when it's piped like this? Suggest adding a `tr -d '\n'` if so to make sure the result the controller receives is just the digest value minus whitespace.

`jq -c ` should do the trick too

vdemeester:
It is green :tada:

Issue #233, opened 2020-03-20
Task(s) to manage kubernets resources in a cluster


afrittoli:
# Feature description
In the context of MLOps, but most likely also beyond that, it's a common use case to create / update / delete kubernetes resources as part of a Tekton pipeline.

The proposal here is to have a task or set of tasks in the catalog that would define a standard way of doing that in Tekton, which a clear and re-usable interface.

Inputs would be:
1. The resource meta, including type, name (template), namespace, labels and annotations, and the spec which could be an arbitrary JSON/YAML, which we would have to define how it is conveyed to the task
1. The target cluster, which could be a `cluster` PipelineResource or a set of params
1. The operation to be performed - i.e. create, replace, delete

This would be about managing single resources, so it shall not support provisioning the content of a folder or a generic YAML with multiple resources.

This could be implemented as one generic tasks that takes rather generic inputs, which could be usable for all resources, but have a weak validation - or a set of tasks, one per resource type, at least for common resources, which could have more strict checks on input params.

vdemeester:
@afrittoli should this issue take over https://github.com/tektoncd/catalog/issues/37 ? :upside_down_face:

afrittoli:
> #37

Oh, I. missed that one :)

We, we could do that. I guess for deployments we might also have tasks that support provisioning a set of resources at once, so everything that is needed to deploy a service (including deployment, configmap, secrets, ingress etc).

ckadner:
In order to streamline that implementation from a [KFP-Tekton compiler](https://github.com/kubeflow/kfp-tekton/issues/49) perspective, we would like to follow the precedence laid out by Argo's ResourceTemplate, and create a Tekton Task with the following parameters:

- `action`: `['get', 'create', 'apply', 'delete', 'replace', 'patch']` -- the action to perform to the resource
- `merge_strategy`: `['strategic', 'merge', 'json']` -- the strategy used to merge a patch, defaults to "strategic"
- `success_condition`: a label selector expression which describes the `success` condition
- `failure_condition`: a label selector expression which describes the `failure` condition
- `manifest`: the kubernetes manifest

animeshsingh:
@vincent-pli @vdemeester @afrittoli this is a high priority requirement coming from our internal product teams. Would be good to know how can we get this prioritized? I have also discussed these with the OpenShift pipelines team, the relevance and importance of the work effort.

vincent-pli:
[...]

PR #237, opened 2020-03-25
Add Kubectl example

vincent-pli:
Use a image with `kubectl` installed, we can setup `step` of `task` in tekton to create arbitrary rsource of kubernates.

# Changes

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/kubectl/README.md
+++ b/kubectl/README.md
@@ -1,0 +1,51 @@
+# Kubectl
+
+This Task deploys (or delete) a Kubernates resource (pod). It uses
+[`kubectl`](https://kubernetes.io/zh/docs/reference/kubectl/kubectl/) for that.
+
+## Install the Task
+
+```
+kubectl apply -f https://raw.githubusercontent.com/tektoncd/catalog/master/kubectl/kubectl-deploy.yaml
+```
+
+## Install ClusterRolebinding
+
+```
+kubectl apply -f https://raw.githubusercontent.com/tektoncd/catalog/master/kubectl/clusterrolebinding.yaml
+```
+
+## Inputs 
+
+### Parameters
+
+* **manifest:**: The content of the resource to deploy
+
+## Usage
+
+This TaskRun runs the Task to deploy the given Kubernetes resource.
+
+```
+apiVersion: tekton.dev/v1alpha1
+kind: TaskRun
+metadata:
+  name: kubectl-deploy-pod
+spec:
+  taskRef:
+    name: kubectl-deploy-pod
+  inputs:
+    params:
+    - name: manifest
+      value: |
+        apiVersion: v1
+        kind: Pod
+        metadata:
+          name: myapp-pod
+          labels:
+            app: myapp
+        spec:
+          containers:
+          - name: myapp-container
+            image: docker
+            command: ['sh', '-c', 'echo Hello Kubernetes! && sleep 3600']
+```
--- a/kubectl/clusterrolebinding.yaml
+++ b/kubectl/clusterrolebinding.yaml
@@ -1,0 +1,12 @@
+apiVersion: rbac.authorization.k8s.io/v1beta1
+kind: ClusterRoleBinding
+metadata:
+  name: default-admin
+subjects:
+  - kind: ServiceAccount
+    name: default
+    namespace: default
+roleRef:
+  kind: ClusterRole
+  name: cluster-admin
+  apiGroup: rbac.authorization.k8s.io
--- a/kubectl/kubectl-deploy.yaml
+++ b/kubectl/kubectl-deploy.yaml
@@ -1,0 +1,19 @@
+apiVersion: tekton.dev/v1alpha1
+kind: Task
+metadata:
+  name: kubectl-deploy-pod
+  namespace: default
+spec:
+  inputs:
+    params:
+    - name: manifest
+      description: Content of the resource to deploy
+    - name: image
+      default: gcr.io/cloud-builders/kubectl # it is huge
+      description: Kubectl wrapper image
+  steps:
+  - name: kubeconfig
+    image: $(inputs.params.image)
+    script: |
+      echo "$(inputs.params.manifest)" > /tmp/resource.yaml
+      kubectl create -f /tmp/resource.yaml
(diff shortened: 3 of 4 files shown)


Tomcli:
Thanks @vincent-pli, it's a great direction working towards #233. We would like to do some following PRs with this to add more k8s actions and passing results.

kubectl/README.md
```diff
@@ -0,0 +1,51 @@
+# Knative with knctl
```
Tomcli: We should rename it to something like `kubectl` or `Deploy Kubernetes Resource with kubectl`?

ckadner:
Thanks for your PR @vincent-pli. I just took a quick look, but I only saw one parameter `resource`. Ideally we would need 4 more as I described in my comment  above:

https://github.com/tektoncd/catalog/issues/233#issuecomment-602850205

- `action`: `['get', 'create', 'apply', 'delete', 'replace', 'patch']` -- the action to perform to the resource
- `merge_strategy`: `['strategic', 'merge', 'json']` -- the strategy used to merge a patch, defaults to "strategic"
- `success_condition`: a label selector expression which describes the `success` condition
- `failure_condition`: a label selector expression which describes the `failure` condition
- `manifest`: the kubernetes manifest

Assuming your parameter `resource` really is the `manifest`, could we add `action` even if initally it only supports default action `create` or `apply`?

kubectl/clusterrolebinding.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: rbac.authorization.k8s.io/v1beta1
+kind: ClusterRoleBinding
+metadata:
+  name: default-admin
+subjects:
+  - kind: ServiceAccount
+    name: default
+    namespace: default
```
Tomcli: Probably we need to find a better way to map sa in different namespaces

kubectl/kubectl-deploy.yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: tekton.dev/v1alpha1
+kind: Task
+metadata:
+  name: kubectl-deploy-pod
+  namespace: default
```
Tomcli: maybe we should not define static namespace?

kubectl/README.md
ckadner: "Knative"? ...copy-paste artifact? should this say "Kubectl" instead?

kubectl/README.md
```diff
@@ -0,0 +1,51 @@
...
+
+## Inputs
+
+### Parameters
+
+* **resource:**: The content of the resource to deploy
+
+## Usage
+
+This TaskRun runs the Task to deploy the given image as a Knative service.
```
ckadner: is this generic enough to support any K8s resource similar to what Argo supports with `ResourceTemplate` allowing actions like 'get', 'create', 'apply', 'delete', 'replace', 'patch'?

kubectl/README.md
vincent-pli: Ah, you catch it, my bad.

vincent-pli: Ah, again

kubectl/README.md
vincent-pli: I think it could, although it's simple, for example, for `patch`, you could define one more parameter named `resource-name`, and use `kubectl patch $(resource-name) -f `.
for `get`, you could echo the result to a volume and it could be checked by next steps.

kubectl/clusterrolebinding.yaml
vincent-pli: Agree

kubectl/kubectl-deploy.yaml
vincent-pli: Since the `rolebinding` is for the `default` namespace

kubectl/README.md
```diff
@@ -0,0 +1,51 @@
...
```
Tomcli:
```suggestion
This TaskRun runs the Task to deploy the given Kubernetes resource definition.
```

vincent-pli: Done, thanks

kubectl/README.md
```diff
@@ -0,0 +1,51 @@
...
+
+```
+kubectl apply -f https://raw.githubusercontent.com/tektoncd/catalog/master/kubectl/clusterrolebinding.yaml
+```
+
+## Inputs 
+
+### Parameters
+
+* **resource:**: The content of the resource to deploy
```
ckadner: could this be called `manifest` to make it clear that this is not a name or reference but a configuration kind of YAML string?

also, in order to use this in the [KFP-Tekton compiler](https://github.com/kubeflow/kfp-tekton/issues/49) perspective, we would need the following parameters:

- `action`: `['get', 'create', 'apply', 'delete', 'replace', 'patch']` -- the action to perform to the resource
- `merge_strategy`: `['strategic', 'merge', 'json']` -- the strategy used to merge a patch, defaults to "strategic"
- `success_condition`: a label selector expression which describes the `success` condition
- `failure_condition`: a label selector expression which describes the `failure` condition
- `manifest`: the kubernetes manifest

Even though not all of them will be supported yet, can we start out with at least `action` and `manifest` and document that there is a limitation for `action` for now?

I am worried that once this is getting used by a larger group of consumers, it will be hard to make substantial changes to parameters without breaking downstream pipelines

animeshsingh:
@afrittoli @vincent-pli any chance to accelerate this? This is becoming a blocker on our side

kubectl/kubectl-deploy.yaml
```diff
@@ -0,0 +1,16 @@
...
+spec:
+  inputs:
+    params:
+    - name: resource
+      description: Content of the resource to deploy
+  steps:
+  - name: kubeconfig
+    image: gcr.io/cloud-builders/kubectl  # it is huge
```
vdemeester: Should we make this parametrable ?

vincent-pli: Done

kubectl/README.md
vincent-pli: Sorry for delay, involved in a tough project.
The function you mentioned is correct and needed for a complete `kubernete resource hendler` features, it's too complex to just with `script` and `yml`, I think it should be a pre-prepared image.

Whatever, I will dig into and check if i can found something.

vincent-pli:
@vdemeester 
I will port this into v1beta1-wip branch after this one is accepted.

vincent-pli:
Talked with relevant guys, decide to keep the current implements, thanks.

PR #245, opened 2020-03-30
wip: Add back pipelineresources version of Tasks

vdemeester:
# Changes

In order to not break everybody at the same time, this move the Tasks
with workspace in `*-workspace.yaml` files and keep the initial Tasks
with PipelineResources.

@sbwsg @bobcatfish what do you think ? Should we keep them or not ? If we keep them, where should we store them (in this PR using `-workspace.yaml` but it will make the e2e fail) ?

/cc @sbwsg @bobcatfish 

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/buildah/buildah-workspace.yaml
+++ b/buildah/buildah-workspace.yaml
@@ -17,10 +17,9 @@ spec:
   - name: TLSVERIFY
     description: Verify the TLS on the registry endpoint (for push/pull to a non-TLS registry)
     default: "true"
+  workspaces:
+  - name: source
   resources:
-    inputs:
-    - name: source
-      type: git
     outputs:
     - name: image
       type: image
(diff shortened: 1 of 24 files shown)


sbwsg:
Just wanted to document here a proposal I made to @vdemeester via Slack:

I think we should keep the v1beta1-wip branch as-is, removing PipelineResources, and create a branch specifically for the v1alpha1 Tasks.

I also think we should "freeze" `master` and then deprecate it completely, asking all new contributions to go into v1beta1.

Here's a suggested course of action:

- Create a `v1beta1` and `v1alpha1` branch.
- `v1alpha1` will be an alias of where `master` is now.
- `v1beta1` gets all the new workspace-only Tasks.
- After this we ask for all contributions to be based on (and merged into) `v1beta1`.
- `master` and `v1alpha1` will be kept around for archive of alpha only (though perhaps we could add a big message to the README.md of master to point people towards v1beta1?)
- Over time we can deprecate and then phase out the Tasks in `master` completely.

The benefits here will be:

1. Users who are currently applying tasks from master will get those same Tasks until they consciously choose to switch to v1beta1 (or v1alpha1)
2. We create a clear location for permanent archive of the v1alpha1 Tasks w/ PipelineResources.
3. We can announce and then slowly deprecate usage of the `master` branch, eventually using it only to signpost users towards Tasks versioned by branch.

As @vdemeester pointed out via Slack we could at some point make `v1beta1` the default branch that users see when they visit the catalog github repo.

vdemeester:
Alright, let's close this issue and do the following:
- rename `v1beta1-wip` to `v1beta1`
- Update README on `master` and `v1beta1` explaining this (above)
- make `v1beta1` as default branch

PR #256, opened 2020-04-06
Make sure we skip the test if there is no yaml file in there

chmouel:
Test would be failing if there is a directory but no yaml template in there,
which in reality we should not have this happening but it happens sometime.

let's make sure we skip the test if there is no yaml files then.

# Changes

# Submitter Checklist

review them:

for more details._


sbwsg:
@chmouel this seems like a sensible change.  Do you think we still need the hold?

PR #258, opened 2020-04-06
Task for creating GitHub deployments

jace-ys:
# Changes

This introduces a Task for creating GitHub deployments for a given
repository. This allows people who rely on GitHub deployment events
for their CD systems to be able create them from within a Tekton
pipeline.

This task is pretty much just a wrapper around the GitHub API for
creating deployments. The main params required are the ref
and environment for the deployment, and we fall back to the API's
defaults for the other params.

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/task/github-create-deployment/0.1/README.md
+++ b/task/github-create-deployment/0.1/README.md
@@ -1,0 +1,66 @@
+## GitHub Create Deployment
+
+The `github-create-deployment` Task lets you create a GitHub deployment.
+
+See GitHub's deployment API on [creating a deployment](https://developer.github.com/v3/repos/deployments/#create-a-deployment) for more information.
+
+### Install the Task
+
+```
+kubectl apply -f https://raw.githubusercontent.com/tektoncd/catalog/master/task/github-create-deployment/0.1/github-create-deployment.yaml
+```
+
+### Secrets
+
+This Task requires access to a GitHub token set via a Kubernetes Secret. By default, the name of this Secret should be `github` and the secret key should be `token`, but you can configure this via the `GITHUB_TOKEN_SECRET_NAME` and `GITHUB_TOKEN_SECRET_KEY` [parameters](#parameters) described below.
+
+To create such a Secret via `kubectl`:
+
+```
+kubectl create secret generic github --from-literal token="MY_TOKEN"
+```
+
+See GitHub's documentation on [Understanding scopes for OAuth Apps](https://developer.github.com/apps/building-oauth-apps/understanding-scopes-for-oauth-apps/) to figure out what scopes you need to give to this token to create deployments.
+
+### Parameters
+
+- **GITHUB_HOST_URL:**: The GitHub host domain. (_default:_ `api.github.com`)
+- **REPO_FULL_NAME:**: The GitHub repository full name. (_e.g:_ `tektoncd/catalog`)
+- **REF**: The ref to deploy. This can be a branch, tag, or SHA. (_e.g:_ `master`)
+- **TASK**: Specifies a task to execute. (_default:_ `deploy`)
+- **AUTO_MERGE**: Attempts to automatically merge the default branch into the
+  requested ref, if it's behind the default branch. (_default:_ `true`)
+- **REQUIRED_CONTEXTS**: The status contexts to verify against commit status checks. To
+  verify all unique contexts before creating a deployment, pass
+  in `*`. To bypass checking entirely, pass an empty string. (_default:_ `*`)
+- **PAYLOAD**: JSON payload with extra information about the deployment. (_default:_ `""`)
+- **ENVIRONMENT**: Name for the target deployment environment (_e.g:_ `production`).
+- **DESCRIPTION**: Short description of the deployment. (_default:_ `""`).
+- **GITHUB_TOKEN_SECRET_NAME**: The name of the Kubernetes Secret that
+  contains the GitHub token. (_default:_ `github`).
+- **GITHUB_TOKEN_SECRET_KEY**: The key within the Kubernetes Secret that contains the GitHub token. (_default:_ `token`).
+
+## Usage
+
+This TaskRun creates a GitHub deployment for the given repository.
+
+```yaml
+---
+apiVersion: tekton.dev/v1beta1
+kind: TaskRun
+metadata:
+  labels:
+    tekton.dev/task: github-create-deployment
+  name: github-create-tektoncd-catalog-deployment
+spec:
+  taskRef:
+    kind: Task
+    name: github-create-deployment
+  params:
+    - name: REPO_FULL_NAME
+      value: tektoncd/catalog
+    - name: REF
+      value: master
+    - name: ENVIRONMENT
+      value: staging
+```
--- a/task/github-create-deployment/0.1/github-create-deployment.yaml
+++ b/task/github-create-deployment/0.1/github-create-deployment.yaml
@@ -1,0 +1,155 @@
+---
+apiVersion: tekton.dev/v1beta1
+kind: Task
+metadata:
+  name: github-create-deployment
+  labels:
+    app.kubernetes.io/version: "0.1"
+  annotations:
+    tekton.dev/pipelines.minVersion: "0.12.1"
+    tekton.dev/tags: github
+    tekton.dev/displayName: "create github deployment"
+spec:
+  description: >-
+    This Task will create a GitHub deployment for a repository.
+
+  params:
+    - name: GITHUB_HOST_URL
+      description: |
+        The GitHub host, adjust this if you run a GitHub enterprise.
+      default: "api.github.com"
+      type: string
+
+    - name: REPO_FULL_NAME
+      description: |
+        The GitHub repository full name, i.e: tektoncd/catalog.
+      type: string
+
+    - name: REF
+      description: |
+        The ref to deploy. This can be a branch, tag, or SHA.
+      type: string
+
+    - name: TASK
+      description: |
+        Specifies a task to execute (e.g., deploy or deploy:migrations).
+      default: "deploy"
+      type: string
+
+    - name: AUTO_MERGE
+      description: |
+        Attempts to automatically merge the default branch into the
+        requested ref, if it's behind the default branch.
+      default: "true"
+      type: string
+
+    - name: REQUIRED_CONTEXTS
+      description: |
+        The status contexts to verify against commit status checks. To
+        verify all unique contexts before creating a deployment, pass
+        in "*". To bypass checking entirely, pass an empty string.
+      default: "*"
+
+    - name: PAYLOAD
+      description: |
+        JSON payload with extra information about the deployment.
+      default: ""
+      type: string
+
+    - name: ENVIRONMENT
+      description: |
+        Name for the target deployment environment (e.g., production, staging).
+      type: string
+
+    - name: DESCRIPTION
+      description: |
+        Short description of the deployment.
+      default: ""
+      type: string
+
+    - name: GITHUB_TOKEN_SECRET_NAME
+      description: |
+        The name of the Kubernetes Secret that contains the GitHub token.
+      default: "github"
+      type: string
+
+    - name: GITHUB_TOKEN_SECRET_KEY
+      description: |
+        The key within the Kubernetes Secret that contains the GitHub token.
+      default: "token"
+      type: string
+
+  steps:
+    - name: create-deployment
+      env:
+        - name: GITHUBTOKEN
+          valueFrom:
+            secretKeyRef:
+              name: $(params.GITHUB_TOKEN_SECRET_NAME)
+              key: $(params.GITHUB_TOKEN_SECRET_KEY)
+
+      image: registry.access.redhat.com/ubi8/ubi:8.2
+      script: |
+        #!/usr/libexec/platform-python
+        import http.client
+        import json
+        import os
+        import sys
+
+        deployment_url = "/repos/$(params.REPO_FULL_NAME)/deployments"
+
+        if "$(params.REQUIRED_CONTEXTS)" == "*":
+            required_contexts = None
+        elif "$(params.REQUIRED_CONTEXTS)" == "":
+            required_contexts = []
+        else:
+            required_contexts = "$(params.REQUIRED_CONTEXTS)".split(",")
+
+        # payload should either be a JSON object or an empty string
+        if "$(params.PAYLOAD)":
+            payload = json.loads("$(params.PAYLOAD)")
+        else:
+            payload = ""
+
+        data = {
+            "ref": "$(params.REF)",
+            "task": "$(params.TASK)",
+            "auto_merge": json.loads("$(params.AUTO_MERGE)"),
+            "required_contexts": required_contexts,
+            "payload": payload,
+            "environment": "$(params.ENVIRONMENT)",
+            "description": "$(params.DESCRIPTION)",
+        }
+
+        print("Sending this data to GitHub: ")
+        print(data)
+
+        conn = http.client.HTTPSConnection("$(params.GITHUB_HOST_URL)")
+        r = conn.request(
+            "POST",
+            deployment_url,
+            body=json.dumps(data),
+            headers={
+                "User-Agent": "TektonCD, the peaceful cat",
+                "Authorization": "Bearer " + os.environ["GITHUBTOKEN"],
+            })
+        resp = conn.getresponse()
+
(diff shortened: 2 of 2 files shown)


github/README.md
```diff
@@ -44,7 +44,7 @@ kubectl apply -f https://raw.githubusercontent.com/tektoncd/catalog/master/githu
 * **STATE:** The state of the status. Can be one of the following `error`,
   `failure`, `pending`, or `success`.
 
-## Usage
+### Usage
```
jace-ys, line 47: Not sure if I should be changing this, but it improves readability of the markdown for me.

github/create_deployment.yaml
```diff
@@ -0,0 +1,111 @@
...
+        The ref to deploy. This can be a branch, tag, or SHA.
+    - name: TASK
+      description: |
+        Specifies a task to execute (e.g., deploy or deploy:migrations).
+      default: "deploy"
+    - name: AUTO_MERGE
+      description: |
+        Attempts to automatically merge the default branch into the
+        requested ref, if it's behind the default branch.
+      default: "true"
```
jace-ys, line 28: Would be great if boolean was a supported type so that we don't have to do a json.loads() further below to coerce this into a boolean that the API expects. This is my workaround for now.

github/create_deployment.yaml
```diff
@@ -0,0 +1,111 @@
...
+    - name: REPO_FULL_NAME
+      description: |
+        The GitHub repository full name, i.e: tektoncd/catalog.
+    - name: REF
+      description: |
+        The ref to deploy. This can be a branch, tag, or SHA.
+    - name: TASK
+      description: |
+        Specifies a task to execute (e.g., deploy or deploy:migrations).
+      default: "deploy"
```
jace-ys, line 23: I noticed that the other github tasks don't really provide default values, but in the contributor's guidelines [it recommends to provide as many default values](https://github.com/tektoncd/catalog/blob/master/CONTRIBUTING.md#technical-requirements). I'm just falling back to the GitHub API's default values for most of these.

github/create_deployment.yaml
```diff
@@ -0,0 +1,111 @@
...
+        Short description of the deployment.
+      default: ""
+  steps:
+    - name: create-deployment
+      env:
+        - name: GITHUBTOKEN
+          valueFrom:
+            secretKeyRef:
+              name: github
+              key: token
```
jace-ys, line 53: I'm wondering if the secret name and key should be parameterised instead of hardcoded (for this and the other github tasks). This will make it so we don't prescribe names for the secret for users of this task.

github/create_deployment.yaml
```diff
@@ -0,0 +1,111 @@
...
+        import http.client
+
+        deployment_url = "/repos/$(params.REPO_FULL_NAME)/deployments"
+
+        if "$(params.REQUIRED_CONTEXTS)" == "*":
+            required_contexts = None
+        elif "$(params.REQUIRED_CONTEXTS)" == "":
+            required_contexts = []
+        else:
+            required_contexts = "$(params.REQUIRED_CONTEXTS)".split(",")
```
jace-ys, line 69: This is a little bit janky, as there are 3 cases for the `required_contexts` provided by the GitHub API itself.

- You want to verify all unique contexts (this is the default in the API, and requires omitting the field)
- You want to bypass checks entirely (this requires providing an empty array)
- You provide an array of status contexts you want to verify

So in the code:

- Use `*` to assign `None` as a way of omitting the field by pruning it later
- Use empty string as an alias for an empty array, otherwise
- Create a comma-separated list from a string (eg. `ci/circleci,cla/google`)

Open to any better ideas here as this was the best I could think of!

github/create_deployment.yaml
```diff
@@ -0,0 +1,111 @@
...
+        data = {
+            "ref": "$(params.REF)",
+            "task": "$(params.TASK)",
+            "auto_merge": json.loads("$(params.AUTO_MERGE)"),
+            "required_contexts": required_contexts,
+            "payload": "$(params.PAYLOAD)",
+            "environment": "$(params.ENVIRONMENT)",
+            "description": "$(params.DESCRIPTION)" if "$(params.DESCRIPTION)" else None,
+        }
+        data = { k: v for k, v in data.items() if v is not None }
```
jace-ys, line 80: Prune `None` fields to fallback to GitHub's default values when omitted.

github/create_deployment.yaml
```diff
@@ -0,0 +1,111 @@
...
+            exit(1)
+        else:
+            body = json.loads(resp.read().decode())
+            print("GitHub deployment created for "
+                  "$(params.REPO_FULL_NAME): "
+                  "id={} branch={} sha={} environment={}".format(
+                      body["id"],
+                      body["ref"],
+                      body["sha"],
+                      body["environment"]))
```
jace-ys, line 111: I'm not sure if this is the best way to be logging some useful data from the response. Any suggestions?

jace-ys:
@googlebot I signed it!

github/create_deployment.yaml
tragiclifestories: Is this the only way to get the params into the script? Would it be possible to have 
```yaml
env:
  - name: REQUIRED_CONTEXTS
     value: "$(params.REQUIRED_CONTEXTS)"
```

And then in the script you could have

```python
context_string = os.getenv("REQUIRED_CONTEXTS")

if context_string == "*":
    # ... 
    pass
```
Interpolating like this feels kind of error prone.

tragiclifestories: On your q, this is a problem with the github API, so I don't see a 'nice' way to do it. You could pull this if/else stuff into a helper function but that's about it.

github/create_deployment.yaml
tragiclifestories: I'd be interested to see if we hit the same problem we have in the sentry logging ...

github/create_deployment.yaml
jace-ys: I think using the `env` is an option, but it seems more like we're side-stepping the issue instead of actually addressing it. Plus I guess it's less explicit to do it via the `env` cause you don't get validation from the pipelines controller if you miss out on the parameter when running the pipeline.

github/create_deployment.yaml
chmouel: I don't think there is other way unfortunately but i think it's good enough as seen in the task logs output...

github/create_deployment.yaml
chmouel: I think it's fine by me or at least for the first iteration and improve if needed....

vdemeester:
@jace-ys  switched the base branch to master as it got merged with `v1beta1` now :angel:

github/create_deployment.yaml
```diff
@@ -0,0 +1,111 @@
+---
+apiVersion: tekton.dev/v1beta1
+kind: Task
+metadata:
+  name: github-create-deployment
+  description: |
+    This Task will create a GitHub deployment for a repository.
```
SM43, line 7: It would be great if you could add the necessary metadata (see below) so that Tekton Hub and other supporting tools (in development) is able to search the task.

  ```yaml
  metadata:
    labels:
      app.kubernetes.io/version: "0.1"
    annotations:
      tekton.dev/pipelines.minVersion: "0.12.1"   
      tekton.dev/tags: <tags associated>
      tekton.dev/displayName: An Optional name for the task 

  spec:
     description: >-
       One line summary of the task. 

       Description of the task 
  ```
  
  see: [#389](https://github.com/tektoncd/catalog/pull/389) | and [Tekton Hub Preview](https://ui-tekton-hub.apps.openshift-web.p0s5.p1.openshiftapps.com/)

SM43:
- It would be nice if you could add the task with all the files in the task directory 

```
  ./task/         👈 the kind of the resource 
      /argocd       👈 definition file must have same name
        /0.1
          /README.md     👈 [optional] since there can be a fallback
          /argocd.yaml   👈 the file name must be resource name
        /0.2/...
      /golang-build
        /0.1
          /README.md
          /golang-build.yaml
```
Please see : [#389](https://github.com/tektoncd/catalog/pull/389)

jace-ys:
Thanks @SM43 for the heads-up. I have done this now and rebased again latest master 🙂

task/github-create-deployment/0.1/README.md
```diff
@@ -0,0 +1,66 @@
...
+## Create a GitHub deployment
+
+The `github-create-deployment` task lets you create a GitHub deployment.
+
+See GitHub's deployment API on [creating a deployment](https://developer.github.com/v3/repos/deployments/#create-a-deployment) for more information.
+
+### Install the Task
+
+```
+kubectl apply -f https://raw.githubusercontent.com/tektoncd/catalog/master/github/create_deployment.yaml
```
vinamra28, line 25:
```suggestion
kubectl apply -f https://raw.githubusercontent.com/tektoncd/catalog/master/task/github-create-deployment/0.1/github-create-deployment.yaml
```

jace-ys: Thanks! Totally missed that.

task/github-create-deployment/0.1/README.md
```diff
@@ -0,0 +1,66 @@
+# GitHub
+
+A collection of tasks to help working with the [GitHub
+API](https://developer.github.com/v3/).
+
+## GitHub token
+
+Most tasks would expect to have a secret set in the kubernetes secret `github`
```
chmouel, line 8:
```suggestion
This task expect to have a secret set in the kubernetes secret `github`
```

task/github-create-deployment/0.1/github-create-deployment.yaml
```diff
@@ -0,0 +1,127 @@
...
+        resp = conn.getresponse()
+
+        # only look for 201 statuses to signify a deployment was successfully
+        # created, as a merged branch response returns a 202 status but does not
+        # actually create the deployment.
+        # https://developer.github.com/v3/repos/deployments/#merged-branch-response
+        if resp.status != 201:
+            print("Error: %d" % (resp.status))
+            print(resp.read())
+            exit(1)
```
chmouel, line 118: `import sys` at the top and :
```suggestion
            sys.exit(1)
```

task/github-create-deployment/0.1/github-create-deployment.yaml
```diff
@@ -0,0 +1,127 @@
...
+        else:
+            body = json.loads(resp.read().decode())
+            print("GitHub deployment created for "
+                  "$(params.REPO_FULL_NAME): "
+                  "id={} branch={} sha={} environment={}".format(
+                      body["id"],
+                      body["ref"],
+                      body["sha"],
+                      body["environment"]))
```
chmouel, line 127: what about adding a result?

task/github-create-deployment/0.1/README.md
chmouel: (probably need to fix the wording of the other github tasks* who got moved to their own directory)

task/github-create-deployment/0.1/github-create-deployment.yaml
jace-ys: Good catch!

task/github-create-deployment/0.1/github-create-deployment.yaml
jace-ys: I'm not very sure what's the point of adding a result value since if it gets to this point, the API call would have succeeded.

I added a `url` value instead to indicate the successful creation of the deployment. Let me know what you think?

task/github-create-deployment/0.1/README.md
```diff
@@ -0,0 +1,66 @@
+# GitHub
+
+A collection of tasks to help working with the [GitHub
+API](https://developer.github.com/v3/).
```
vinamra28, line 4: I guess we can remove this now as every task has it's own README

task/github-create-deployment/0.1/README.md
```diff
@@ -0,0 +1,66 @@
...
+
+```
+kubectl create secret generic github --from-literal token="MY_TOKEN"
+```
+
+## Create a GitHub deployment
+
+The `github-create-deployment` task lets you create a GitHub deployment.
+
+See GitHub's deployment API on [creating a deployment](https://developer.github.com/v3/repos/deployments/#create-a-deployment) for more information.
```
vinamra28, line 20: Probably we can move this to the starting of README file

task/github-create-deployment/0.1/github-create-deployment.yaml
```diff
@@ -0,0 +1,129 @@
...
+  steps:
+    - name: create-deployment
+      env:
+        - name: GITHUBTOKEN
+          valueFrom:
+            secretKeyRef:
+              name: github
+              key: token
+
+      image: registry.access.redhat.com/ubi8/ubi:latest
```
vinamra28, line 71:
```suggestion
      image: registry.access.redhat.com/ubi8/ubi:8.2
```

task/github-create-deployment/0.1/github-create-deployment.yaml
```diff
@@ -0,0 +1,129 @@
...
+      default: ""
+
+  steps:
+    - name: create-deployment
+      env:
+        - name: GITHUBTOKEN
+          valueFrom:
+            secretKeyRef:
+              name: github
+              key: token
```
Divyansh42, line 69: Probably we can make names of `secret name` and `secret key` more generic like [this](https://github.com/tektoncd/catalog/blob/master/task/github-set-status/0.2/github-set-status.yaml#L89).

task/github-create-deployment/0.1/github-create-deployment.yaml
jace-ys: I see that the other github tasks still use `latest` but I guess that they should all be pinned as well?

jace-ys:
Hi @vinamra28 and @Divyansh42,

Thanks for your feedback! I have taken them into account and adjusted the PR accordingly.

Let me know what you think 😄

Divyansh42 (commented):
Looks good to me!
/cc @chmouel

task/github-create-deployment/0.1/github-create-deployment.yaml
vinamra28: There is a PR #519 which will pin the version of all the tasks :)

task/github-create-deployment/0.1/github-create-deployment.yaml
```diff
@@ -0,0 +1,149 @@
...
+        data = {
+          "ref": "$(params.REF)",
+          "task": "$(params.TASK)",
+          "auto_merge": json.loads("$(params.AUTO_MERGE)"),
+          "required_contexts": required_contexts,
+          "payload": json.loads("$(params.PAYLOAD)") if "$(params.PAYLOAD)" else None,
+          "environment": "$(params.ENVIRONMENT)",
+          "description": "$(params.DESCRIPTION)" if "$(params.DESCRIPTION)" else None,
+          }
+        data = { k: v for k, v in data.items() if v is not None }
```
chmouel, line 117: are you sure it's is not None you want ? you want to check if the string is empty isnt it ? 

I am actually not sure what this line does ? have you copied from somewhere else ?

jace-ys: So what this line does is prune keys from our data payload if its value is None. The reason this is needed is because GitHub will reject a payload with empty string values, eg. `{ description: "" }`.

This line essentially translates into
```python
for k, v in data.items():
    if v is not None:
        data[k] = v
```

chmouel: sounds good 👍

jace-ys: So it seems my earlier statement was incorrect 🤦 (been quite awhile since I did this so context has been lost)

Turns out we don't need to prune empty string values from the payload - my previous implementation was to ensure we can do `json.loads()` correctly on `$(params.PAYLOAD)`, as `json.loads("")` raises an error. I've now modified the script to reflect this and it works well from testing.

Previously:
- Set to `None` if empty string, else `json.loads()`
- Prune None values

Currently:
- If not empty string, do `json.loads()`
- Else leave as empty string

Have also added a comment to explain this: https://github.com/tektoncd/catalog/pull/258/files#diff-ba3bbc0b3249e2e2fce3c0366a4b451dd81de4b117cbd95b9ccb64a01a263659R108

Sorry for the confusion and big roundabout.. 🙈 Hope this makes more sense?

chmouel (approved):

jace-ys:
thanks all 🙌

PR #273, opened 2020-04-13
Add batch merge and refspec support to git task

ad22:
# Changes

Add refspec support to git task corresponding to https://github.com/tektoncd/pipeline/pull/2320

Add a new batch merge task (https://github.com/tektoncd/pipeline/issues/508) that re-uses
the git-init container and subsequently performs the requested batch merge mode on top of
the checked out revision

# Submitter Checklist

review them:

### DIFF
--- a/git/README.md
+++ b/git/README.md
@@ -31,6 +31,7 @@ as well as
 
 * **url**: git url to clone (_required_)
 * **revision**: git revision to checkout (branch, tag, sha, ref…) (_default:_ master)
+* **refspec**: git refspec to fetch before checking out revision (_default_:refs/heads/master:refs/heads/master)
 * **submodules**: defines if the resource should initialize and fetch the submodules (_default_: true)
 * **depth**: performs a shallow clone where only the most recent commit(s) will be fetched (_default_: 1)
 * **sslVerify**: defines if http.sslVerify should be set to true or false in the global git config (_default_: true)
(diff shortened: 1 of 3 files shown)


sbwsg:
Sorry for the extra work but would it be possible to split this into two PRs?  One for the refspec flag and another for the batch merge task?

The main reason for this is that the `refspec` flag won't be available from the `git-init` binary until your PR (https://github.com/tektoncd/pipeline/pull/2320) is merged and a new Pipelines release is cut.  If we merged this as-is then anyone `kubectl apply`ing the `git-clone` Task would end up with breakages because the `git-init` binary would complain about the missing flag: `flag provided but not defined: -refspec`. 

So ideally I'd like to be able to review and merge the batchmerge PR in isolation.

ad22:
@sbwsg The batch merge functionality depends on the refspec feature being available, which is why I created them together. I can see the value of opening another PR for the batch merge, regardless, just because it addresses a separate requested feature.

Both features can wait till https://github.com/tektoncd/pipeline/pull/2320 is merged and released. Feel free to add the appropriate labels on this PR.

ad22:
Added tests, additional git merge/cherry-pick options and switched docker image to git-init:v0.12.0. Still waiting on the corresponding pipeline release, do not merge.

git/README.md
```diff
@@ -135,3 +136,33 @@ spec:
     persistentVolumeClaim:
       claimName: workspace-pvc
 ```
+
+## `git-batch-merge`
```
sbwsg, line 140: One minor suggestion here: it might be a good idea to explain the intended usage of this Task.  What's an example use case where this is useful?

git/git-batch-merge.yaml
```diff
@@ -0,0 +1,117 @@
...
+  workspaces:
+    - name: output
+      description: The git repo will be cloned onto the volume backing this workspace
+  params:
+    - name: url
+      description: git url to clone
+      type: string
+    - name: revision
+      description: base git revision to checkout (branch, tag, sha, ref…)
+      default: master
```
sbwsg, line 15: suggest adding `type: string` in all the params where this is missing

sbwsg:
@ad22 I've removed the hold label since 0.12 has been released.  Now's a good time to rebase on top of current `v1beta1` branch and address feedback if you'd still like to pursue this PR.  Thanks for your patience!

git/README.md
ad22: Added a description

git/git-batch-merge.yaml
ad22: Done

ad22:
@sbwsg I've updated the PR. Its really a small wrapper around git-init. Please review the updated description, tektoncd/pipeline#508 and accept this if the you find it relevant. Thanks!

git/README.md
sbwsg: Excellent, thank you!

git/README.md
```diff
@@ -135,3 +136,44 @@ spec:
...
+  &nbsp;&nbsp;`merge`: corresponds to git merge -s recursive. This is the default mode used by github <br>
+  &nbsp;&nbsp;`cherry-pick`: corresponds to git cherry-pick <br>
+  See [git-merge][git-merge] and [git-cherry-pick][git-cherry-pick]
+* **gitUserName**: git user name to use for creating the batched commit (First Last)
+    (_default_: GitBatch Task). See [git-user-config][git-user-config]
+* **gitUserEmail**: git user email to use for creating the batched commit (First.Last@domain.com)
+  (_default_: GitBatch.Task@tekton.dev). See [git-user-config][git-user-config]
+
+There is 1 additional result in addition to the one(s) mentioned for the git-clone task
+* **tree**: The [git tree][git-tree] object SHA1 that was created after batch merging the refs on HEAD.
```
sbwsg, line 171: Suggest just listing all the results this task emits, rather than referring to the results of another task.  This is so the user doesn't have to go looking for the results of another task to figure out what this one emits.  And there might come a time when either a) the git-clone task emits more results that this one does or b) we split these tasks into separate directories / readmes.

PR #277, opened 2020-04-14
git-clone: fix git-init flags

### LINKED ISSUE #275: git-clone: some param values are inadvertently ignored
# Expected Behavior

Using the `git-clone` task, with the following params:

```yaml
- name: sslVerify
  value: "false"
- name: submodules
  value: "false"
- name: depth
  value: "2"
```

The task should not fail if the underlying git repository's https server is
using an untrusted self-signed certificate, it should not attempt to fetch
submodules, and it should fetch 2 commits.

# Actual Behavior

None of the above param values take effect - the default values are used.
The task will fail if there is an untrusted certificate in the chain. If it is a
trusted certificate, submodules will still be fetched and only a single commit
will be pulled.

It is difficult to setup a repro for `sslVerify` but the underlying issue is the
same. The following will repro the issues with `submodules` and `depth`.

Assuming pipelines are installed and the git-clone Task exists, create the
following pipeline:

```yaml
apiVersion: tekton.dev/v1beta1
kind: Task
metadata:
  name: git-log
spec:
  workspaces:
    - name: source
      mountPath: /source
  params:
    - name: subdirectory
      description: Subdirectory inside "source" workspace that contains the README.md.
      default: "."
  steps:
    - name: git-log
      image: alpine/git
      script: cd "$(workspaces.source.path)/$(params.subdirectory)"; git log --oneline
---
apiVersion: tekton.dev/v1beta1
kind: Pipeline
metadata:
  name: git-log-pipeline
spec:
  workspaces:
    - name: shared-workspace
      description: The git repo will be cloned into this workspace and git log will be run
  tasks:
    - name: fetch-repository
      taskRef:
        name: git-clone
      workspaces:
        - name: output
          workspace: shared-workspace
      params:
        - name: url
          value: https://github.com/tektoncd/pipeline.git
        - name: subdirectory
          value: pipeline-checkout
        - name: deleteExisting
          value: "true"
        - name: submodules
          value: "false"
        - name: depth
          value: "2"
    - name: git-log
      taskRef:
        name: git-log
      runAfter:
        - fetch-repository # required to ensure clone occurs before reading
      workspaces:
        - name: source
          workspace: shared-workspace
      params:
        - name: subdirectory
          value: pipeline-checkout
```

And then run it with:

```yaml
kind: PersistentVolumeClaim
apiVersion: v1
metadata:
  name: workspace-pvc
spec:
  accessModes:
    - ReadWriteOnce
  resources:
    requests:
      storage: 500Mi
---
apiVersion: tekton.dev/v1beta1
kind: PipelineRun
metadata:
  name: git-log-pipeline-run
spec:
  pipelineRef:
    name: git-log-pipeline
  workspaces:
    - name: shared-workspace
      persistentVolumeClaim:
        claimName: workspace-pvc
```

In the logs for the `step-clone`, you will still see:

```sh
{"level":"info","ts":1586897972.7579875,"caller":"git/git.go:133","msg":"Successfully initialized and updated submodules in path /workspace/output/pipeline-checkout"}
...
```

And the `step-git-log` logs will show just a single commit:

```sh
git log --oneline
fe5981e Add volumeClaimTemplate as a Workspace volume source
```

# Additional Info

This appears to just be an issue with how the [bool and int flags](https://gist.github.com/ryanewk/eed221119a0b0366b0f13db5a587873a) are passed to git-init.
Right now, the command being executed looks like this:

```sh

/ko-app/git-init -url https://github.com/tektoncd/pipeline.git -revision master -path /workspace/output/pipeline-checkout -sslVerify=true -submodules=false -depth=2
```

If we change it to:

```sh
/ko-app/git-init -url https://github.com/tektoncd/pipeline.git -revision master -path /workspace/output/pipeline-checkout -sslVerify=true -submodules=false -depth=2
```

We get the expected behavior.

ryanewk:
fixes #275

# Changes

Use flag=val for bool and int flags passed to git-init

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/git/git-clone.yaml
+++ b/git/git-clone.yaml
@@ -66,9 +66,9 @@ spec:
         -url "$(params.url)" \
         -revision "$(params.revision)" \
         -path "$CHECKOUT_DIR" \
-        -sslVerify "$(params.sslVerify)" \
-        -submodules "$(params.submodules)" \
-        -depth "$(params.depth)"
+        -sslVerify="$(params.sslVerify)" \
+        -submodules="$(params.submodules)" \
+        -depth="$(params.depth)"
       cd "$CHECKOUT_DIR"
       RESULT_SHA="$(git rev-parse HEAD | tr -d '\n')"
       EXIT_CODE="$?"


ryanewk:
@googlebot I signed it!

chmouel:
/lgtm
/approve

Thanks 👍🏼🔥

chmouel:
I think I merged too quickly @ryanewk ! (but that doesn't do any harms)

I don't think your patch would change anything, using `=` or not in arguments for go flags would do the same thing.

Are you sure this patch fix your problem ?

sbwsg:
This is interesting, I had to look it up.  Turns out `go` does require `=` for bool flags (https://golang.org/pkg/flag/#hdr-Command_line_flag_syntax)

Doesn't mention int flags though.  So I wrote a test program:

```go
package main

import (
  "flag"
  "log"
)

func main() {
  var b bool
  var u uint
  flag.BoolVar(&b, "b", true, "bool")
  flag.UintVar(&u, "u", 99, "uint")
  flag.Parse()

  log.Printf("b:%b u:%d", b, u)
}
```

Output:

```bash
$ ./ft -b "false" -u "100"
b:%!b(bool=true) u:99

./ft -b="false" -u "100"
b:%!b(bool=false) u:100
```

So if the bool flag fails to parse then the rest of the flags get ignored :(  That's some nasty surprising behaviour and a good argument for ramping up testing in the catalog.

chmouel:
This is indeed surprising, thanks for digging this up..

I had some tests added part of my shell rewrite here https://github.com/tektoncd/catalog/pull/188/files#diff-952ab5bb59fcb353a880d9acd3843f59

if we are not going to use the rewrite maybe we can use the tests?

ryanewk:
thanks, all. and sorry it was kind of hidden in the issue but I had linked to a test program that demonstrated it too: https://gist.github.com/ryanewk/eed221119a0b0366b0f13db5a587873a. The behavior was surprising to me as well.

sbwsg:
> if we are not going to use the rewrite maybe we can use the tests?

Great idea.  I still think we should have a conversation about whether `git-clone` should be a plain script.  I think there are good arguments on both sides.  But having a separate PR just for the tests is a great idea.

PR #278, opened 2020-04-15
Add Kubectl sample, porting from version "v1alpha1"

vincent-pli:
Fix issue: #233 , Porting from `v1alpha1`

Use a image with `kubectl` installed, we can setup `step` of `task` in tekton to create arbitrary rsource of kubernates.

# Changes

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/kubectl/clusterrolebinding.yaml
+++ b/kubectl/clusterrolebinding.yaml
@@ -1,0 +1,12 @@
+apiVersion: rbac.authorization.k8s.io/v1beta1
+kind: ClusterRoleBinding
+metadata:
+  name: default-admin
+subjects:
+  - kind: ServiceAccount
+    name: default
+    namespace: default
+roleRef:
+  kind: ClusterRole
+  name: cluster-admin
+  apiGroup: rbac.authorization.k8s.io
--- a/kubectl/kubectl-deploy.yaml
+++ b/kubectl/kubectl-deploy.yaml
@@ -1,0 +1,18 @@
+apiVersion: tekton.dev/v1beta1
+kind: Task
+metadata:
+  name: kubectl-deploy-pod
+  namespace: default
+spec:
+  params:
+  - name: manifest
+    description: Content of the resource to deploy
+  - name: image
+    default: gcr.io/cloud-builders/kubectl # it is huge
+    description: Kubectl wrapper image
+  steps:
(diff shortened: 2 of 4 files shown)


kubectl/kubectl-deploy.yaml
```diff
@@ -0,0 +1,18 @@
...
+spec:
+  params:
+  - name: manifest
+    description: Content of the resource to deploy
+  - name: image
+    default: gcr.io/cloud-builders/kubectl # it is huge
+    description: Kubectl wrapper image
+  steps:
+  - name: kubeconfig
+    image: $(inputs.params.image)
```
vdemeester, line 15: This needs to a `$(params.image)` instead :wink:

kubectl/kubectl-deploy.yaml
```diff
@@ -0,0 +1,18 @@
...
+    script: |
+      echo "$(inputs.params.manifest)" > /tmp/resource.yaml
```
vdemeester, line 17: `$(params.manifest)` here

kubectl/kubectl-deploy.yaml
vincent-pli: oh, my mistake, thanks 👍

kubectl/kubectl-deploy.yaml
vincent-pli: Done

kubectl/clusterrolebinding.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: rbac.authorization.k8s.io/v1beta1
+kind: ClusterRoleBinding
+metadata:
+  name: default-admin
```
ImJasonH, line 4: This seems like dangerous default behavior to recommend to users.

This grants the `default` SA (which runs every Pod in the cluster unless otherwise specified), _complete `cluster-admin` access_.

Can we either:

1. Make it clearer to users that this is only intended forr insecure/non-critical use cases, and should be modified after installation to match the user's use case and needs, and/or
2. Provide a more scoped-down role binding initially, which users can add to as they need?

vincent-pli: @ImJasonH 
Thanks for the comments, it's make sense, I will work for a pr for it.

PR #282, opened 2020-04-18
Add telegram example

GijsvanDulmen:
# Changes

A task for sending messages towards telegram as an alternative to slack.

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/telegrammessage/send-to-telegram.yaml
+++ b/telegrammessage/send-to-telegram.yaml
@@ -1,0 +1,27 @@
+apiVersion: tekton.dev/v1beta1
+kind: Task
+metadata:
+  name: send-to-telegram
+spec:
+  params:
+  - name: bot-token-secret
+    type: string
+    description: Token of the bot through 
+  - name: chat-id
+    type: string
+    description: ID of the chat to send the message towards
+  - name: message
+    type: string
+    description: The message to notify about
+  steps:
+  - name: post
+    image: curlimages/curl:7.68.0
+    script: |
+      #!/bin/sh
+      curl -X POST -H 'Content-Type: application/json' -d '{"chat_id": $(params.chat-id), "text": "$(params.message)" }' https://api.telegram.org/bot${BOT_TOKEN}/sendMessage
+    env:
+    - name: BOT_TOKEN
+      valueFrom:
+        secretKeyRef:
+          name: $(params.token-secret)
+          key: token
--- a/telegrammessage/send-to-telegram.yaml
+++ b/telegrammessage/send-to-telegram.yaml
@@ -18,10 +18,16 @@ spec:
     image: curlimages/curl:7.68.0
     script: |
       #!/bin/sh
-      curl -X POST -H 'Content-Type: application/json' -d '{"chat_id": $(params.chat-id), "text": "$(params.message)" }' https://api.telegram.org/bot${BOT_TOKEN}/sendMessage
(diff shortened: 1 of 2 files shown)


GijsvanDulmen:
> @googlebot I signed it!

@googlebot I signed it!

telegrammessage/send-to-telegram.yaml
```diff
@@ -0,0 +1,27 @@
...
+    description: ID of the chat to send the message towards
+  - name: message
+    type: string
+    description: The message to notify about
+  steps:
+  - name: post
+    image: curlimages/curl:7.68.0
+    script: |
+      #!/bin/sh
+      curl -X POST -H 'Content-Type: application/json' -d '{"chat_id": $(params.chat-id), "text": "$(params.message)" }' https://api.telegram.org/bot${BOT_TOKEN}/sendMessage
```
ImJasonH, line 21: I'm worried this might not escape correctly if the message has a `"` in it. Have you tested with this case?

GijsvanDulmen: Just tested it. Won't work with quotes in it. Could introduce this:
`export MESSAGE='$(params.message-to-send)'
        MESSAGE=`echo $MESSAGE | sed -e 's/"/\\\\"/g'``

But injection of the params.message-to-send could also contain single quotes and won't work as well. Is there any way to inject a variable "safely"? Otherwise the task run should just not use quotes I guess ;-)

chmouel: I had the same issue when writing the github/add_comment task and to do it I had to  avoid the shell and used in python like this : 

https://github.com/tektoncd/catalog/blob/e99f248a0e86475459ab10838c081a3800ce83ed/github/add_comment.yaml#L57-L59

chmouel: we should probably do the same for the slack tasks too, since it using shell and don't try to escape quotes...

GijsvanDulmen: Does it handle all situations? If so, that would indeed be better. I am wondering if there is a solution in the $(params.name) injection. Something like $(params.name|jsonquoted) or something with a better name...

chmouel: yes it will with the python conversion, escaped function would be something nice to have, but I am not sure if that's something planned

GijsvanDulmen: I've updated the command to make it work. The initial injection of variables can be done through ENV variables in a better way (quotes are handled). For future reference we could introduce a "curl + jq" image to keep it small like this:
`jq -n --arg chat_id "$CHAT_ID" --arg text "$TEXT" \
   '{chat_id: $chat_id, text: $text}' |
 /usr/bin/curl -H "Content-Type: application/json" -X POST -k -d@- \
   https://api.telegram.org/bot${BOT_TOKEN}/sendMessage`

chmouel:
/lgtm

thanks,

GijsvanDulmen:
> /lgtm
> 
> thanks,

Nice! Who is going to approve this? And who is going to add the OWNERS file etc?

telegrammessage/send-to-telegram.yaml
GijsvanDulmen: See: https://github.com/tektoncd/catalog/pull/300

PR #286, opened 2020-04-21
Fix image digest result in kaniko task

### LINKED ISSUE #285: kaniko image digest result does not work
# Expected Behavior

The kaniko task defines a task result called IMAGE_DIGEST that holds the digest of the new image.  A pipeline author can use the result as a parameter value in a subsequent task using the syntax $(tasks.xxx.results.IMAGE_DIGEST). 

# Actual Behavior

The parameter value is not substituted with the digest.  The result reference $(tasks.xxx.results.IMAGE_DIGEST) is passed as-is to the task instead.

1. Create a pipeline that uses the kaniko task.
2. Add another pipeline task that has a parameter with value $(tasks.xxx.results.IMAGE_DIGEST) where xxx is the pipeline task name associated with the kaniko task.

# Additional Info

There are a couple of problems.

First, a result name containing an underscore does not work.  I cannot find any documentation regarding the allowed characters in a result name.  However the code in `pkg/apis/pipeline/v1beta1/resultref.go` defines the following regex for result references which does not allow an underscore:
```
	variableSubstitutionFormat = `\$\([A-Za-z0-9-]+(\.[A-Za-z0-9-]+)*\)`
```
Interestingly the pipeline will store a result under a name containing an underscore but it won't let you successfully reference it.

Second, the last step of the kaniko task is a script:

```
    cat image-digested | jq '.[0].digest' | tee /tekton/results/IMAGE_DIGEST
```

`image-digested` is a file containing a json structure.  This gets piped into jq which is expecting to find the digest under a key named `digest`.  This is not correct.  Here is what the image-digested file looks like:

```
[{"key":"digest","value":"sha256:75608fb75f36d41139a2d377f8c0a234cee654d168c0c68b55bb1f6a751e2123","resourceRef":{"name":"us.icr.io/gregd/myimage"}}]
 ```

The digest is actually under a key named `value`.

Another problem with the jq command as written is that it will put the value in quotes and add a newline to the end.  Anyone trying to consume the digest will not be happy with this.  It is necessary to use the -j flag to get the value without quotes and without a newline.

Here is the corrected version of the script with the jq changes and the result name changed to use a dash instead of an underscore:

```
cat image-digested | jq -j '.[0].value' | tee /tekton/results/IMAGE-DIGEST
```

The kaniko task would need to declare the result name as IMAGE-DIGEST.  Alternatively if underscore is supposed to be accepted then an issue needs to be raised on the pipelines project.

GregDritschler:
Fixes #285 

# Changes

* Fixes kaniko.yaml as described in the issue.
* Updates the e2e test to test that the digest result is working.

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/kaniko/README.md
+++ b/kaniko/README.md
@@ -34,7 +34,7 @@ kubectl apply -f https://raw.githubusercontent.com/tektoncd/catalog/master/kanik
 
 ## Results
 
-* **digest**: The digest of the image just built.
+* **IMAGE-DIGEST**: The digest of the image just built.
 
 ## ServiceAccount
 
--- a/kaniko/kaniko.yaml
+++ b/kaniko/kaniko.yaml
@@ -20,7 +20,7 @@ spec:
   workspaces:
   - name: source
   results:
-  - name: IMAGE_DIGEST
+  - name: IMAGE-DIGEST
     description: Digest of the image just built.
 
   steps:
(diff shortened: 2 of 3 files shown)


GregDritschler:
@vdemeester Hi Vincent.  It looks like the tests passed.  Can you review?

PR #287, opened 2020-04-22
Add aws-cli task

Divyansh42:
# Changes

This task performs operations on Amazon Web Services resources using `aws`.

Authors: @vinamra28 @divyansh42

Signed-off-by: Divyansh42 <diagrawa@redhat.com>

# Submitter Checklist

review them:

### DIFF
--- a/aws-cli/README.md
+++ b/aws-cli/README.md
@@ -1,0 +1,87 @@
+# aws 
+
+This task performs operations on Amazon Web Services resources using `aws`.
+
+All aws cli commands can be found [here](https://docs.aws.amazon.com/cli/latest/reference/).
+
+## Install the Task
+
+```
+kubectl apply -f https://raw.githubusercontent.com/tektoncd/catalog/v1beta1/aws-cli/aws-cli.yaml
+```
+
+## Parameters
+
+- **SCRIPT**: The script of aws commands to execute e.g. `aws $1 $2` This will take
+ the first value and second value of ARGS as `s3` and `ls` (default: `aws $@`)
+- **ARGS**: The arguments to pass to `aws` CLI, which are appended 
+    to `aws` e.g. `s3 ls` ( default: `["help"]` ).
+
+
+## Workspaces
+
+- **source**: To mount file which is to be uploaded to the aws resources, 
+    this should be mounted using any volume supported in workspace.
+- **secrets**: A workspace that consists of credentials required by the `aws` which needs to be mounted to their default path as required by the aws.
+
+
+## Secret
+
+AWS `credentials` and `config` both should be provided in the form of `secret`.
+
+[This](../aws-cli/example/secret.yaml) example can be referred to create `aws-credentials`.
+
+Refer [this](https://docs.aws.amazon.com/sdk-for-java/v1/developer-guide/setup-credentials.html) guide for setting up AWS Credentials and Region.
+
+
+## Usage
+
+After creating the task, you should now be able to execute `aws` commands by 
+specifying the command you would like to run as the `ARGS` or `SCRIPT` param. 
+
+The `ARGS` param takes an array of aws subcommands that will be executed as 
+part of this task and the `SCRIPT` param takes the multiple commands that you would like to run on aws CLI.
+
+This [example](../aws-cli/example/secret.yaml), can be referred to create secret file for aws credentials.
+
+In the [example](../aws-cli/example/run.yaml), ConfigMap as the volume is used. In place of ConfigMap, any volume supported in workspace can be used.
+
+Following `command` can be used to create `ConfigMap` from the `file`.
+```
+kubectl create configmap upload-file --from-file=demo.zip
+```
+Here `upload-file` is the name of the `ConfigMap`, this can be changed based on the requirements.
+
+See [here](../aws-cli/example/run.yaml) for example of `aws s3` command.
+
+
+### Note
+
+
+- Either `SCRIPT` or `ARGS` must be provided in the `params` of the task.
+
+- To support multiple `aws` commands to run on a single task, SCRIPT can be used as follows:
+
+  ```
+  - name: SCRIPT
+    value: |
+      aws s3 mb s3://test-bucket
+      aws s3api put-object --bucket test-bucket --key test/
+      aws s3 cp $(workspaces.source.path)/demo.zip s3://test-bucket/test/demo.zip
+  ```
+
+- In case there is no requirement of file that is to be uploaded on the aws resources,
+ then `emptyDir` needs to be mounted in the `workspace` as shown below:
+    ```
+    workspaces:
+      - name: source
+        emptyDir: {}
+    ```
+    otherwise, if `Volume(e.g. ConfigMap) is needed:
+
+    ```
+    workspaces:
+      - name: source
+        configmap:
+            name: upload-file 
+    ```
--- a/aws-cli/aws-cli.yaml
+++ b/aws-cli/aws-cli.yaml
@@ -1,0 +1,24 @@
+apiVersion: tekton.dev/v1beta1
+kind: Task
+metadata:
+  name: aws-cli
+spec:
+  workspaces:
+    - name: source
+    - name: secrets
+      mountPath: /tekton/home/.aws
+  params:
+    - name: SCRIPT
+      description: The AWS script to run
+      type: string
+      default: "aws $@"
+    - name: ARGS
+      description: AWS cli arguments to be passed
+      type: array
+      default: ["help"]
+  steps:
+    - name: awscli
+      image: amazon/aws-cli:latest
+      script: "$(params.SCRIPT)"
+      args:
+        - "$(params.ARGS)"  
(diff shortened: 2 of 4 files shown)


Divyansh42:
@chmouel I haven't added test directory here, due to the aws credentials problem.  Is that the reason for the failure of the integration tests?

Divyansh42:
@sbwsg @chmouel can you please review this PR?

aws-cli/README.md
```diff
@@ -0,0 +1,68 @@
...
+- Either `SCRIPT` or `ARGS` must be provided in the `params` of the task.
+
+- In case there is no requirement of file that is to be uploaded on the aws resources,
+ then `emptyDir` needs to be mounted in the `workspace` as shown below:
+    ```
+    workspaces:
+      - name: images-url
+        emptyDir: {}
+    ```
+    otherwise, if ConfigMap is needed:
```
savitaashture, line 61: Can it be any volume which workspace supports or only `ConfigMap` ?

Divyansh42: Yes, It can be any volume. Will update this.

aws-cli/README.md
```diff
@@ -0,0 +1,68 @@
...
+
+## Install the Task
+
+```
+kubectl apply -f https://raw.githubusercontent.com/tektoncd/catalog/v1beta1/aws-cli/aws-cli.yaml
+```
+
+## Parameters
+
+- **SCRIPT**: The script to run on `aws` CLI e.g `aws s3 ls` ( default: `aws $@` ).
```
vinamra28, line 13: Can't this be done as `aws $1 $2` and pass the values in ARGS

Divyansh42:
@vdemeester can you please review this PR? :slightly_smiling_face:

aws-cli/README.md
```diff
@@ -0,0 +1,72 @@
...
+- **source**: To mount file which is to be uploaded to the aws resources, 
+    this should be mounted using any volume supported in workspace.
+
+
+## Secret
+
+AWS `credentials` and `config` both should be provided in the form of `secret`,
+ which is mounted as `volume` in the task.
+
+[This](./example/secret.yaml) example can be referred to create `aws-secrets`.
```
bobcatfish, line 30: hm this link doesnt work in the preview! this is a super minor thing, but i think if you drop the `.` it'll probably work :D

bobcatfish: besides this example, do you have any other docs you can link users to with guidance around creating these secrets? (e.g. aws docs)

I find that setting up secrets correctly can be one of the most challenging parts of getting Tasks to work :D

aws-cli/README.md
```diff
@@ -0,0 +1,72 @@
...
+
+
+## Usage
+
+After creating the task, you should now be able to execute `aws` commands by 
+specifying the command you would like to run as the `ARGS` or `SCRIPT` param.
```
bobcatfish, line 36: is it possible to have a link here as well to docs on aws commands that users could potentially run?

aws-cli/aws-cli.yaml
```diff
@@ -0,0 +1,30 @@
...
+      default: "aws $@"
+    - name: ARGS
+      description: AWS cli arguments to be passed
+      type: array
+      default: ["help"]
+  steps:
+    - name: awscli
+      image: amazon/aws-cli:latest
+      workingDir: "/tekton/home/.aws"
+      script: "$(params.SCRIPT)"
```
bobcatfish, line 21: can you explain more about why someone would want to override SCRIPT? with the user being able to override both SCRIPT and ARGS it seems like they could use this task to generically execute _anything_ inside the aws-cli image - maybe that's okay, but its a different choice than we've made with other Tasks where we try to make the scope more constrained

my recommendation here would be to at least always be executing the `aws` command - but if you feel strongly we can leave it like this!

aws-cli/aws-cli.yaml
```diff
@@ -0,0 +1,30 @@
...
+      args:
+        - "$(params.ARGS)"
+      volumeMounts:
+        - name: aws-secrets
+          mountPath: "/tekton/home/.aws"
+  volumes:
+    - name: aws-secrets
+      secret:
+        secretName: aws-secrets
```
bobcatfish, line 30: is it possible to use workspaces here instead of volumes? https://github.com/tektoncd/pipeline/blob/master/docs/workspaces.md we're trying to avoid using volumes b/c they give folks less control at runtime

vinamra28: @bobcatfish we thought of using the volumes so as that users can mount their aws credentials at the right place. There is also one option of mounting the secrets for AWS is via the environment. Could you please suggest which one would be better.
Ref:- https://docs.aws.amazon.com/sdk-for-java/v1/developer-guide/setup-credentials.html

aws-cli/README.md
vinamra28: for the secrets https://github.com/tektoncd/catalog/pull/287#discussion_r414690879 
and sure will correct the link in the README.

aws-cli/README.md
vinamra28: Sure. Will do that :wink:

aws-cli/aws-cli.yaml
vinamra28: Mounting secrets via environment just like https://github.com/tektoncd/pipeline/blob/master/docs/tasks.md#using-a-secret-as-an-environment-source

aws-cli/aws-cli.yaml
Divyansh42: Yes, I tried to generalize the task here, so that we can run any `aws` commands.
Also, I took reference from existing `openshift-client` task. https://github.com/tektoncd/catalog/blob/v1beta1/openshift-client/openshift-client-task.yaml

cc @nikhil-thomas

Divyansh42:
> I left a few thoughts and just a couple minor changes I'd like to see, I think the workspaces change is the one I feel the most strongly about.
> 
> @Divyansh42 before we merge, can you confirm that you can successfully run the Task? Ideally we'll have some test infrastructure setup so we can actually run this against AWS (or a test double of some kind) but since we're not there yet I don't want to block you (interested in your thoughts on this @vdemeester @chmouel @afrittoli )

@bobcatfish  Thanks for the review, this will be very helpful in improving this task. :slightly_smiling_face: 
Will take all your comments into the consideration, and update with the better approach :sweat_smile:

Yes, I was able to run this task successfully. I tried with the `aws s3` commands. But a testing infrastructure can really help to test the task.

aws-cli/README.md
vdemeester: > hm this link doesnt work in the preview! this is a super minor thing, but i think if you drop the `.` it'll probably work :D

why wouldn't it ? it's relative to the folder of the file :thinking:

aws-cli/aws-cli.yaml
vdemeester: Right, it's similar to `openshift-client`, a very "generic" Task. Allowing writing their own scripts make the task more powerful.

Default it will do `aws help`, if you give args ("foo bar") it will do `aws foo bar`. But if you need something more complex you could do

```yaml
params:
- name: SCRIPT
  value: |
#!/usr/bin/env bash
while getopts u:d:p:f: option
do
case "${option}"
in
p) PREINIT=${OPTARG};;
n) NEXT=${OPTARG};;
esac
done
if [[ -z "$PREINIT" ]]; do 
    # run preinit
    $PRINIT
fi
echo "Do something"
aws subcommand $NEXT
echo "Do the rest"
aws anothersubcommand "$@"
- name: ARGS
  value: ["-p", "$(params.preinit)", "s3", "upload"]
```

aws-cli/aws-cli.yaml
vdemeester: We should try to use `workspace` as much as possible in place of `volumes`. You know you want the secrets in a volume, you don't need to force the user to name it specifically, he will be able to bind it at runtime.

vdemeester: (as the docs describe in @bobcatfish link)

aws-cli/README.md
Divyansh42: Yes, it is working. I cross checked that. 
@bobcatfish can you please check once again? :sweat_smile:

aws-cli/aws-cli.yaml
Divyansh42: Here, I have added `SCRIPT` param to support multiple aws commands (as, in ARGS we can only pass single command). It will help us to avoid creating pipeline for running multiple commands. We can run all commands having only single taskrun.

Will add more explanation for this in README.

aws-cli/aws-cli.yaml
Divyansh42: Okay, will update the task to use workspace. :slightly_smiling_face:

vinamra28:
@vdemeester @bobcatfish could you please review the changes :sweat_smile:

Divyansh42:
@vdemeester @bobcatfish If there is no improvement needed then can you please merge this PR? :slightly_smiling_face:

PR #291, opened 2020-04-23
Fix kaniko digest results 🔧

bobcatfish:
# Changes

I tried to use the kaniko task to update the build + deploy pipelinerun
example in the pipelines repo and run into some trouble which I have
fixed here:
- The jq parsing of the digest from the imagedigestexporter output
  wasn't quite right, the name of the field was incorrect and it was
  surrounded by quotes (removed with -r) and followed by a newline
  (removed with -j)
- If the kaniko Task was used more than once with the same workspace,
  even if it was building from different directories, the digests would
  conflict b/c it was always writing to the same files. Now the files it
  writes to are relative to the context the image is alreayd using to
  build

I've also extended the test so that it both verifies that a digest is
parsed.

# Submitter Checklist

review them:

- [n/a] Includes [docs](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if user facing)
- [😅 ] Yaml file complies with [yamllint](https://github.com/adrienverge/yamllint) rules.

for more details._

### DIFF
--- a/kaniko/README.md
+++ b/kaniko/README.md
@@ -64,12 +64,4 @@ spec:
   - name: source
     persistentVolumeClaim:
       claimName: my-source
-  resources:
-    outputs:
-    - name: image
-      resourceSpec:
-        type: image
-        params:
-        - name: url
-          value: gcr.io/my-repo/my-image
 ```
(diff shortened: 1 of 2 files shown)


vdemeester:
Isn't this a little duplicate of https://github.com/tektoncd/catalog/pull/286 :stuck_out_tongue_closed_eyes:

bobcatfish:
Whoops! Some of it is for sure, some of it isn't - I'll remove the parts fixed in #286 :D

Issue #265, opened 2020-04-07
Buildah task would be faster with `--no-cache` than `--layers`


holly-cummins:
# Expected Behavior

The buildah task should have the optimum arguments to go fast in a CI/CD environment. 

# Actual Behavior

I can improve buildah build times from ~10m to ~6m by adjusting the cache settings. I don't think investing in caching onto an ephemeral volume is likely to be the right behaviour for most scenarios; even with my multi-stage build performance is faster without the laters.

# Steps to Reproduce the Problem

1. Run a build in a Tekton pipeline, with the builah task. I'm using OpenShift 4.
2. Measure the image build time.
3. Fork the buildah task, replace `--layers` with `--no-cache`, and update the pipeline to use the fork. 
4. Run a build and measure the image build time. 

For my cluster and dockerfile, the update seems to have improved image build times from ~10m to ~6m. 

I don't believe there's much value in the `--layers` option in the task as written, because it uses an `emptyDir()` volume, so any cached layers are ephemeral. Because there's no caching between builds, the cost of writing the layers outweighs the saving from the cache. (https://github.com/containers/buildah/issues/1455 has some discussion of caching in CI/CD environments.)

# Additional Info

I explored the possibility of replacing the `emptyDir{}` volume with a persistent volume claim, but (after a bit of a struggle with storage types and storage drivers) I discovered https://github.com/containers/buildah/issues/2215. I think that means that if if there were a persistent volume, tekton wouldn't use the cache much because the timestamps of the source files are always new. 

This is my Dockerfile, but I don't think there's anything distinctive in it; I think the perf characteristics would be similar no matter what the dockerfile.

```
# base image
FROM node:alpine AS builder
# set working directory
WORKDIR /app
# install client node packages
COPY ./client/package.json /client/yarn.lock ./
RUN yarn install --network-timeout 100000
# yarn build for client
RUN printenv
COPY ./client .
# yarn assumes build is within root directory
RUN yarn --cwd ./client run build
# build our server
FROM node:alpine
WORKDIR /app
# install server node packages
COPY ./server/package.json /server/yarn.lock ./
RUN yarn install --network-timeout 100000
# import build files from client
COPY --from=builder /app/build ./public
RUN printenv
COPY ./server .

#This isn't necessary in a privileged context, or when built with docker, but is needed with buildah
ENV HOME /app

CMD ["yarn", "start"]
```

holly-cummins:
[...]

PR #293, opened 2020-04-28
Do not use layers in buildah task, since layer caching depends on timestamps (unreliable in a CI/CD

holly-cummins:
# Changes

See discussion in https://github.com/tektoncd/catalog/issues/265. Although using
 layer caching in the buildah build would be nice, there are two prerequisites:
* A PVC to store the cache between pipeline runs
* A fix for https://github.com/containers/buildah/issues/2215

The buildah issue is that buildah uses file timestamps to invalidate the cache. Locally, this works well, but in a CI/CD the files will always be freshly downloaded from source control, so they will be newer than the cache. This means the cache would never be used, even if it was available in a PVC. 
 

# Submitter Checklist

review them:

- [ n/a] Includes [docs](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if user facing)
- [x ] Commit messages follow [commit message best practices](https://github.com/tektoncd/community/blob/master/standards.md#commit-messages)
- [x ] Yaml file complies with [yamllint](https://github.com/adrienverge/yamllint) rules.

for more details._

### DIFF
--- a/buildah/buildah.yaml
+++ b/buildah/buildah.yaml
@@ -29,7 +29,7 @@ spec:
   - name: build
     image: $(params.BUILDER_IMAGE)
     workingDir: $(workspaces.source.path)
-    command: ['buildah', 'bud', '--format=$(params.FORMAT)', '--tls-verify=$(params.TLSVERIFY)', '--layers', '-f', '$(params.DOCKERFILE)', '-t', '$(params.IMAGE)', '$(params.CONTEXT)']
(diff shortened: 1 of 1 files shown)


chmouel:
Thanks for submitting this @holly-cummins would you like to sign the CLA process so we can merge this patch ?

let us know if you have difficulties with the CLA process.

holly-cummins:
Hi @chmouel, I was indeed having problems with the CLA. I just re-clicked through the steps and whatever it was is resolved, so I'm good. Thanks for following up!

PR #294, opened 2020-04-29
Enhance "kubectl wrapper

vincent-pli:
The requirements discussed in #233
- action: ['get', 'create', 'apply', 'delete', 'replace', 'patch'] -- the action to perform to the resource
- merge_strategy: ['strategic', 'merge', 'json'] -- the strategy used to merge a patch, defaults to "strategic"
- success_condition: a label selector expression which describes the success condition
- failure_condition: a label selector expression which describes the failure condition
- manifest: the kubernetes manifest"

# Changes

# Submitter Checklist

review them:

for more details._


kubectl/taskrun.yaml
```diff
@@ -6,16 +6,30 @@ spec:
   taskRef:
     name: kubectl-deploy-pod
   params:
+  - name: action
+    value: create
```
savitaashture, line 10: When action is `GET` where can we mention the object name 

Ex: 
For creating pod we can mention action as `create` and yaml content in `manifest` section
similarly for `GET`, `DELETE` where do we mention which resource needs to be deleted

vincent-pli: For now, all `action` need a `manifest`, so you need supply `manifest` for `Get` and `Delete`, maybe with a simplified version, maybe with `ObjectReference` only.

savitaashture: Can we have a one more task run specifying GET or DELETE operation with manifest
It would be easy to understand how to use task when we wanted to do GET or DELETE operation

vincent-pli: Agree with you, thanks

kubectl/kubectl-deploy.yaml
```diff
@@ -5,14 +5,60 @@ metadata:
...
+    description: A label selector express to decide if the action on resource is failure.
+  - name: output
+    default: ""
+    description: An express to retrieval data from resource.
+  - name: set-ownerreference
+    default: "false"
+    description: Enable set owner reference for created resource.
   - name: image
-    default: gcr.io/cloud-builders/kubectl # it is huge
+    default: index.docker.io/aipipeline/kubeclient:v0.0.2 # it is huge
```
vinamra28, line 28: What is the difference between this image and `gcr.io/cloud-builders/kubectl`? Just had a query :sweat_smile:

vincent-pli: @vinamra28 The `gcr.io/cloud-builders/kubectl` only supply cmd `kubectl`, so we build another one based on it to support more complex features: `success-condition/failure-conditio`, `set-ownerreference` .etc.

vinamra28: Okay. Thanks

kubectl/README.md
```diff
@@ -34,17 +49,31 @@ spec:
...
-        name: myapp-pod
+        generateName: myapp-pod-
         labels:
           app: myapp
       spec:
         containers:
         - name: myapp-container
           image: docker
-          command: ['sh', '-c', 'echo Hello Kubernetes! && sleep 3600']
+          command: ['sh', '-c', 'echo Hello Kubernetes! && sleep 30']
```
vinamra28, line 78: Can we mount the `MANIFESTs` via the `ConfigMap` and `Workspaces` instead of specifying them over here in the `run`?  This will make the code cleaner and also we can mount many manifests at the same time eg mounting `deployment.yaml` and `service.yaml`. WDYT?
Will it satisfy your usecase also?
cc @savitaashture

vincent-pli: @vinamra28 
Thanks for your comments, 
I'm afraid for now can not support multiple manifest in one step.
The manifest will be created success, but the monitor/trace will failed, is that a common requirement?

vinamra28: Okay...No issues.
Actually I was thinking to add a task which can support `kubectl` commands which may or may not require manifests eg `kubectl rollout status deployment.v1.apps/nginx-deployment`  or if in case we provide manifests via the workspaces then we are able to `apply` that also.

vincent-pli: Seems the version before this one could satisfy you 😸

vinamra28: The previous version could only satisfy one of the scenario of applying multiple manifests by adding workspaces :sweat_smile: , one more support I want to add is to perform `kubectl` commands even if there is no manifest provided. (eg `kubectl get pods`). A kind of very generic task which can perform k8s actions :smile: . I'll try raising a PR for it. :smiley:

chmouel:
/lgtm

thanks, we should probably use this as a "template" for other meta tasks (ie: the git discussion)

chmouel:
It's unfortunate tho, we don't have any tests for this since I have the feeling this could be quite a critical task (like the git-clone task)

PR #300, opened 2020-05-05
Update to curl jq variant with better quote handling and reliable image

GijsvanDulmen:
# Changes

Based on a discussion in PR https://github.com/tektoncd/catalog/pull/282 we needed a better way of doing simple web requests from catalog tasks. Introducing a simple image with curl and jq (see plumbing PR https://github.com/tektoncd/plumbing/pull/372) we can handle quotes better and make it more stable and reliable therefore.

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/slackmessage/send-to-channel-slack.yaml
+++ b/slackmessage/send-to-channel-slack.yaml
@@ -15,13 +15,21 @@ spec:
     description: plain text message
   steps:
   - name: post
-    image: curlimages/curl:7.68.0
+    image: gcr.io/tekton-releases/dogfooding/catalog-web-requests:latest
     script: |
       #!/bin/sh
-      /usr/bin/curl -X POST -H 'Content-type: application/json' -H 'Authorization: Bearer '$TOKEN --data '{"channel":"$(params.channel)", "text":"$(params.message)"}' https://slack.com/api/chat.postMessage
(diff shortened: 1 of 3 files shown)


slackmessage/send-to-channel-slack.yaml
```diff
@@ -15,13 +15,21 @@ spec:
     description: plain text message
   steps:
   - name: post
-    image: curlimages/curl:7.68.0
+    image: gcr.io/tekton-releases/dogfooding/catalog-web-requests:latest
```
ImJasonH, line 18: Please add a comment describing what this image is for, what it provides, etc.

I'm not actually sure we want to depend on it, since Tekton release infra might change to no longer require it, at which point it might get removed or stop getting updates. @afrittoli any thoughts?

afrittoli: I agree @ImJasonH, infra images are built and maintained for infra, we probably should not use them in a catalog for our users.

GijsvanDulmen: It's an image from the other PR on the plumbing repo specifically meant for catalog tasks. I would prefer to put the Dockerfile over here and make the catalog ci/cd build the image though. Is that possible?

vdemeester:
@GijsvanDulmen  switched the base branch to master as it got merged with `v1beta1` now :angel:

PR #303, opened 2020-05-06
Add Github Release task

Divyansh42:
# Changes
This task can be used to make a release in the Github.

Multiple `assets` including `binaries` of the released version and the `release notes`, can also be uploaded with the release.

Signed-off-by: Divyansh42 <diagrawa@redhat.com>

# Submitter Checklist

review them:

### DIFF
--- a/github/create-github-release/create-github-release.yaml
+++ b/github/create-github-release/create-github-release.yaml
@@ -1,0 +1,62 @@
+apiVersion: tekton.dev/v1beta1
+kind: Task
+metadata:
+  name: create-github-release
+spec:
+  workspaces:
+    - name: input
+    - name: source
+    - name: release-notes
+  params:
+    - name: TAG
+      description: A git tag that will be created with this release (e.g. v1.0.0)
+      type: string
+    - name: REVISION
+      type: string
+      description: Git revision to create a release from (branch, tag, sha, ref…).
+      default: master
+    - name: RELEASE_FILE_NAME
+      type: string
+      description: Name of the file that has to be uploaded as release notes.
+      default: release.md
+    - name: GITHUB_TOKEN_SECRET
+      type: string
+      description: Name of the secret holding the github-token.
+      default: github-token
+    - name: GITHUB_TOKEN_SECRET_KEY
+      type: string
+      description: Name of the secret key holding the github-token.
+      default: GITHUB_TOKEN
+  steps:
+    - name: create-release
+      workingdir: $(workspaces.source.path)
+      image: quay.io/diagrawa/github-hub:latest
+      script: |
+        #!/usr/bin/env bash
+        set -ex
+
+        # Appending command to upload multiple release assets.
+
+        UPLOAD_ASSET=$(workspaces.input.path)/*
+        cmd=""
+        for file in $UPLOAD_ASSET
+        do
+          cmd="$cmd\
+            --attach $file"
+        done
+
+        # Create a release
+        echo "Creating release $(params.TAG)"
+
+        hub release create \
+          --commitish $(params.REVISION) \
+          --file $(workspaces.release-notes.path)/$(params.RELEASE_FILE_NAME) \
+          $cmd \
+          $(params.TAG)
+
+      env:
+        - name: GITHUB_TOKEN
+          valueFrom:
+            secretKeyRef:
+              name: $(params.GITHUB_TOKEN_SECRET)
+              key: $(params.GITHUB_TOKEN_SECRET_KEY)
--- a/github/create-github-release/example/run.yaml
+++ b/github/create-github-release/example/run.yaml
@@ -1,0 +1,69 @@
+apiVersion: tekton.dev/v1beta1
+kind: Pipeline
+metadata:
+  name: create-github-release-pipeline
+spec:
+  workspaces:
+    - name: shared-workspace
+    - name: upload
+    - name: release-notes
+  tasks:
+    - name: fetch-repository
+      taskRef:
+        name: git-clone
+      workspaces:
+        - name: output
+          workspace: shared-workspace
+      params:
+        - name: url
+          value: https://github.com/divyansh42/aws-cli.git
+        - name: subdirectory
+          value: ""
+        - name: deleteExisting
+          value: "true"
+    - name: create-github-release
+      taskRef:
+        name: create-github-release
+      params:
+        - name: TAG
+          value: "v1.0"
+        - name: REVISION
+          value: master
+        - name: RELEASE_FILE_NAME
+          value: release.md
+        - name: GITHUB_TOKEN_SECRET
+          value: github-token
+        - name: GITHUB_TOKEN_SECRET_KEY
+          value: GITHUB_TOKEN
+      runAfter:
+        - fetch-repository
+      workspaces:
+        - name: source
+          workspace: shared-workspace
+        - name: input
+          workspace: upload
+        - name: release-notes
+          workspace: release-notes
+---
+apiVersion: tekton.dev/v1beta1
+kind: PipelineRun
+metadata:
+  name: create-github-release-pipeline-run
+spec:
+  pipelineRef:
+    name: create-github-release-pipeline
+  workspaces:
+    - name: shared-workspace
+      volumeClaimTemplate:
+        spec:
+          accessModes:
+          - ReadWriteOnce
+          resources:
+            requests:
+              storage: 500Mi 
+    - name: upload
+      configmap:
+        name: upload-asset
+    - name: release-notes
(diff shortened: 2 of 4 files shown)


sbwsg:
I wonder if this might be better split into two separate tasks, one for github and one for gitlab?  Lots of the params are required or used by only one platform or the other.  We already have a `github` directory here: https://github.com/tektoncd/catalog/tree/v1beta1/github so I suppose a `gitlab` directory would also make sense.  Kind of related as well, `git-actions` doesn't communicate a ton (at least to me) about what else might live in here and why it's different from the `git` directory of tasks.  WDYT?

Divyansh42:
> I wonder if this might be better split into two separate tasks, one for github and one for gitlab? Lots of the params are required or used by only one platform or the other. We already have a `github` directory here: https://github.com/tektoncd/catalog/tree/v1beta1/github so I suppose a `gitlab` directory would also make sense. Kind of related as well, `git-actions` doesn't communicate a ton (at least to me) about what else might live in here and why it's different from the `git` directory of tasks. WDYT?

@sbwsg Yes, it will be better if we split the task for github and gitlab, as there are only few common params.
I will add the github release task to the existing github directory by making a seperate directory for the github release inside the github directory which will have github-release task and all related README and example for the same.
And for gitlab release will create a seperate directory in the catalog for the gitlab. 
Is this approach sounds good to you?

sbwsg:
> I will add the github release task to the existing github directory by making a seperate directory for the github release inside the github directory which will have github-release task and all related README and example for the same.
> And for gitlab release will create a seperate directory in the catalog for the gitlab.
> Is this approach sounds good to you?

This sounds great to me!

Divyansh42:
> > I will add the github release task to the existing github directory by making a seperate directory for the github release inside the github directory which will have github-release task and all related README and example for the same.
> > And for gitlab release will create a seperate directory in the catalog for the gitlab.
> > Is this approach sounds good to you?
> 
> This sounds great to me!

Okay, will raise a seperate PR for the gitlab and modify this one for github.

Divyansh42:
@sbwsg Please review the PR. This is for the Github release.

github/github-release/README.md
```diff
@@ -0,0 +1,97 @@
...
+kubectl apply -f https://raw.githubusercontent.com/tektoncd/catalog/v1beta1/github/github-release/github-release.yaml
+```
+
+## Parameters
+
+- **TAG_NAME**: The tag where the release will be created from (_e.g:_`v1.0.0`).
+- **NAME**: The Name of the release (_e.g:_`First release`).
+- **BODY**: A short description of the release (_default:_`""`).
+- **OWNER**: The `user name` of the Github owner.
+- **REPO_NAME**: Full name of the `repository` where the release is to made.
```
sbwsg, line 25: Does this include the domain name like `github.com/tektoncd/pipeline`?  Or is it an org + repo like `tektoncd/pipeline`? Suggest including a small example in the description.

github/github-release/README.md
```diff
@@ -0,0 +1,97 @@
...
+
+## Install the Task
+
+```
+kubectl apply -f https://raw.githubusercontent.com/tektoncd/catalog/v1beta1/github/github-release/github-release.yaml
+```
+
+## Parameters
+
+- **TAG_NAME**: The tag where the release will be created from (_e.g:_`v1.0.0`).
```
sbwsg, line 21: The tag will be created when this runs, right?  Suggest updating the description to:

> A git tag name that will be created with this release (_e.g._ `v1.0.0`)

github/github-release/README.md
```diff
@@ -0,0 +1,97 @@
...
+- **NAME**: The Name of the release (_e.g:_`First release`).
+- **BODY**: A short description of the release (_default:_`""`).
+- **OWNER**: The `user name` of the Github owner.
+- **REPO_NAME**: Full name of the `repository` where the release is to made.
+- **BRANCH**: Branch from the repository on which release has to be made (_default:_`master`).
+- **UPLOAD_ASSET_FILE_NAME**: The file name of the asset that needs to be uploaded.
+- **github-token-secret**: The name of the `secret` holding the github-token (_default:_`github-token`).
```
sbwsg, line 28: I notice that the param names here are a mix of ALL_UPPERCASE and all_lowercase.  Suggest standardizing on a single format.

github/github-release/github-release.yaml
```diff
@@ -0,0 +1,106 @@
+apiVersion: tekton.dev/v1beta1
+kind: Task
+metadata:
+  name: github-release
```
sbwsg, line 4: suggest create-github-release as a name here.  It might be that at some point we also want a remove-github-release or draft-github-release.

github/github-release/github-release.yaml
```diff
@@ -0,0 +1,106 @@
...
+      default: github-secret
+    - name: github-token-secret-key
+      type: string
+      description: name of the secret key holding the github-token.
+      default: api-token
+  steps:
+    - name: release
+      image: registry.access.redhat.com/ubi8/ubi-minimal:latest
+      script: |
+        #!/usr/bin/env bash
```
sbwsg, line 45: hm.  there's a tool called [`hub`](https://hub.github.com/#scripting) that might be useful here?  It would replace the need to have bash marshaling JSON & making a series of curl requests.

sbwsg: We do something similar to this in Tekton's plumbing for releases.  See the Task we use here for an example of utilizing `hub`: https://github.com/tektoncd/plumbing/blob/master/tekton/resources/release/github_release.yaml#L212-L223

github/github-release/README.md
Divyansh42: Here, `OWNER` will be `tektoncd` and `REPO_NAME` will be `pipeline`. 
Will add a example in the description.

github/github-release/README.md
Divyansh42: Okay, thanks for the suggestion :slightly_smiling_face:

github/github-release/github-release.yaml
Divyansh42: Okay. will change this.

github/github-release/github-release.yaml
Divyansh42: Okay, will go through this tool. I wasn't aware of this :sweat_smile: 
I guess, this will definitely improve the task. Thanks :+1:

github/create-github-release/create-github-release.yaml
```diff
@@ -0,0 +1,62 @@
...
+      description: Name of the secret holding the github-token.
+      default: github-token
+    - name: GITHUB_TOKEN_SECRET_KEY
+      type: string
+      description: Name of the secret key holding the github-token.
+      default: GITHUB_TOKEN
+  steps:
+    - name: create-release
+      workingdir: $(workspaces.source.path)
+      image: gcr.io/tekton-releases/dogfooding/hub
```
sbwsg, line 33: OK, apologies one further wrinkle: we probably shouldn't rely on one of Tekton's infra images since we don't want to be responsible for modifications there breaking a catalog task.  Looks like there are several publicly available images for `hub`:

- https://hub.docker.com/r/kanisterio/alpine-git-hub
- https://hub.docker.com/r/tianon/github-hub/

Or it might be a good time to build a fresh image and publish it to docker hub / quay to support the specific needs of this Task?

Divyansh42: Yeah sure, I have build the image and updated the task. Also added link to the `Dockerfile` in the `README.md`

github/create-github-release/example/run.yaml
```diff
@@ -0,0 +1,64 @@
...
+apiVersion: tekton.dev/v1beta1
+kind: PipelineRun
+metadata:
+  name: create-github-release-pipeline-run
+spec:
+  pipelineRef:
+    name: create-github-release-pipeline
+  workspaces:
+    - name: shared-workspace
+      persistentVolumeClaim:
```
sbwsg, line 57: you could use a `volumeClaimTemplate` here, which would mean you could drop the separate PVC YAML file if you wanted.

This workspace would become:

```
- name: shared-workspace
  volumeClaimTemplate:
    spec:
      accessModes:
      - ReadWriteOnce
      resources:
        requests:
          storage: 500Mi 
```

The nice benefit is that someone using this example gets the PVCs cleaned up for free when they delete the PipelineRun.

Divyansh42: Okay, thanks for suggestion. This is really helpful :slightly_smiling_face:

Divyansh42:
@sbwsg can you please add `ok-to-test` label :sweat_smile:

PR #309, opened 2020-05-10
[WIP] Add generic git task - with alternative auth

jlpettersson:
# Changes
The existing [git-clone task](https://github.com/tektoncd/catalog/tree/v1beta1/git) is very opinionated and custom. This task use a generic _git image_. The plan is to add tasks for `git-add`, `git-commit` and `git-push` as well.

This commit also shows an alternative way to setup authentication. This authentication is independent of [Tekton Pipeline Auth documentation](https://github.com/tektoncd/pipeline/blob/master/docs/auth.md) and does not use _annotations on Secrets_ and no need for custom _credential initialization process_ and does not copy or move any sensitive values, but mount them using a projected volume.

Related to https://github.com/tektoncd/pipeline/issues/2343

/kind feature

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/git-ssh/git-clone-ssh.yaml
+++ b/git-ssh/git-clone-ssh.yaml
@@ -1,0 +1,28 @@
+apiVersion: tekton.dev/v1beta1
+kind: Task
+metadata:
+  name: git-clone-ssh
+spec:
+  workspaces:
+    - name: output
+      description: The git repo will be cloned onto the volume backing this workspace
+  params:
+    - name: url
+      type: string
+      description: git url to clone
+  steps:
+    - name: git-clone
+      image: bitnami/git:2.26.2
+      command: ['git', '-c', 'core.sshCommand=ssh -i /etc/ssh/id_rsa', 'clone', '$(params.url)', '$(workspaces.output.path)']
+      volumeMounts:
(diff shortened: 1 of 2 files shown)


git-ssh/git-clone-ssh.yaml
```diff
@@ -0,0 +1,28 @@
...
+          name: github-ssh-auth
+  volumes:
+    - name: github-ssh-auth
+      projected:
+        defaultMode: 0400
+        sources:
+          - secret:
+              name: github-known-hosts
+          - secret:
+              name: github-private-key
```
jlpettersson, line 28: both secret names should be _params_ in a proper implementation.

sbwsg: Is this a limitation of Tekton at the moment - params can't be used as variables in volumes?

jlpettersson: Yes, it is the Secret Name within a projected volume, were we don't have implemented _variable substitution_. But as the commit is implemented now, it is working. But for proper support, we should add _variable substitution_ here, so that _end-users_ can name the secrets how they want.

PR #329, opened 2020-05-19
Add proxy settings to kaniko task

micxer:
# Changes

Add proxy settings to kaniko task

# Submitter Checklist

review them:


chmouel:
Thanks for your contribution, i think we may want to think about this, if we are start adding those env variables we will need to add this to every tasks that does http requests.... I wonder if there is a better way to do this ?

micxer:
Agreed. Now that you mentioned it, having a general way to do this might be better. I will have a look into [PodTemplates](https://github.com/tektoncd/pipeline/blob/master/docs/podtemplates.md) to see if that helps.

chmouel:
@micxer i think it would be great if it doesn't to try to solve this directly in pipeline (as github.com/tektoncd/pipeline)

vdemeester:
@micxer  switched the base branch to master as it got merged with `v1beta1` now :angel:

PR #332, opened 2020-05-23
Add git-clone-ssh

jlpettersson:
# Changes

This is a git-clone Task using SSH authentication that is easy to configure. With this Task Tekton only orchestrate tasks, but does not handle Secrets. The user declare the Secrets he has configured and Tekton does not manage them. This is similar to _immutable infrastructure_ practices and serves as an alternative.

Here, _runtime considerations_, e.g. volumes and secret names is handled in a way as described in https://github.com/tektoncd/pipeline/issues/2680
This depends on the bugfix in https://github.com/tektoncd/pipeline/pull/2683

Closes https://github.com/tektoncd/catalog/pull/309

/kind feature

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/git/README.md
+++ b/git/README.md
@@ -177,6 +177,17 @@ There are 4 additional parameters in addition to the ones mentioned above for th
 
 ### Usage
 
+## `git-clone-ssh`
+
+This task does `git clone` using SSH with the authentication described in [Git SSH Auth](./git-ssh-auth.md).
+
+This `Task` has two input parameters:
+
+1. `url` (**required**) is the url to the git repository
+2. `path` (optional) is the path on the workspace volume, `code` is default directory.
+
+See [example `Pipeline`](./git-ssh-auth.md#example-pipeline)
+
 [git-ref]: https://git-scm.com/book/en/v2/Git-Internals-Git-References
 [git-merge]: https://git-scm.com/docs/git-merge
 [git-cherry-pick]: https://git-scm.com/docs/git-cherry-pick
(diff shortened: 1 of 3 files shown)


chmouel:
Hi Jonas, thanks for submitting this,

Maybe I am misunderstanding how immutable infrastructure should look like but I am not sure what does it bring more from how we do natively with pipeline, i.e: https://github.com/tektoncd/pipeline/blob/master/docs/auth.md#using-ssh-authentication-in-your-own-git-tasks 

If it's still something we want, maybe we can expand the git-clone task instead of having a new task ?

(and side note, I think we want to be OS agnostic and not target OSX pbcopy/pbpaste in the documentation)

jlpettersson:
>Maybe I am misunderstanding how immutable infrastructure should look like but I am not sure what does it bring more from how we do natively with pipeline, i.e: https://github.com/tektoncd/pipeline/blob/master/docs/auth.md#using-ssh-authentication-in-your-own-git-tasks

This is a setup where Tekton does not _touch_ or move the `Secrets`. The user declare them, and Kubernetes mount them - Tekton is only involved in Task orchestration.

>If it's still something we want, maybe we can expand the git-clone task instead of having a new task ?

It serves as an alternative. It is easier to handle secrets this way, in my opinion. But I am also fine with not letting it in to the Catalog. Since the authentication works in a different way, it is harder to converge with the existing _git-clone_ since that is it's own Go and custom image. This is an alternative solution for the same job.

>(and side note, I think we want to be OS agnostic and not target OSX pbcopy/pbpaste in the documentation)

That is a good point. I should update that section, if we want this in the Catalog.

PR #338, opened 2020-05-27
Migrate away from PipelineResources in kn task

mgencur:
# Changes

According to https://github.com/tektoncd/pipeline/blob/master/docs/migrating-v1alpha1-to-v1beta1.md#replacing-pipelineresources-with-tasks the PipelineResources stay at alpha version. This PR moves the `kn` task to using Tasks instead of PipelineResources where sources and images are shared between tasks rather than through the PipelineResources custom resource. This is a more straightforward approach which can be seen on the number of lines added vs. removed in this PR.

This also fixes documentation for the `kn` task - fix links to us v1beta1, fix some typos, update examples to use v1beta1.

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/kn/knative-dockerfile-deploy/build_deploy/README.md
+++ b/kn/knative-dockerfile-deploy/build_deploy/README.md
@@ -46,7 +46,7 @@ spec:
       value: ""
     - name: deleteExisting
       value: "true"
-  - name: build-helloworld
+  - name: buildah-build
     taskRef:
       name: buildah
     runAfter:
@@ -61,7 +61,7 @@ spec:
     taskRef:
       name: kn
     runAfter:
-      - build-helloworld
+      - buildah-build
     params:
     - name: kn-image
       value: "gcr.io/knative-nightly/knative.dev/client/cmd/kn"
--- a/kn/knative-dockerfile-deploy/build_deploy/README.md
+++ b/kn/knative-dockerfile-deploy/build_deploy/README.md
@@ -67,13 +67,7 @@ spec:
       value: "gcr.io/knative-nightly/knative.dev/client/cmd/kn"
     - name: ARGS
       value:
-      - "service"
-      - "create"
-      - "hello"
-      - "--revision-name=hello-v1"
-      - "--image=$(params.IMAGE)"
-      - "--env=TARGET=Tekton"
-      - "--service-account=kn-deployer-account"
+        - "$(params.ARGS)"
 ```
 
  - You can also create this Pipeline using the YAML file present in this repo using
@@ -135,7 +129,15 @@ spec:
       value: "https://github.com/navidshaikh/helloworld-go"
     - name: IMAGE
       value: "quay.io/navidshaikh/helloworld-go"
-
+    - name: ARGS
+      value:
+        - "service"
+        - "create"
+        - "hello"
+        - "--revision-name=hello-v1"
+        - "--image=quay.io/navidshaikh/helloworld-go"
+        - "--env=TARGET=Tekton"
+        - "--service-account=kn-deployer-account"
 ```
 
  - You can also create this PipelineRun using the YAML file present in this repo using
--- a/kn/knative-dockerfile-deploy/build_deploy/README.md
+++ b/kn/knative-dockerfile-deploy/build_deploy/README.md
@@ -8,7 +8,7 @@ This Pipeline clones source code using `git-clone` builds and push the git sourc
 The following Pipeline definition refers:
 - `git-clone` to clone the source code
 - `buildah` and `kn` tasks (we've installed these tasks in One time Setup section)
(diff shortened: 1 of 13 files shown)


kn/knative-dockerfile-deploy/build_deploy/README.md
```diff
@@ -46,63 +46,53 @@ spec:
       value: ""
     - name: deleteExisting
       value: "true"
-  - name: buildah-build
+  - name: build-helloworld
```
savitaashture, line 49: Can we keep the original name `buildah-build` because `build-helloworld` indicates specific to `helloworld`

mgencur: The `kn-service-create` task below is specific to `helloworld` as well. So we would use a generic name in the first task but then in the other task it would be specific to helloworld anyway. Any thoughts?

mgencur: Alright. The original was a little different. The Pipeline was generic and the PipelineRun made it specific to helloworld. Now it's a bit different. The Pipeline itself is more specific. There's this problem with passing params.IMAGE we discussed on slack. I guess the ARGS could be moved back to PipelineRun but then the args would have to hardcode the image name like this:
```
- name: IMAGE
  value: "quay.io/navidshaikh/helloworld-go"
- name: ARGS
  value:
      - "service"
      - "create"
      - "hello"
      - "--revision-name=hello-v1"
      - "--image=quay.io/navidshaikh/helloworld-go"
      - "--env=TARGET=Tekton"
      - "--service-account=kn-deployer-account"
```
How does that look?

mgencur: I'd say this looks quite OK.

savitaashture: I see 
I don't have any objection such :smiley: lets keep this way then

kn/knative-dockerfile-deploy/build_deploy/README.md
```diff
@@ -32,6 +29,9 @@ spec:
   - name: GIT_URL
     type: string
     description: Git url to clone
+  - name: IMAGE
+    type: string
+    description: The application image built by Buildah and used by kn task.
```
navidshaikh, line 34:
```suggestion
    description: The application image to be built by Buildah and deployed by kn task.
```

kn/knative-dockerfile-deploy/build_deploy/README.md
mgencur: @navidshaikh WDYT about the suggestion above? It would keep the `Pipeline` quite generic (not specific to helloworld). Hardcoding the image name on two lines in the single file sounds ok-ish. We can't use `--image=$(params.IMAGE)` there because it won't resolve the variable.

navidshaikh: Yes, the pipeline should be generic and able to function for different set of inputs.
PipelineRun can be edited for different runs. IMO, its okay to repeat the image name until we've better provision for referencing parameters in PipelineRun itself.
+1 for bringing back the ARGS to PipelineRun.

mgencur: Added a commit that fixes that.

kn/knative-dockerfile-deploy/service_update/README.md
```diff
@@ -40,67 +33,42 @@ spec:
...
 ```
 
 ## PipelineRun
 
-- Create the PipelineRun to trigger the Pipeline and input the kn CLI
-parameters to deploy a new Revision to `hello` Service
+- Create the PipelineRun to trigger the Pipeline and input the kn CLI parameters to 
+deploy a new Revision to `hello` Service. Note that in the below example, we're referencing a 
+different image, other than the one we created earlier using buildah.
+                                                                                                                            other than the one we created earlier using buildah.
```
navidshaikh, line 44:
```suggestion
- Create the PipelineRun to trigger the Pipeline and input the kn CLI parameters to 
deploy a new Revision to `hello` Service. Note that in the below example, we're referencing a 
different image, other than the one we created earlier using buildah.
```

navidshaikh: remove repeated `other than the one we created earlier using buildah.`

kn/knative-dockerfile-deploy/build_deploy/README.md
```diff
@@ -46,29 +46,22 @@ spec:
       value: ""
     - name: deleteExisting
       value: "true"
-  - name: buildah-build
+  - name: build-helloworld
```
navidshaikh, line 49:
```suggestion
  - name: builldah-build
```

the name can be reverted as the pipeline is now generic

PR #351, opened 2020-06-04
Create a new Prettier task version 1

YolandaDu1997:
# Changes

Create a new Prettier task to check codes from a git repo (need to use the git-clone task to fetch codes).

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/prettier/README.md
+++ b/prettier/README.md
@@ -1,0 +1,18 @@
+# Prettier Tasks
+
+This `Task` is for running [Prettier](https://prettier.io/) code formatter against the provided source code in the Workspace. By default this `Task` will check the format of files in the root of your Workspace.
+
+## Workspaces
+
+ - **source**: A workspace for this Task to fetch the source the code. 
+ 
+## Parameters
+
+ - **args**: Prettier arguments used to run Prettier. See [Prettier CLI](https://prettier.io/docs/en/cli.html) for the arguments. (_default:_ `--check .`)
+ - **prettierImage**: Prettier image used to run Prettier. (_default:_ elnebuloso/prettier:latest)
+
+## Usage
+
+The following pipeline uses the `prettier-check` `Task` to format the file `build.yaml` from [`catalog/golang`](https://github.com/tektoncd/catalog/tree/v1beta1/golang). The pipeline has two tasks: the [`git-clone`](https://github.com/tektoncd/catalog/blob/v1beta1/git/git-clone.yaml) `Task` is the first task to clone the repository to the "shared-workspace"; the `prettier-check` `Task` runs after `git-clone` `Task` to format the source code.
+
+ - [Formatting a file](./tests/run.yaml)
--- a/prettier/prettier.yaml
+++ b/prettier/prettier.yaml
@@ -1,0 +1,26 @@
+apiVersion: tekton.dev/v1beta1
+kind: Task
+metadata:
+  name: prettier
+spec:
+  description: Run the prettier code formatter against provided source code. See https://prettier.io for more information.
+  workspaces:
+  - name: source
+    mountPath: /source
+  params:
+  - name: args
+    type: array
+    description: Prettier arguments used to run Prettier.
+    default:
+      - "--check"
+      - "."
+  - name: prettierImage
+    type: string
+    description: Prettier image used to run Prettier.
+    default: elnebuloso/prettier:latest
+  steps:
+  - name: check
+    image: $(params.prettierImage)
+    workingDir: "$(workspaces.source.path)"
+    command: [prettier]
+    args: ["$(params.args[*])"]
--- a/prettier/tests/pre-apply-task-hook.sh
+++ b/prettier/tests/pre-apply-task-hook.sh
@@ -1,0 +1,4 @@
+#!/bin/bash
+
+# Add git-clone
+kubectl -n ${tns} apply -f ./git/git-clone.yaml
--- a/prettier/tests/run.yaml
+++ b/prettier/tests/run.yaml
@@ -1,0 +1,46 @@
+---
+apiVersion: tekton.dev/v1beta1
+kind: Pipeline
+metadata:
+  name: prettier-pipeline
+spec:
+  workspaces:
+  - name: shared-workspace
+    # description: The git repo will be cloned into this workspace and the config-defaults.yaml will be read from it.
+  tasks:
+  - name: fetch-repo
+    taskRef:
+      name: git-clone
+    workspaces:
+    - name: output
+      workspace: shared-workspace
+    params:
+    - name: url
+      value: https://github.com/tektoncd/catalog.git
+    - name: revision
+      value: 13037cbd0d8131dd05fe2eae92e8a5156807ff5c
+  - name: check-or-format
+    taskRef:
+      name: prettier
+    runAfter:
+    - fetch-repo
+    workspaces:
+    - name: source
+      workspace: shared-workspace
+    params: 
+    - name: args
+      value: 
+        - --write 
+        - "golang/build.yaml"
+---
+apiVersion: tekton.dev/v1beta1
+kind: PipelineRun
+metadata:
+  name: prettier-pipeline-run
+spec:
+  pipelineRef:
+    name: prettier-pipeline
+  workspaces:
+  - name: shared-workspace
+    persistentVolumeClaim:
+      claimName: workspace-prettier-pvc
--- a/prettier/tests/resources.yaml
+++ b/prettier/tests/resources.yaml
@@ -1,0 +1,10 @@
+kind: PersistentVolumeClaim
+apiVersion: v1
+metadata:
+  name: workspace-prettier-pvc
+spec:
+  accessModes:
+    - ReadWriteOnce
+  resources:
+    requests:
+      storage: 500Mi


prettier/prettier-check.yaml
```diff
@@ -0,0 +1,27 @@
...
+  - name: source
+    mountPath: /source
+  params:
+  - name: subdirectory
+    type: string
+    description: Subdirectory inside "source" workspace that contains the target files.
+    default: "."
+  - name: filepath
+    type: string
+    description: path of file of directory inside subdirectory that contains the target file(s).
```
sbwsg, line 16: Did you mean "file or directory" instead of "file of directory" here?

prettier/prettier-check.yaml
```diff
@@ -0,0 +1,27 @@
+apiVersion: tekton.dev/v1beta1
+kind: Task
+metadata:
+  name: prettier-check-from-gitrepo
```
sbwsg, line 4: This Task doesn't require a git repo does it?  It looks like it'll work on any code provided in a workspace.  Suggest `prettier-check` as the Task name if that's the case, dropping the "from-gitrepo" bit.

prettier/prettier-check.yaml
```diff
@@ -0,0 +1,27 @@
...
+spec:
```
sbwsg, line 5: I suggest adding a `description` field under the `spec` that describes in plain english what this Task (and prettier) does.  Something like the following:

```
description: Run the prettier code formatter against provided source code. See https://prettier.io for more information.
```

prettier/prettier-check.yaml
YolandaDu1997: Yeah, sorry it is a typo

prettier/prettier-check.yaml
YolandaDu1997: I do require a git repo. I notice that there is already a git-clone task in the catalog, which will fetch the code from the git repo. Should I add prettier-check task which acts the same as git-clone?

YolandaDu1997:
It depends on the user's input. The default mode is just to check the code without formatting. But the user can define their own options to format the code from the parameter "options".

prettier/prettier-check.yaml
```diff
@@ -0,0 +1,27 @@
...
+  workspaces:
+  - name: source
+    mountPath: /source
+  params:
+  - name: subdirectory
+    type: string
+    description: Subdirectory inside "source" workspace that contains the target files.
+    default: "."
+  - name: filepath
```
wlynch, line 14: Looking at the docs, it looks like you can provide multiple paths:

```
prettier [options] [file/dir/glob ...]
```

We may want to make some small changes to allow multiple paths:
1. rename `options` to `args`, default it to `--check .`
2. set `workingDir` to the workspace path (this way all paths can be relative to the workspace)

prettier/prettier-check.yaml
wlynch: The cool thing about workspaces is that you don't need to know if it's a git repo or not! :D 
You could also have a workspace that's populated by a GCS bucket, a hg/svn repo, or anything else!

All this task needs to care about is that some filesystem is being passed to it.

chmouel:
Thanks @YolandaDu1997 for contributing this, is there any chance if you can look if it's not too difficult to add a test for this ? there is some examples in the repository and some [documentation](https://github.com/tektoncd/catalog/blob/v1beta1/CONTRIBUTING.md#end-to-end-testing) that should get you started...

prettier/prettier-check.yaml
```diff
@@ -0,0 +1,22 @@
+apiVersion: tekton.dev/v1beta1
+kind: Task
+metadata:
+  name: prettier-check
```
wlynch, line 4: nit: rename this from `prettier-check`  -> `prettier` since this might do more than check.

prettier/examples/prettier-format-from-gitrepo.yaml
```diff
@@ -0,0 +1,54 @@
+kind: PersistentVolumeClaim
```
wlynch, line 1: If you move this file from `examples` -> `tests` this will also double as e2e tests! :D

prettier/examples/prettier-format-from-gitrepo.yaml
```diff
@@ -0,0 +1,54 @@
...
+  tasks:
+  - name: fetch-repo
+    taskRef:
+      name: git-clone
+    workspaces:
+    - name: output
+      workspace: shared-workspace
+    params:
+    - name: url
+      value: https://github.com/tektoncd/pipeline.git
```
wlynch, line 29: Let's pin the revision fetched to the SHA of this commit. We don't want this test breaking without an explicit change.

prettier/README.md
```diff
@@ -0,0 +1,17 @@
...
+
+This `Task` is for running [Prettier](https://prettier.io/) code formatter against the provided source code in the Workspace. By default this `Task` will check the format of files in the root of your Workspace.
+
+## Workspaces
+
+ - **source**: A workspace for this Task to fetch the source the code. 
+ 
+## Parameters
+
+ - **args**: Prettier arguments used to run Prettier. See [Prettier CLI](https://prettier.io/docs/en/cli.html) for the arguments. (_default:_ --check .)
```
wlynch, line 11: nit: 
```
`--check .`
```

prettier/examples/prettier-format-from-gitrepo.yaml
YolandaDu1997: I'm not sure what you mean. Does this mean that this way would probably lead to this test breaking?

wlynch: Talked about this more in chat. Intent here is to pin the SHA being used so that an unexpected change to master doesn't cause this test to suddenly fail.

YolandaDu1997:
> I don't think we should include a copy of the git-clone task here. We should be able to use a [`pre-apply-task-hook.sh`](https://github.com/tektoncd/catalog/blob/v1beta1/CONTRIBUTING.md#end-to-end-testing) script to install the git-clone task remotely.
> 
> e.g.
> 
> ```shell
> kubectl apply -f https://raw.githubusercontent.com/tektoncd/catalog/bf7e538778ce6c1ea8ebc72bd0d050362e63a716/git/git-clone.yaml
> ```

Sure. I am working on it.

wlynch:
Looks good! :D 

The only thing left to do is squash all the commits down to 1 commit so the commit history is clean.

Thanks! :)

YolandaDu1997:
> Looks good! :D
> 
> The only thing left to do is squash all the commits down to 1 commit so the commit history is clean.
> 
> Thanks! :)

Sure!

prettier/prettier.yaml
```diff
@@ -0,0 +1,22 @@
...
+  params:
+  - name: args
+    type: array
+    description: Prettier arguments used to run Prettier.
+    default:
+      - "--check"
+      - "."
+  steps:
+  - name: check
+    image: elnebuloso/prettier
```
dibyom, line 19: v minor thing: We are not very consistent about doing this across our tasks but some tasks allow a user to specify the container image as a param (with a default set). This allows the user to use the same task but with say a different image (say, newer version of prettier)

prettier/tests/pre-apply-task-hook.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+
+# Add git-clone
+kubectl -n ${tns} apply -f https://raw.githubusercontent.com/tektoncd/catalog/bf7e538778ce6c1ea8ebc72bd0d050362e63a716/git/git-clone.yaml
```
dibyom, line 4: very cool! Did not know the test framework allowed this!

prettier/tests/run.yaml
```diff
@@ -0,0 +1,45 @@
...
+    runAfter:
+    - fetch-repo
+    workspaces:
+    - name: source
+      workspace: shared-workspace
+    params: 
+    - name: args
+      value: 
+        - --write 
+        - "config/config-defaults.yaml"
```
dibyom, line 33: curious, is this default config coming from the container image? and users are able to specify a prettier config file from the workspace right?

prettier/prettier.yaml
YolandaDu1997: Should I also make the image as a parameter and make this `elnebuloso/prettier` as the default image?

prettier/tests/run.yaml
YolandaDu1997: This is the file from the git repo used by `prettier` to format. It acts as a test file to be formatted in the repository.

prettier/prettier.yaml
dibyom: yeah, that'd be nice. We can set the default to one of the versions from here: https://hub.docker.com/r/elnebuloso/prettier/tags

prettier/tests/run.yaml
dibyom: ahh! ignore my comment! I was thinking of the [prettier config file](https://prettier.io/docs/en/configuration.html) which allows the user to set their linter settings...but it looks like prettier would auto-detect the file based on file naming convention.

prettier/prettier.yaml
YolandaDu1997: Sure!

prettier/tests/pre-apply-task-hook.sh
chmouel: what is the issue doing it directly from the current repository ? like the other task are doing it ? i.e: 

https://github.com/tektoncd/catalog/blob/v1beta1/buildah/tests/pre-apply-task-hook.sh#L7-L8

prettier/tests/run.yaml
```diff
@@ -0,0 +1,45 @@
...
+  tasks:
+  - name: fetch-repo
+    taskRef:
+      name: git-clone
+    workspaces:
+    - name: output
+      workspace: shared-workspace
+    params:
+    - name: url
+      value: https://github.com/tektoncd/pipeline.git
```
chmouel, line 18: I haven't checked how long this run, but should we maybe use a repository that's not as big as `pipeline`? The current catalog tests can get a long time to run and using a smaller git repository may help, i.e: something like a helloworld https://github.com/go-training/helloworld repository?

chmouel:
A few suggestions about the tests and we are almost there for merging 😅

It would be great if you could squash all the commits before we can merge it so we can have a clean history.

chmouel:
Looks good to me, but i'd like to know about if it was a requirements i have missed to have to do remote install of the git-clone task and not installing the git-clone task directly from the local repo, i.e: this comment : 

https://github.com/tektoncd/catalog/pull/351/files#diff-
ddbbbde00b873c8899f4e20314f26fc6R4

we may end up with an old git-clone in the future...

YolandaDu1997:
> Looks good to me, but i'd like to know about if it was a requirements i have missed to have to do remote install of the git-clone task and not installing the git-clone task directly from the local repo, i.e: this comment :
> 
> https://github.com/tektoncd/catalog/pull/351/files#diff-ddbbbde00b873c8899f4e20314f26fc6R4
> 
> we may end up with an old git-clone in the future...

Sorry, I left it out yesterday. I'll fix it.

prettier/tests/run.yaml
wlynch: I would prefer to not depend on an external repo. If/when they make a change to their repo, it might break our tests downstream.

Since it looks like the catalog repo is available for tests anyway, how about we run prettier over the catalog? It has an added benefit that we'll now have a catalog tasks to ensure consistent formatting! 😄

YolandaDu1997: That makes sense. Should I use the commit SHA from the `master` branch or `v1beta1` branch or both okay?

wlynch: I think we may not even need the git-clone task.
Based on https://github.com/tektoncd/catalog/pull/351#issuecomment-642575792 it sounds like the catalog repo is already available to running tests by default? So we may just be able to run the linter on the local path.

You'd have to play around with the test environment to find out for sure.

YolandaDu1997:
> Could you squash the commits to one? Otherwise good to go!



> Could you squash the commits to one? Otherwise good to go!

Done！

chmouel:
/approve

thank you,

(i don't think we need to pin the catalog to a sha)

chmouel:
maven task is failing because of a repo rename from "master" to "main" 🤦 


```
+ /ko-app/git-init -url https://github.com/spring-projects/spring-petclinic -revision master -refspec  -path /workspace/output/ '-sslVerify=true' '-submodules=true' -depth 1
{"level":"error","ts":1592287842.688938,"caller":"git/git.go:41","msg":"Error running git [fetch --recurse-submodules=yes --depth=1 origin --update-head-ok --force master]: exit status 128\nfatal: couldn't find remote ref master\n","stacktrace":"github.com/tektoncd/pipeline/pkg/git.run\n\tgithub.com/tektoncd/pipeline/pkg/git/git.go:41\ngithub.com/tektoncd/pipeline/pkg/git.Fetch\n\tgithub.com/tektoncd/pipeline/pkg/git/git.go:116\nmain.main\n\tgithub.com/tektoncd/pipeline/cmd/git-init/main.go:53\nruntime.main\n\truntime/proc.go:203"}
{"level":"fatal","ts":1592287842.6890328,"caller":"git-init/main.go:54","msg":"Error fetching git repository: failed to fetch [master]: exit status 128","stacktrace":"main.main\n\tgithub.com/tektoncd/pipeline/cmd/git-init/main.go:54\nruntime.main\n\truntime/proc.go:203"}
```

Let's merge https://github.com/tektoncd/catalog/pull/365 first then,

PR #352, opened 2020-06-04
Add Tekton task to include linters

vinamra28:
# Changes

This tekton task can be used to perform lint check on file types:- typescript, yaml, markdown, shell, and k8s field validation. Also this contains prettier which is an opinionate code formatter which verifies whether code is formatted or not.

# Submitter Checklist

review them:

for more details._


sbwsg:
Hey, thanks for the contribution and sorry to be a pain but: I think I'd prefer to see these broken out into separate Tasks for each language rather than having a single large image which contains them.  The reason for this is that if we add more languages then we'll progressively increase the size of the container image, but I as a user will probably only ever use a couple of these in a Pipeline.

So, for me, at least in terms of the Catalog and its Tasks, it doesn't make a lot of sense to bundle together some linters for some languages in one image.  I am curious what others think about this though?

Also, just to let you know that someone is already working on an individual Prettier Task here: https://github.com/tektoncd/catalog/pull/351

wlynch:
+1 to every thing @sbwsg said.

I'd also recommend using canonical images where possible - I'd be hesitant to recommend users using images with unclear support mechanisms if things don't work as expected. This raises an interesting question though - I'm not sure if we have any guidance around where to host images if there is not a canonical repo.

/cc @vdemeester @bobcatfish who may have opinions here.

chmouel:
I agree with what @sbwsg said too!

Thinking out loud, maybe we want to have languages dependent categories, i.e: 

* `python/lint.yaml`
* `ruby/lint.yaml`
* `go/lint.yaml`

with other language specifics tasks in those directory
etc...

something maybe to talk about for next WG meeting, (in an hour and 15mn at the time of writing this comment)

chmouel:
> Thinking out loud, maybe we want to have languages dependent categories, i.e:

We chatted about it at the last WG, I think having different task in the same repo would be fine, since the new catalog organization __should__ handle tags, so it would be best if this can be split up to different tasks like this : 

lint/python.yaml
lint/ruby.yaml
lint/go.yaml
etc...

I guess we will add the tags when we have specs for it...

vinamra28:
@chmouel @sbwsg could you please suggest a possible set of languages to start with for the linter tasks. As I can see there is already a linter task present for `golang` and for `python` we have #364 . So apart from these two languages which all langauges can we proceed with. Please suggest :slightly_smiling_face:

vdemeester:
@iancoffey switched the base branch to master as it got merged with `v1beta1` now :angel:

vinamra28:
@chmouel @sbwsg @vdemeester I just found an official docker image by Github. Although it's not a canonical one but it has support for many languages. 
Link to docker image registry :- https://hub.docker.com/r/github/super-linter
Link to the languages supoorted :- https://github.com/github/super-linter#supported-linters
Please let me know if we can include this as a part of linter or shall we go for the cannonical one only :sweat_smile:

linter/examples/pre-apply-task-hook.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
```
sthaha, line 1: `#!/usr/bin/env bash`

linter/examples/run.yaml
```diff
@@ -0,0 +1,60 @@
...
+        name: git-clone
+      workspaces:
+        - name: output
+          workspace: shared-workspace
+      params:
+        - name: url
+          value: https://github.com/vinamra28/blue-green-deployment-k8s
+        - name: subdirectory
+          value: ""
+        - name: deleteExisting
```
sthaha, line 21: shouldn't this follow the same naming convention use through out the task `delete-existing` ?

linter/examples/pre-apply-task-hook.sh
vinamra28: Thank you @sthaha for the comment but I guess this needs to be changes throughout catalog :sweat_smile: because I took it from existing tekton tasks :sweat_smile:

linter/examples/run.yaml
vinamra28: @sthaha thank you...will change this :slightly_smiling_face: . Anyhow I am working on restructuring this task and breaking them into individual linters.

SM43:
- It would be nice if you could add the task with all the files in the task directory 

```
  ./task/         👈 the kind of the resource 
      /argocd       👈 definition file must have same name
        /0.1
          /README.md     👈 [optional] since there can be a fallback
          /argocd.yaml   👈 the file name must be resource name
        /0.2/...
      /golang-build
        /0.1
          /README.md
          /golang-build.yaml
```
Please see : [#389](https://github.com/tektoncd/catalog/pull/389)

linter/README.md
```diff
@@ -0,0 +1,143 @@
...
+- `shellcheck` (Shell script linter)
+- `rubocop` (Ruby linter)
+- `tslint` (TypeScript linter)
+
+## Installing the Tasks
+
+### Installing the `check-make` Task
+
+```bash
+kubectl apply -f https://raw.githubusercontent.com/tektoncd/catalog/master/linter/check-make.yaml
```
PuneetPunamiya, line 17: Once the task is moved to the task directory can you please change the install path command  as well

PuneetPunamiya:
@vinamra28  I see multiple task here. Can you please split the tasks and have them as separate task with each task having it's respective `yaml`, `readme` and `samples directory` if any.

See : #401

task/ruby-lint/0.1/README.md
```diff
@@ -0,0 +1,82 @@
+# Ruby Linter
+
+The following task is used to provide static analysis on YAML files mounted using `rubocop` (Ruby linter).
+
+## Installing the Task
+
+```bash
+kubectl apply -f https://raw.githubusercontent.com/tektoncd/catalog/master/task/ruby-lint/0.1/rubocop-lint.yaml
```
PuneetPunamiya, line 8: Can you correct the yaml file name here please, it doesn't match the yaml file name

task/check-make/0.1/README.md
```diff
@@ -0,0 +1,82 @@
+# Makefile Linter
+
+The following task is used to provide static analysis on YAML files mounted using `checkmake` (Makefile linter).
```
vdemeester, line 3: Ahaha I love that name :stuck_out_tongue_closed_eyes:

vinamra28:
I guess this PR is now totally ready for review as I have rebased it and the tests are also passing. 
/cc @vdemeester @chmouel

chmouel:
That looks good! checked all the images if they are properly maintained and all seems to be up to date!

thanks

/lgtm
/approve

Looking Good! 🤙🏽

chmouel:
Weird error from docker, hopefully it's only a flake 


```
+ docker build --no-cache -f ./Dockerfile -t localhost:5000/nocode .
unable to resolve docker endpoint: open /certs/client/ca.pem: no such file or directory
```

/retest

PR #357, opened 2020-06-07
create wget-task version1

yuege01:
# Changes

# Submitter Checklist

review them:

for more details._


yuege01:
@wlynch @dibyom

wget/README.md
```diff
@@ -0,0 +1,51 @@
+
+# `wget`
+
+This task performs wget operation to download file from internet to workspace .
```
dibyom, line 4:
```suggestion
This task uses wget  to download files from the internet to a workspace .
```

wget/README.md
```diff
@@ -0,0 +1,51 @@
...
+
+## Install the Task
+
+```
+kubectl apply -f https://raw.githubusercontent.com/tektoncd/catalog/v1beta1/wget/wget.yaml
+```
+
+## Parameters
+
+* **mwendler/wget**: `wget` CLI container image to run this task.
```
dibyom, line 14: Suggest we rename it to `wget-image` or something that makes it clear that this is the container image to be used. Also we should mention the default value for this paramter

wget/wget.yaml
```diff
@@ -0,0 +1,34 @@
...
+     type: string
+     default: ""
+   - name: diroptions
+     description: options of wget to change file location
+     type: array
+     default: []
+   - name: dir
+     description: options of dir
+     type: string
+     default: "$(workspaces.myworkspace.path)"
```
dibyom, line 18: doesn't look like the dir param is being used. let's remove it

wget/wget.yaml
```diff
@@ -0,0 +1,34 @@
...
+metadata:
+  name: wget
+spec:
+  params:
+   - name: url
+     description: URL to wget
+     type: string
+     default: ""
+   - name: diroptions
+     description: options of wget to change file location
```
dibyom, line 12: let's use the same descriptions that you used in the readme :D

wget/wget.yaml
```diff
@@ -0,0 +1,34 @@
...
+   - name: options
+     description: options of wget
+     type: array
+     default: []
+   - name: filename
+     description: name of file we want to change
+     type: string
+     default: ""
+  workspaces:
+   - name: myworkspace
```
dibyom, line 28: small nit: suggest we use a more descriptive name than `myworkspace`. How about `wget-workspace`?

wget/tests/run.yaml
```diff
@@ -0,0 +1,18 @@
+apiVersion: tekton.dev/v1beta1
```
dibyom, line 1: Nice! Thanks for adding the test

wget/README.md
```diff
@@ -0,0 +1,51 @@
...
+      value:
+        - "-P"
+  taskRef:
+    name: wget
+  workspaces:
+    - name: myworkspace
+      emptyDir: {}
+```
+
+Run it with `kubectl create -f wget-run.yaml`
```
dibyom, line 49: nit: the taskrun uses name but the kubectl command usage uses create. We should either switch the taskrun to use `generateName` or use `kubectl apply`

wget/README.md
dibyom: Also, looks like this is not actually a parameter that is used? I think it is ok for this to not be a param. The `wget` command does not change frequently so its less likely that someone might want to use a different image

wget/README.md
yuege01: Thanks, Dibyo, now the description makes more sense!

wget/README.md
yuege01: Ok, I will just delete this line

wget/wget.yaml
```diff
@@ -0,0 +1,30 @@
...
+   - name: filename
+     description: The filename we want to change our file to
+     type: string
+     default: ""
+  workspaces:
+   - name: wget-workspace
+     description: The folder where we write the wget'ed file to  
+  steps:
+    - name: wget
+      image: mwendler/wget
```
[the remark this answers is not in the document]
yuege01: Sure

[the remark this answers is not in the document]
chmouel: Bear in mind that the wget shipped with alpine is a minimal wget builtin to busybox : 

image

so perhaps not something the user would expect....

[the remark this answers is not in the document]
jjasghar: Hmm, good point. We could create a (re)container with like the original one? Maybe put a newer version in the tecktoncd namespace?

(might be a good case study of using tekton paying attention a repository and deploying a new container automagicaly)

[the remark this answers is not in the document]
yuege01: How about for now we add image url as param, so users can specify any versions in taskrun?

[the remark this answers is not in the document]
dibyom: yeah, I don't think we have a good story around hosting the images for catalog yet. To me, using a default image and letting the user override sounds reasonable.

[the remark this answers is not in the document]
chmouel: What about we are using this image as default not ideal because the image purpose is for a different thing and may be slighty bigger but it does have the wget official binary and it's an official image from docker.inc : 


```
[root@openshift2 fedora]# docker run -it buildpack-deps
root@927c2eb6149c:/# wget --version
GNU Wget 1.20.1 built on linux-gnu.

-cares +digest -gpgme +https +ipv6 +iri +large-file -metalink +nls
+ntlm +opie +psl +ssl/gnutls

Wgetrc:
    /etc/wgetrc (system)
Locale:
    /usr/share/locale
Compile:
    gcc -DHAVE_CONFIG_H -DSYSTEM_WGETRC="/etc/wgetrc"
    -DLOCALEDIR="/usr/share/locale" -I. -I../../src -I../lib
    -I../../lib -Wdate-time -D_FORTIFY_SOURCE=2
    -I/usr/include/p11-kit-1 -DHAVE_LIBGNUTLS -DNDEBUG -g -O2
    -fdebug-prefix-map=/build/wget-EiPT9d/wget-1.20.1=.
    -fstack-protector-strong -Wformat -Werror=format-security
    -DNO_SSLv2 -D_FILE_OFFSET_BITS=64 -g -Wall
Link:
    gcc -I/usr/include/p11-kit-1 -DHAVE_LIBGNUTLS -DNDEBUG -g -O2
    -fdebug-prefix-map=/build/wget-EiPT9d/wget-1.20.1=.
    -fstack-protector-strong -Wformat -Werror=format-security
    -DNO_SSLv2 -D_FILE_OFFSET_BITS=64 -g -Wall -Wl,-z,relro -Wl,-z,now
    -lpcre2-8 -luuid -lidn2 -lnettle -lgnutls -lz -lpsl ftp-opie.o
    gnutls.o http-ntlm.o ../lib/libgnu.a

Copyright (C) 2015 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later
<http://www.gnu.org/licenses/gpl.html>.
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.

Originally written by Hrvoje Niksic <hniksic@xemacs.org>.
Please send bug reports and questions to <bug-wget@gnu.org>.
```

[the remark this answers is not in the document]
jjasghar: Honest question; We have `https://hub.docker.com/u/tektoncd` couldn't we just create a `tektoncd/wget` alpine image and host it there?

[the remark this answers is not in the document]
dibyom: Nice! Yeah using a pinned version of `buildpack-deps` sounds good to me. Looks like they specifically have a `curl` tag that only contains `wget` and `curl`.:
```
curl
This variant includes just the curl, wget, and ca-certificates packages. This is perfect for cases like the Java JRE, where downloading JARs is very common and necessary, but checking out code isn't.
```

@jjasghar We could host the image too but then we'd have to figure out a way to keep it updated on a regular basis. I'd lean towards using an already maintained image if possible!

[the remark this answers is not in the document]
chmouel: > Honest question; We have https://hub.docker.com/u/tektoncd couldn't we just create a tektoncd/wget alpine image and host it there?

Yes we could, in the plumbing we have already a couple of them cc @afrittoli 

but we really need a policy/story around images and how we want to manage them/generate them and if we want to provide/support them (as an organisation)

(or after reading @dibyom 's comment that's exactly what i meant)

wget/wget.yaml
```diff
@@ -0,0 +1,34 @@
...
+     type: array
+     default: []
+   - name: filename
+     description: The filename we want to change our file to
+     type: string
+     default: ""
+   - name: wget-option
+     description: The wget docker image to be used
+     type: string
+     default: "buildpack-deps:stable-curl"
```
chmouel, line 26: Did you mean stable-wget, here instead?

chmouel:
/hold

i think we need the wget image..

wget/wget.yaml
yuege01: No, it is actually stable-curl, because this image does not have a wget flag version, its wget is included in this curl image. Dibyo already mentioned this in the comment above.

yuege01: I've also ran the test to make sure it works

chmouel: ah okay! thanks for checking!

chmouel:
/hold cancel
/approve

thank you!

PR #359, opened 2020-06-08
Add gcs-generic Task and update README.md with a TaskRun example

dorismeixing:
# Changes

# Submitter Checklist

review them:

for more details._


dorismeixing:
Add reviewers @dibyom @wlynch

gcs/gcs-generic.yaml
```diff
@@ -0,0 +1,42 @@
+apiVersion: tekton.dev/v1beta1
+kind: Task
+metadata:
+ name: gcs-generic
+spec:
+  workspaces:
+  - name: credentials
+    description: A secret with a service account key to use as GOOGLE_APPLICATION_CREDENTIALS.
+  params:
+  - name: command
```
wlynch, line 10: We can probably combine command and subcommand into a single args param. This might make it easier if there are any actual subcommands (e.g. `acl get` vs command args).

gcs/gcs-generic.yaml
```diff
@@ -0,0 +1,42 @@
...
+  - name: subcommand
+    description: The subcommand line that you would like to enter.
+    type: array 
+  - name: serviceAccountPath
+    description: |
+      The path inside the credentials workspace to the GOOGLE_APPLICATION_CREDENTIALS key file.
+    type: string
+    default: service_account.json
+  steps:
+  - name: run-gsutil
```
wlynch, line 22: This doesn't actual run the gsutil command. We should either rename this or combine this.

Leaning towards combining since idk if there's added value in splitting this up into 2 containers if they both use the same image.

gcs/gcs-generic.yaml
```diff
@@ -0,0 +1,42 @@
...
+    image: google/cloud-sdk
```
wlynch, line 23: Let's add another parameter that defaults to this to allow users to pass in their own image.

This will help if people want to pin to certain gsutil versions.

gcs/gcs-generic.yaml
```diff
@@ -0,0 +1,42 @@
...
+      set -xe
+
+      CRED_PATH="$(workspaces.credentials.path)/$(params.serviceAccountPath)"
+
+      if [[ -f "$CRED_PATH" ]]; then
+        GOOGLE_APPLICATION_CREDENTIALS="$CRED_PATH"
+      fi
+
+      if [[ "${GOOGLE_APPLICATION_CREDENTIALS}" != "" ]]; then
+        echo GOOGLE_APPLICATION_CREDENTIALS is set, activating Service Account...
```
wlynch, line 35: nit: Add quotes e.g. `echo "GOOGLE_APPLICATION_CREDENTIALS is set, activating Service Account..."`

gcs/gcs-generic.yaml
dibyom: that's a good idea. We should do it for the other gcs tasks too 😃  what do you think @sbwsg?

gcs/gcs-generic.yaml
dibyom: there isn't a good way to combine unfortunately. We need script mode here for the auth activate. And we need command and args fields since array params are expanded if used within a [larger string array](https://github.com/tektoncd/pipeline/blob/master/docs/tasks.md#substituting-array-parameters)

gcs/gcs-generic.yaml
dibyom: (Don't have strong opinions on this though...so open to changing if you feel so!)

gcs/README.md
```diff
@@ -244,3 +244,33 @@ spec:
...
+  - name: credentials
+    secret: 
+      secretName: release-secret
+  params:
+  - name: command
+    value: ls
+  - name: subcommand
+    value: 
+     - "-l"
+     - "gs://"
```
dibyom, line 275: let's provide a full gcs bucket something like `gs://tekton-releases` should work! (or even `gs://my-cool-bucket`)

gcs/gcs-generic.yaml
dorismeixing: Keep the two params approach but update the names as "command" and "options"

gcs/gcs-generic.yaml
dorismeixing: It will get error if combine these two, so we will have to split it into two contianers.

gcs/gcs-generic.yaml
dorismeixing: added a parameter for image and set default as "google/cloud-sdk"

chmouel:
> One potential roadblock here: The test Taskrun requires a workspace that contains a GCP service account key (usually stored as a `Secret`). Is there a way to set this up using a `pre-task-hook`? Is there a common secret or something that can be used?

We could probably add something to help with this in the "test-runner", 

I didn't get to implement this since I would have rather not having to put the burden onto `plumbing` repo maintainers to have to manage the creds rotation/security/update etc.. this and have to deal with potential external dependencies which probably never happen with Google Cloud but may not be the case for other tasks that require an external account

I usually would rather encourage the task author to have a task test with a sidecar implementation of the server (ie: a container registry server for kaniko/buildah/jjb etc..) so we don't have to care much about crerds and external dependencies.

I guess in some case (like here) there is not that much way around it and we should have Secret support for it.

Maybe a topic worth to chat about at the next catalog WG?

gcs/README.md
```diff
@@ -244,3 +244,33 @@ spec:
   - name: bucketName
     value: gs://my-fancy-new-test-bucket-12345
 ```
+
+### `gcs-generic`
+
+This TaskRun uses the gcs-generic Task to list all the objects in the directory from a service account.
+
+```yaml
```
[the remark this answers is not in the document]
dibyom: Remove the bit about ServiceAccount. Not needed for the example.

chmouel:
/approve

thank you for your contribution,

chmouel:
maven task is failing because of a repo rename from "master" to "main" 🤦 


```
+ /ko-app/git-init -url https://github.com/spring-projects/spring-petclinic -revision master -refspec  -path /workspace/output/ '-sslVerify=true' '-submodules=true' -depth 1
{"level":"error","ts":1592287842.688938,"caller":"git/git.go:41","msg":"Error running git [fetch --recurse-submodules=yes --depth=1 origin --update-head-ok --force master]: exit status 128\nfatal: couldn't find remote ref master\n","stacktrace":"github.com/tektoncd/pipeline/pkg/git.run\n\tgithub.com/tektoncd/pipeline/pkg/git/git.go:41\ngithub.com/tektoncd/pipeline/pkg/git.Fetch\n\tgithub.com/tektoncd/pipeline/pkg/git/git.go:116\nmain.main\n\tgithub.com/tektoncd/pipeline/cmd/git-init/main.go:53\nruntime.main\n\truntime/proc.go:203"}
{"level":"fatal","ts":1592287842.6890328,"caller":"git-init/main.go:54","msg":"Error fetching git repository: failed to fetch [master]: exit status 128","stacktrace":"main.main\n\tgithub.com/tektoncd/pipeline/cmd/git-init/main.go:54\nruntime.main\n\truntime/proc.go:203"}
```

Let's merge https://github.com/tektoncd/catalog/pull/365 first then,

Issue #379, opened 2020-06-22
Catalog Integration Tests are failing


dibyom:
# Expected Behavior
All tests pass

# Actual Behavior

git-batch-merge-test-mode-merge-cherry-pick- seems to be the one failing consistently across many PRs:
- https://github.com/tektoncd/catalog/pull/378
- https://github.com/tektoncd/catalog/pull/375
etc

/kind bug
/area testing

dibyom:
The test succeeds locally. I looked at the logs in StackDriver for #375: https://gist.github.com/dibyom/b52930b16e05eb9a1c889a46600f3288


The last message is that the step has finished executing:
```
  "message": "All Steps have completed executing"
```

So, looks like the task actually succeeded but the test runner thought it failed?

dibyom:
snippet for failing log:

```
+ breakit=
+ for status in ${all_status}
+ [[ True == *ERROR ]]
+ [[ Pending Pending Pending Running Running Running Running Pending Succeeded == *Fail* ]]
+ [[ Pending Pending Pending Running Running Running Running Pending Succeeded == Couldnt* ]]
+ [[ True == True ]]
+ breakit=True
+ [[ True == True ]]
+ echo -n 'SUCCESS: git pipelinerun has successfully executed: '
SUCCESS: git pipelinerun has successfully executed: ++ kubectl get pod -o name -n git-545
+ for pod in $(kubectl get pod -o name -n ${tns})
+ kubectl logs --all-containers -n git-545 pod/git-batch-merge-test-mode-merge-cherry-pick-pod-xzjt5
Error from server (BadRequest): container "step-clone" in pod "git-batch-merge-test-mode-merge-cherry-pick-pod-xzjt5" is waiting to start: PodInitializing
+ teardown_test_resources
+ ((  IS_BOSKOS  ))
+ return
2020/06/19 17:22:08 process.go:155: Step '/home/prow/go/src/github.com/tektoncd/catalog/test/e2e-tests.sh --run-tests' finished in 4m59.250225372s
2020/06/19 17:22:08 main.go:319: Something went wrong: encountered 1 errors: [error during /home/prow/go/src/github.com/tektoncd/catalog/test/e2e-tests.sh --run-tests: exit status 1]
++ get_test_return_code
```

PR #383, opened 2020-06-23
test-runner: Do not show logs on success

chmouel:
# Changes

This can grow the log pretty big with message that we probably don't care
about.

when we show the logs on failure, make sure we are not failing if we can't show
them. Sometime some container may be in "Terminated" state until k8 cleans them
up. (i.e: affinity-assistant) so no need to worry if those are failing/exiting before
showing the next container log.

# Submitter Checklist

review them:


chmouel:
@dibyom  this should be a followup to further fix https://github.com/tektoncd/catalog/issues/379

Issue #373, opened 2020-06-18
Create Tekton based Pipeline which has capability to run against multiple tekton pipelines versions


bobcatfish:
# Expected Behavior

* Tasks should be able to declare what versions of Tekton Pipelines they are compatible with (https://github.com/tektoncd/pipeline/issues/2588)
* The catalog test infrastructure can use this information to determine what versions of Tekton Pipelines to test against

# Actual Behavior

All catalog Tasks are tested against 1 version of Tekton Pipelines, hard coded at https://github.com/tektoncd/catalog/blob/v1beta1/test/e2e-tests.sh#L19

# Additional Info

[Overall catalog test infrastructure proposal](https://docs.google.com/document/d/1-czjvjfpuIqYKsfkvZ5RxIbtoFNLTEtOxaZB71Aehdg/edit)

chmouel:
That's a good idea, i can pick this up if you like.

I wonder what's is the best way to do this tho. Since pipeline supports does not support per namespace install.

Perhaps something like this : 

- Collect all pipeline versions of all tasks in catalog
- for each pipeline versions uninstall previous and install the new pipeline version.
- Run all catalog tasks that want that specifics version against it.

I think that's going to exponentially add a lot of time to each run. 

Perhaps we want to start to implement the feature to only test tasks that has been changed/added first?

bobcatfish:
Hey @chmouel ! There's a sketch for how this could look [in the test infra design doc](https://docs.google.com/document/d/1-czjvjfpuIqYKsfkvZ5RxIbtoFNLTEtOxaZB71Aehdg/edit#heading=h.hzcpipuqmb1u) - i'm (slowly) working on a prototype, my biggest goal is to do this using tekton pipelines

> 1. Collect all pipeline versions of all tasks in catalog
> 2. for each pipeline versions, un-install the previous and install the new pipeline version.

For (2) I want to do that with a Pipeline that takes a version as an argument - also if we use boskos we can create multiple clusters and install whatever versions we want. If we use a completely separate GCP project for each one we're going to run out pretty quickly tho.

This is the psuedo pipeline code I've got so far:

```yaml
apiVersion: tekton.dev/v1beta1
kind: Pipeline
metadata:
  name: catalog-test
spec:
  params:
    - name: pipelines-version
      default: "v0.13.2"
      description: The verision of tekton-pipelines to install and test with
    - name: task-paths
      description: A comma separated lists of the paths to the catalog tasks to test
  workspaces:
[...]

PR #385, opened 2020-06-23
Test tasks that only has been modified or added

chmouel:
# Changes

We are nnow only doing integration tests on the tasks that has been modified or added. This will improve the speed of each PR iteration not testing everything while the PR modify only some part of it.

/cc @sthaha 
if you know if there is any other git trickery i can use to detect
file changes between the 'main' branch and the PR

/cc @bobcatfish 
cf #373 discussion

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/test/e2e-tests.sh
+++ b/test/e2e-tests.sh
@@ -19,6 +19,7 @@ set -x
 export RELEASE_YAML=https://github.com/tektoncd/pipeline/releases/download/v0.17.1/release.yaml
 
 source $(dirname $0)/../vendor/github.com/tektoncd/plumbing/scripts/e2e-tests.sh
+source $(dirname $0)/../vendor/github.com/tektoncd/plumbing/scripts/presubmit-tests.sh
 source $(dirname $0)/e2e-common.sh
 
 TMPF=$(mktemp /tmp/.mm.XXXXXX)
@@ -29,7 +30,7 @@ trap clean EXIT
 [[ -z ${LOCAL_CI_RUN} ]] && {
 
     # Initialize cluster
-    initialize $@
+    initialize "$@"
 
     # Install the latest Tekton CRDs.
     install_pipeline_crd
@@ -48,10 +49,29 @@ TEST_YAML_IGNORES=${TEST_YAML_IGNORES:-""}
 # test for example "s2i"
 TEST_TASKRUN_IGNORES=${TEST_TASKRUN_IGNORES:-""}
 
+# Define this variable if you want to run all tests and not just the modified one.
+TEST_RUN_ALL_TESTS=${TEST_RUN_ALL_TESTS:-""}
+
 set -ex
 set -o pipefail
 
-test_yaml_can_install
-test_task_creation task/*/*/tests
+all_tests=$(echo task/*/*/tests)
+
+if [[ -z ${TEST_RUN_ALL_TESTS} ]];then
+    # Defining WORK_DIR so we can get list_changed_files function working
+    # properly.
+    WORK_DIR="$(mktemp -d)"
+    export WORK_DIR
+
+    all_tests=$({ grep '/tests/[^/]*yaml' "$(list_changed_files)" || true; } | sed 's/\(.*\)\/.*/\1/')
+    [[ -z ${all_tests} ]] && {
+        echo "No tests has been detected in this PR. exiting."
+        success
+    }
+fi
+
+test_yaml_can_install "${all_tests}"
+
+test_task_creation "${all_tests}"
 
 success
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -89,6 +89,11 @@ The way the functional tests works is that if you have a directory called
 `tests/` inside the task, it would start creating a random `Namespace`, apply
 the task and then every yaml files that you have in that `tests/` directory.
 
+Note that the test runner for the integration tests will only test the tasks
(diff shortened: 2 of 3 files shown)


chmouel:
Seems to work okay, it only detect the change in curl and wget :

```
+ all_tests='curl/tests
wget/tests'
...
+ test_task_creation curl/tests wget/tests
```

(i had another test that test as well new addition)

I am not testing test_yaml_creation, I will add a further PR to add it but it's not such a big problem currently since those tests are pretty quick to execute compared to the integration tests

I think we would probably want some full integration weekly test run still, we may have set this up from the plumbing repository.

/hold cancel

chmouel:
> So, while I agree this is good to have, I think at the same time we need to setup some way to run all the tests, either on demand, or periodically. Perhaps we should have an "info only" catalog test running on pipeline changes, or one that could be triggered on demand.

Yes agreed this is what i was mentioning in my comment [here](https://github.com/tektoncd/catalog/pull/385#issuecomment-648024899) we would need to have some weekly periodic jobs for this...

chmouel:
Further note, the integration tests before this PR is taking 20mn to run, if we are adding pipeline supported version matrix this time would exponentially grow pretty big

bobcatfish:
> Further note, the integration tests before this PR is taking 20mn to run, if we are adding pipeline supported version matrix this time would exponentially grow pretty big

only if they run one at a time :D

bobcatfish:
i wonder if doing this will prevent us from catching issues - for example if we bump the version of tekton pipelines the tests are using, but we don't run any of the tests, we wont catch any problems 🤔 

maybe we can add a nightly run of the tests to make up for it?

vdemeester:
> i wonder if doing this will prevent us from catching issues - for example if we bump the version of tekton pipelines the tests are using, but we don't run any of the tests, we wont catch any problems thinking
> 
> maybe we can add a nightly run of the tests to make up for it?

Yes, we need to setup a nightly I think :upside_down_face:

chmouel:
I have added 16e1352e0b2544bbdf4cfc2385cc718f4eb89922 so we can define the variable `TEST_RUN_ALL_TESTS` to force tests everything.... 

Now to figure out on how to add a nightly catalog job in `openshift/plumbing`

test/e2e-tests.sh
```diff
@@ -48,10 +49,31 @@ TEST_YAML_IGNORES=${TEST_YAML_IGNORES:-""}
...
+
+# Main branch to detect the file changes against to for which test to run
+MAIN_BRANCH=master
+
 set -ex
 set -o pipefail
 
+
+if [[ -n ${TEST_RUN_ALL_TESTS} ]];then
+    all_tests=$(/bin/ls -1d */tests)
```
sthaha, line 64: why /bin/ls? did you mean `\ls` ?

vdemeester: `/usr/bin/env ls` maybe ? :stuck_out_tongue_closed_eyes: :joy_cat:

sthaha: what I mean was to use `\` - https://www.tldp.org/LDP/GNU-Linux-Tools-Summary/html/c1195.htm#:~:text=Use%20a%20%E2%80%9C%5C%E2%80%9D%20(backslash,run%20it%20without%20the%20alias.&text=The%20backslash%20escape%20character%20can,command%20to%20override%20any%20aliases.

chmouel: I guess we could, i like better the /usr/bin/env approach since less magic (and non-obvious) than the backslash solution (fyi neat, on zsh `=` is a even better shortcut :D)

sthaha: to be explicit i think the recommended way is `command`  e.g. `command ls -al ...`

popcor255:
All tests are isolated in the given namespace. They are created at the begging of tests and cleaned up at the end of tests. Since,  namespace are fairly predictable "local tns="${testname}-${version}". Maybe we should clean up all the resources in the namespace and not the namespace itself when the test ends. We should maybe have a task at the begging of the test suite to clean up old namespaces. At which point we only run tests if the namespace does not exist.
line 107 (e2e-common.sh)
```bash
local tns="${testname}-${version}"
```

I am spit balling here, tbh. Just trying to give a suggesiton.

test/e2e-tests.sh
```diff
@@ -48,10 +49,35 @@ TEST_YAML_IGNORES=${TEST_YAML_IGNORES:-""}
 # test for example "s2i"
 TEST_TASKRUN_IGNORES=${TEST_TASKRUN_IGNORES:-""}
 
+# Define this variable if you want to run all tests and not just the modified one.
+TEST_RUN_ALL_TESTS=${TEST_RUN_ALL_TESTS:-""}
```
vdemeester, line 53: We need to update the [catalog periodic](https://github.com/tektoncd/plumbing/blob/master/prow/config.yaml#L1175) to set this environment variable, right ? If yes, we should do it ahead of time as it is won't affect the current run.

vdemeester (approved):

test/e2e-tests.sh
chmouel: Yep, let me make sure this works tho before remving the DNM 👍

chmouel:
/hold cancel

this was tested as working, i have added previously a commit to this pr  which had a change in the task/trigger-jenkins-job and it properly tested only that PR.

piyush-garg:
lgtm

PR #403, opened 2020-07-09
Update READMEs of tasks using workspaces in place of PipelineResources

sbwsg:
# Changes

The tasks in this commit were updated at some point to use workspaces instead of
pipeline resources but their readmes still refer to the resources.

This commit updates the readmes to remove mention of pipeline resources and link
out to pipeline's workspaces doc.

# Submitter Checklist

review them:

### DIFF
--- a/buildkit/README.md
+++ b/buildkit/README.md
@@ -63,7 +63,7 @@ task.tekton.dev/buildkit created
 
 ## Workspaces
 
-* **source**: A `git`-type `PipelineResource` specifying the location of the source to build.
+* **source**: A [Workspace](https://github.com/tektoncd/pipeline/blob/master/docs/workspaces.md) volume containing the source to build.
 
 ## Resources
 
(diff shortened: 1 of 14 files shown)


sbwsg:
image

😟 sorry folks, did not realise how many people this would pull in

vdemeester:
> image
> 
> worried sorry folks, did not realise how many people this would pull in

that's not a problem :wink:

buildkit/README.md
```diff
@@ -63,7 +63,7 @@ task.tekton.dev/buildkit created
 
 ## Workspaces
 
-* **source**: A `git`-type `PipelineResource` specifying the location of the source to build.
+* **source**: A [Workspace](https://github.com/tektoncd/pipeline/blob/master/docs/workspaces.md) volume containing the source to build.
```
bobcatfish, line 66: total nit but I'd probably drop the word "volume" from these - not sure if a "workspace volume" is a thing? or maybe this is trying to distinguish from secrets/configmaps, etc.? if so, i feel like this might indicate that we should allow Tasks to declare more type info on a workspace to indicate what type is required 🤔  (which might be in line with your credential work?)

sbwsg: Ah, I hadn't put that much thought into it.  I'll make a pr with "volume" removed.

PR #401, opened 2020-07-08
Split tasks based on Catalog Organization Proposal

PuneetPunamiya:
# Changes

* **NOTE**: This patch covers the second part of the issue where it splits the tasks and moves the tasks to the task directory. The first part of the issue is done in PR (#389)

* Currently the catalog has some task in which there are multiple tasks

  For e.g.
  ```markdown
  $ ls git 

  examples  git-batch-merge.yaml  git-cli.yaml  git-clone.yaml  git-rebase.yaml  README.md  tests
  ```

*   Based on the proposal the git task is split as:

    ```
    git-batch-merge
    └── 0.1
        ├── examples
        │   ├── git-cli
        │   │   ├── pipeline.yaml
        │   │   ├── pvc.yaml
        │   │   ├── secret.yaml
        │   │   └── service-account.yaml
        │   ├── git-clone-checking-out-a-branch.yaml
        │   ├── git-clone-checking-out-a-commit.yaml
        │   ├── git-rebase
        │   │   ├── run.yaml
        │   │   ├── secret.yaml
        │   │   └── service-account.yaml
        │   └── using-git-clone-result.yaml
        ├── git-batch-merge.yaml
        ├── README.md
        └── tests
            └── run.yaml

    ```

    ```
    git-cli
    └── 0.1
        ├── examples
        │   ├── git-cli
        │   │   ├── pipeline.yaml
        │   │   ├── pvc.yaml
        │   │   ├── secret.yaml
        │   │   └── service-account.yaml
        │   ├── git-clone-checking-out-a-branch.yaml
        │   ├── git-clone-checking-out-a-commit.yaml
        │   ├── git-rebase
        │   │   ├── run.yaml
        │   │   ├── secret.yaml
        │   │   └── service-account.yaml
        │   └── using-git-clone-result.yaml
        ├── git-cli.yaml
        ├── README.md
        └── tests
            └── run.yaml

    ```

    ```
    git-clone
    └── 0.1
        ├── examples
        │   ├── git-cli
        │   │   ├── pipeline.yaml
        │   │   ├── pvc.yaml
        │   │   ├── secret.yaml
        │   │   └── service-account.yaml
        │   ├── git-clone-checking-out-a-branch.yaml
        │   ├── git-clone-checking-out-a-commit.yaml
        │   ├── git-rebase
        │   │   ├── run.yaml
        │   │   ├── secret.yaml
        │   │   └── service-account.yaml
        │   └── using-git-clone-result.yaml
        ├── git-clone.yaml
[...]

PR #394, opened 2020-06-29
Add generic kubectl with path task

a-roberts:
# Changes

Adds an, IMO, more convenient kubectl task to the catalog for folks to enjoy

# Submitter Checklist

These are the criteria that every PR should meet, please check them off as you
review them:

- [x] Includes [docs](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if user facing)
- [x] Commit messages follow [commit message best practices](https://github.com/tektoncd/community/blob/master/standards.md#commit-messages)
- [ ] Yaml file complies with [yamllint](https://github.com/adrienverge/yamllint) rules.

_See [the contribution guide](https://github.com/tektoncd/catalog/blob/master/CONTRIBUTING.md)
for more details._

Please don't merge yet, will test fully with my example "all of the shelf" pipeline

PR #389, opened 2020-06-24
Reorganize Catalog based on Catalog Organization Proposal

PuneetPunamiya:
# Changes

* **NOTE**: This patch covers the first part of the issue   where it modifies all the tasks and moves the tasks which don't need to be split to the task directory. The task which has to be split and moved to the task directory will be added in the next pr (https://github.com/tektoncd/catalog/pull/401) to make review easier

1. Each resource follows the following structure
```
    ./task/         👈 the kind of the resource 
      /argocd       👈 definition file must have same name
        /OWNERS     👈 owners of this ressource
        /README.md  👈 fallback README if version is missing README
        /0.1
          /README.md     👈 [optional] since there can be a fallback
          /argocd.yaml   👈 the file name must be resource name
          /samples/deploy-to-k8s.yaml
        /0.2/...
      /golang-build
        /0.1
          /README.md
          /golang-build.yaml
          /samples/golang-build.yaml
```
2. Resource YAML file includes following changes
  *  Labels include the version of the resource.
  *  Annotations include `minimum pipeline version` supported by the resource, 
     `tags` associated with the resource and `displayName` of the resource 
    
  ```yaml
    
 labels:
    app.kubernetes.io/version: "0.1"            👈 version of the resource
    
  annotations:
    tekton.dev/pipelines.minVersion: "0.12.1"   👈 will work from x version  of pipeline
    tekton.dev/tags: "ansible, cli"                        👈 Comma separated list of tags 
    tekton.dev/displayName: "Ansible Tower Cli"    
    
spec:
  description: |-
    ansible-tower-cli task simplifies 
    workflow, jobs, manage users...   👈 # Summary
    
    Ansible Tower (formerly ‘AWX’) is a ...
  
  ```

# Submitter Checklist

These are the criteria that every PR should meet, please check them off as you
review them:

- [ ] Includes [docs](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if user facing)
- [x] Commit messages follow [commit message best practices](https://github.com/tektoncd/community/blob/master/standards.md#commit-messages)
- [x] Yaml file complies with [yamllint](https://github.com/adrienverge/yamllint) rules.

_See [the contribution guide](https://github.com/tektoncd/catalog/blob/master/CONTRIBUTING.md)
for more details._

PR #408, opened 2020-07-10
Add Tasks to acquire and release boskos resources 🐑

bobcatfish:
# Changes

Boskos is a tool that allows one to create a pool of cloud projects
(definitely GCP, I think it supports other providers as well), and
manages acquiring, releasing, and cleaning them between leases.

We use it for Tekton test infrastructure for our end to end tests and
we'd like to use it for our catalog Tasks as well.

This commit adds boskos acquire and release Tasks.

The acquire Task also creates a pod in the running cluster to perform
heartbeating so that boskos knows that the resource is still in use.
(Lemme know if you have a better idea for making this work! I have this
slightly crazy idea that Pipeline level sidecars might make sense ....)

The intention of the release Task is that it would be run in a
Pipeline's `finally` clause, however today that would be difficult
because finally Tasks can't yet use the results of other Tasks, but this
functionality is on the way: https://github.com/tektoncd/pipeline/issues/2557

This is part of the work in #373 to create a Pipeline for the catalog.

Thanks to @ixdy for helping me figure this out! This also might be relevant to your interests @afrittoli 

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/task/boskos-acquire/0.1/README.md
+++ b/task/boskos-acquire/0.1/README.md
@@ -1,0 +1,39 @@
+# Boskos Acquire
+
+The `boskos-acquire` Task can be used to acquire cloud projects from a pool with
+[Boskos](https://github.com/kubernetes-sigs/boskos#boskos), by invoking `boskosctl`.
+
+To release projects obtained with `boskos-acquire` can be released with
+[`boskos-release`](../boskos-release).
+
+It is implemented using [`boskosctl`](https://github.com/kubernetes-sigs/boskos/tree/master/cmd/boskosctl).
+
+_The Task assumes already have Boskos up and running. To set it up yourself, you
+can look at [this example deployment](https://github.com/kubernetes-sigs/boskos/tree/master/deployments/overlays/example).
+
+## ServiceAccount
+
+After acquiring a project, the Task will start a running `pod` in your cluster to send
+heartbeat requests to Boskos (Boskos will automatically clean up abandoned clusters).
+This means the Task must be run with a serviceAccount that has the ability to interact
+with `pods` (see [service-account.yaml](samples/service-account.yaml) for an example).
+
+## Parameters
+
+* **server-url**: The URL of the running boskos server. (_default_: http://boskos.test-pods.svc.cluster.local)
+* **type**: The type of resource to request. Resource types are specified in the resource
+            ConfigMap provided to the Boskos server. (_default_: gke-project)
+* **owner-name**: A string that identifies the owner of the leased resource to request. (_required_)
+
+## Results
+
+* **leased-resource**: The name of the leased resource
+
+## Usage
+
+See [samples/pipelinerun.yaml](samples/pipelinerun.yaml) for an example of a Pipeline that obtains
+a resource using Boskos, then waits (this is when you would do whatever you need to do with the resource),
+and finally releases it with [`boskos-release`](../boskos-release).
+
+Boskos doesn't do anything to provide you with the credentials you need to interact with the resource you
+have leased. Setting up and managing these credentials is outside the scope of Boskos's responsibilities.
--- a/task/boskos-acquire/0.1/boskos-acquire.yaml
+++ b/task/boskos-acquire/0.1/boskos-acquire.yaml
@@ -1,0 +1,63 @@
+apiVersion: tekton.dev/v1beta1
+kind: Task
+metadata:
+  name: boskos-acquire
+  labels:
+    app.kubernetes.io/version: "0.1"
+  annotations:
+    tekton.dev/pipelines.minVersion: "0.12.1"
+    tekton.dev/tags: "boskos,test"
+spec:
+  description: |
+    Acquire a project using Boskos.
+
+    The boskos-acquire Task will request a resource of the specified type from the
+    server-url. If successful, it will start a pod that will run the boskosctl heartbeat
+    command. When you are done with the resource, release it with boskos-release.
+  params:
+  - name: server-url
+    description: The URL of the running boskos server
+    default: "http://boskos.test-pods.svc.cluster.local"
+  - name: type
+    description: |
+      The type of resource to request. Resource types are specified in the resource
+      ConfigMap provided to the Boskos server.
+    default: gke-project
+  - name: owner-name
+    description: A string that identifies the owner of the leased resource to request.
+  results:
+  - name: leased-resource
+    description: The name of the leased resource
+  steps:
+  - name: boskosctl-acquire
+    image: gcr.io/k8s-staging-boskos/boskosctl@sha256:a7fc984732c5dd0b4e0fe0a92e2730fa4b6bddecd0f6f6c7c6b5501abe4ab105
+    script: |
+      RESOURCE=$(boskosctl acquire \
+        --server-url=$(params.server-url) \
+        --owner-name=$(params.owner-name) \
+        --type=$(params.type) \
+        --state=free \
+        --target-state=busy)
+      echo $RESOURCE > /workspace/full-resource-output.json
+      echo $RESOURCE | jq -rj ".name" > /tekton/results/leased-resource
+  - name: create-heartbeat-pod-yaml
+    image: lachlanevenson/k8s-kubectl@sha256:3a5e22a406a109f4f26ec06b5f1f6a66ae0cd0e185bc28499eb7b7a3bbf1fe09
+    script: |
+      FULL_RESOURCE_OUTPUT=$(cat /workspace/full-resource-output.json)
+      LEASED_RESOURCE=$(cat /tekton/results/leased-resource)
+      cat <<EOF | kubectl apply -f -
+      apiVersion: v1
+      kind: Pod
+      metadata:
+        name: boskos-heartbeat-$LEASED_RESOURCE
+      spec:
+        containers:
+        - name: heartbeat
+          image: gcr.io/k8s-staging-boskos/boskosctl@sha256:a7fc984732c5dd0b4e0fe0a92e2730fa4b6bddecd0f6f6c7c6b5501abe4ab105
+          args:
+          - heartbeat
+          - --server-url=$(params.server-url)
+          - --owner-name=$(params.owner-name)
+          - --resource=$FULL_RESOURCE_OUTPUT
+          - --period=5m
+      EOF
--- a/task/boskos-acquire/0.1/samples/pipelinerun.yaml
+++ b/task/boskos-acquire/0.1/samples/pipelinerun.yaml
@@ -1,0 +1,34 @@
+apiVersion: tekton.dev/v1beta1
+kind: PipelineRun
+metadata:
+  generateName: do-boskos-
+spec:
+  serviceAccountName: boskos-heartbeat
+  pipelineSpec:
+    params:
+    - name: owner-name
+      default: trying-out-boskos
+    tasks:
+      - name: boskos-acquire
+        taskRef:
+          name: boskos-acquire
+        params:
+          - name: owner-name
+            value: $(params.owner-name)
+      - name: wait
+        runAfter: [boskos-acquire]
+        taskSpec:
+          steps:
+          - image: alpine
+            script: |
+              sleep 1m
+      - name: boskos-release
+        runAfter: [wait]
+        taskRef:
+          name: boskos-release
+        params:
+          - name: leased-resource
+            # TODO(pipeline#2557): This is a good candidate for a finally clause
+            value: $(tasks.boskos-acquire.results.leased-resource)
+          - name: owner-name
+            value: $(params.owner-name)
--- a/task/boskos-acquire/0.1/samples/service-account.yaml
+++ b/task/boskos-acquire/0.1/samples/service-account.yaml
@@ -1,0 +1,29 @@
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: boskos-heartbeat
+  namespace: default
+---
+kind: Role
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: pod-editor
+  namespace: default
+rules:
+  - apiGroups: [""]
+    resources: ["pods"]
+    verbs: ["get", "list", "create", "update", "delete", "patch", "watch"]
+---
+apiVersion: rbac.authorization.k8s.io/v1beta1
+kind: RoleBinding
+metadata:
+  name: boskos-heartbeat-pod-editor
+  namespace: default
+subjects:
+  - kind: ServiceAccount
+    name: boskos-heartbeat
+    namespace: default
+roleRef:
+  kind: Role
+  name: pod-editor
+  apiGroup: rbac.authorization.k8s.io
--- a/task/boskos-acquire/OWNERS
+++ b/task/boskos-acquire/OWNERS
@@ -1,0 +1,2 @@
+approvers:
+- bobcatfish
--- a/task/boskos-release/0.1/README.md
+++ b/task/boskos-release/0.1/README.md
@@ -1,0 +1,31 @@
+# Boskos Release
+
+The `boskos-release` Task can be used to release cloud projects
+which were obtained from a pool with [Boskos](https://github.com/kubernetes-sigs/boskos#boskos)
+via [`boskos-acquire`](../boskos-acquire).
+
+It will mark the resource as `dirty` so that Boskos will clear any obtained resources.
+
+It is implemented using [`boskosctl`](https://github.com/kubernetes-sigs/boskos/tree/master/cmd/boskosctl).
+
+_The Task assumes already have Boskos up and running. To set it up yourself, your
+best bet would be to find an example of it already configured, for example
+[the boskos configuration used for Tekton itself](https://github.com/tektoncd/plumbing/tree/master/boskos)._
+
+## ServiceAccount
+
+[`boskos-acquire`](../boskos-acquire) will start a `pod` in your cluster to maintain the lease
+via a heartbeat. `boskos-release` will delete that pod, and so the Task must be run with a
+serviceAccount that has the ability to delete `pods` (see [service-account.yaml](samples/service-account.yaml) for an example).
+
+## Parameters
+
+* **server-url**: The URL of the running boskos server. (_default_: http://boskos.test-pods.svc.cluster.local)
+* **leased-resource**: The name of the leased resource. (_required_)
+* **owner-name**: A string that identifies the owner of the leased resource to request. (_required_)
+
+## Usage
+
+See [samples/pipelinerun.yaml](samples/pipelinerun.yaml) for an example of a Pipeline that obtains
+a resource using [`boskos-acquire`](../boskos-acquire), then waits (this is when you would do whatever
+you need to do with the resource), and finally releases it with `boskos-release`.
--- a/task/boskos-release/0.1/boskos-release.yaml
+++ b/task/boskos-release/0.1/boskos-release.yaml
@@ -1,0 +1,41 @@
+apiVersion: tekton.dev/v1beta1
+kind: Task
+metadata:
+  name: boskos-release
+  labels:
+    app.kubernetes.io/version: "0.1"
+  annotations:
+    tekton.dev/pipelines.minVersion: "0.12.1"
+    tekton.dev/tags: "boskos,test"
+spec:
+  description: |
+    Release a project acquired using Boskos.
+
+    The boskos-release Task will release the specified resource from the boskos instance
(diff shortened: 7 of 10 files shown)


bobcatfish:
Anyone who'd like to join the OWNERS files with me lemme know :D

boskos-acquire/0.1/README.md
```diff
@@ -0,0 +1,40 @@
...
+The `boskos-acquire` Task can be used to acquire cloud projects from a pool with
+[Boskos](https://github.com/kubernetes-sigs/boskos#boskos), by invoking `boskosctl`.
+
+To release projects obtained with `boskos-acquire` can be released with
+[`boskos-release`](../boskos-release).
+
+It is implemented using [`boskosctl`](https://github.com/kubernetes-sigs/boskos/tree/master/cmd/boskosctl).
+
+_The Task assumes already have Boskos up and running. To set it up yourself, your
+best bet would be to find an example of it already configured, for example
```
ixdy, line 12: There is now an example deployment in https://github.com/kubernetes-sigs/boskos/tree/master/deployments/overlays/example, but other docs are still a bit lacking.

boskos-acquire/0.1/boskos-acquire.yaml
```diff
@@ -0,0 +1,67 @@
...
+      RESOURCE=$(boskosctl acquire \
+        --server-url=$(params.server-url) \
+        --owner-name=$(params.owner-name) \
+        --type=$(params.type) \
+        --state=free \
+        --target-state=busy)
+      echo $RESOURCE > /workspace/full-resource-output.json
+      echo $RESOURCE | jq -rj ".name" > /tekton/results/leased-resource
+  - name: create-boskosctl-heartbeat-pod
+    image: ubuntu
```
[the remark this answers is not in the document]
vdemeester: I was gonna comment that same. We can combined the two step into one indeed :angel:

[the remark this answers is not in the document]
bobcatfish: i got really excited about this and then i realized that we'd need a way to run these lines:

```
      FULL_RESOURCE_OUTPUT=$(cat /workspace/full-resource-output.json)
      LEASED_RESOURCE=$(cat /tekton/results/leased-resource)
```

I guess if we're running cat in the next step, we'd be able to run cat inline?

alternatively: what about using a smaller image here instead of ubuntu, e.g. busy box?

boskos-acquire/0.1/boskos-acquire.yaml
```diff
@@ -0,0 +1,67 @@
+apiVersion: tekton.dev/v1beta1
+kind: Task
+metadata:
+  name: boskos-acquire
+  labels:
+    app.kubernetes.io/version: "0.1"
+  annotations:
+    tekton.dev/pipelines.minVersion: "0.12.1"
+    tekton.dev/tags: "boskos,test"
```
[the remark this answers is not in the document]
bobcatfish: @afrittoli the recent catalog PRs starting with #389 and then #401 introduced `tekton.dev/tags` and i think `tekton.dev/displayName`; we still need to formally accept these as part of our overall catalog plans but I didn't want to block those giant PRs on this and I think they probably make sense in the long run

`tekton.dev/pipelines.minVersion` is included in https://github.com/tektoncd/pipeline/issues/2588

but TL;DR you're totally right that we need to explicitly propose these tags, make it obvious to task authors they need to use them and also add automation to verify them! ( #413 )

boskos-acquire/0.1/boskos-acquire.yaml
```diff
@@ -0,0 +1,67 @@
...
+      spec:
+        containers:
+        - name: heatbeat
+          image: gcr.io/k8s-staging-boskos/boskosctl@sha256:a7fc984732c5dd0b4e0fe0a92e2730fa4b6bddecd0f6f6c7c6b5501abe4ab105
+          args:
+          - heartbeat
+          - --server-url=$(params.server-url)
+          - --owner-name=$(params.owner-name)
+          - --resource=$FULL_RESOURCE_OUTPUT
+          - --period=10s
```
bobcatfish, line 60: note to self: i made this 10 seconds for testing but we probably dont need it to be that frequent, for example kubetest uses 5m (https://github.com/kubernetes/test-infra/blob/94ee2a70262d1c2f2c878cf4e986b4b41180f5f6/kubetest/main.go#L753-L759) and afaik the default timeout boskos uses is 30m

boskos-acquire/0.1/boskos-acquire.yaml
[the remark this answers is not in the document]
vdemeester: @bobcatfish I don't get what does it change ? writing the file vs piping it to `kubectl` is doing the same. Instead of writing `/workspace/heartbeat.yaml` and then running kubectl with it, we *pipe* what was written in that file directly to `kubectl`, nothing else changes :thinking:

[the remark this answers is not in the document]
bobcatfish: My train of thought was something like:
1. cool we can just use the kubectl image!
2. oh wait, what about those values from the files (not in andrea's example)
3. oh we'll need to run a shell
4. oh we're already running a shell
5. i guess the advantage here is that we're running 1 image instead of 2?

I think it's around (3): now we need a shell in the image that runs kubectl and (5) that i get a lot less excited.

it seems clearer to me to have 2 steps, one that makes the file and another that uses the file vs squishing them together for efficiency reasons. so im wondering if using smaller images for both might be another solution?

the kubectl image here already only contains sh and does not contain bash:

```
(⎈ |euca:default)➜  catalog git:(boskos) ✗ docker run -it --entrypoint /bin/bash lachlanevenson/k8s-kubectl
docker: Error response from daemon: OCI runtime create failed: container_linux.go:346: starting container process caused "exec: \"/bin/bash\": stat /bin/bash: no such file or directory": unknown.
(⎈ |euca:default)➜  catalog git:(boskos) ✗ docker run -it --entrypoint /bin/sh lachlanevenson/k8s-kubectl
~ #
```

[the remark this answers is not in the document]
vdemeester: right, my question is why do we even need to writet the file (`heartbeat.yaml`). On bash/sh, I don't think there is any "bash" specifics in the script used, it would work with `/bin/sh` just as well, so using `lachlanevenson/k8s-kubectl` works just fine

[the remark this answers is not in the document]
bobcatfish: note that @a-roberts  just added a kubectl task (https://github.com/tektoncd/catalog/pull/394) this is getting me thinking about sharing steps again (https://github.com/tektoncd/pipeline/issues/1260) 🤔 

> right, my question is why do we even need to writet the file (heartbeat.yaml)

we dont need to write it, i think the trade off is something like:

1. is it clearer when it's 2 steps (maybe a little? maybe not? personally i think 2 steps is clearer than a script)
2. does the image contain everything you need to make it 1 step (it contains /bin/sh but not bash, so we wouldnt have the full power of bash replacement)

I think it's a bit clearer with 2 steps, so if i can switch out ubuntu for a faster image, does that address your concerns, or do you feel like the overhead of writing the file before applying it isn't worth it?

(oh oh oh - if only you could pipe the stdout of one step into another 😍😍😍  https://github.com/tektoncd/pipeline/issues/2925 ??? what do you think? there'd be more file i/o for sure, but wouldn't it be so clear ???)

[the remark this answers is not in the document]
afrittoli: I would definitely use a smaller image (busybox?).
 
Personally I would just collapse the two steps, not only because it is slightly more efficient, but also because it would seem more readable to me: 1 step gets the lease from boskos, 1 step starts the heartbeat. 
If you prefer having the two separate steps I don't mind.

How would you implement piping stdout from one step into another? Would the two steps then run in parallel?  If not I feel it would be misleading to talk about "pipe".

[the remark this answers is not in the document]
bobcatfish: > Would the two steps then run in parallel? If not I feel it would be misleading to talk about "pipe".

whoa TIL that all the processes connected with pipes are started together (https://en.wikipedia.org/wiki/Pipeline_(Unix)#Implementation) - i had no idea! i thought `process1 | process2 | process3` would run process1, then 2, then 3, i didnt realize theyd all start together

TBH that's even more similar to the way our steps work, since they do actually all start simultaneously. to do this in a truly unix pipe-like manner, we could implement "inter process pipes" if we wanted to

but what im suggesting in #2925 is much simpler, not actually streaming data between steps (maybe day? probably crazy) but just providing stdout to the next step once the previous step is completed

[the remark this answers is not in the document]
bobcatfish: > Personally I would just collapse the two steps, not only because it is slightly more efficient, but also because it would seem more readable to me: 1 step gets the lease from boskos, 1 step starts the heartbeat.
If you prefer having the two separate steps I don't mind.

we're probably just getting into personal preference now but i'll say my final 2 cents - having 2 steps makes sense to me b/c:

1. The step that creates the config needs a shell, so use an image with a shell
2. The step that applies the config doesn't need a shell, it just needs kubectl, so don't use a shell, just use kubectl

boskos-acquire/0.1/README.md
bobcatfish: ill use that example instead!

boskos-acquire/0.1/boskos-acquire.yaml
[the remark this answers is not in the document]
bobcatfish: too bad the kubectl image is 50mb X'D (maybe they should remove sh XD)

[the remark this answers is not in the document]
bobcatfish: i bet kubectl needs sh to run

bobcatfish:
PTAL @vdemeester @afrittoli - i'm still using 2 tasks for writing the file + applying it, but I've changed from ubuntu to bash :O

boskos-release/0.1/samples/pipelinerun.yaml
```diff
@@ -0,0 +1 @@
+boskos-acquire/0.1/examples/pipelinerun.yaml
```
PuneetPunamiya, line 1: Can you please change it to `boskos-acquire/0.1/samples/pipelinerun.yaml`

boskos-release/0.1/samples/service-account.yaml
```diff
@@ -0,0 +1 @@
+boskos-acquire/0.1/examples/service-account.yaml
```
PuneetPunamiya, line 1: Can you please change it to `boskos-acquire/0.1/samples/pipelinerun.yaml`

boskos-acquire/0.1/boskos-acquire.yaml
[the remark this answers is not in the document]
vdemeester: > > Personally I would just collapse the two steps, not only because it is slightly more efficient, but also because it would seem more readable to me: 1 step gets the lease from boskos, 1 step starts the heartbeat.
> > If you prefer having the two separate steps I don't mind.
> 
> we're probably just getting into personal preference now but i'll say my final 2 cents - having 2 steps makes sense to me b/c:
> 
>     1. The step that creates the config needs a shell, so use an image with a shell
>     2. The step that applies the config doesn't need a shell, it just needs kubectl, so don't use a shell, just use kubectl

I am not actually sure it's too much about personal preferences, but also about user experiences. More step means more containers (if you have things like LimitRange, …, this means more resource reserved), more different images means more images to pull, more image to store on nodes :upside_down_face:.

I think our point (andrea and I) is that you don't need to "create" the config (as a file), you can just pipe it to kubectl. And as you said, the `kubectl` image already have `sh`, so it's all you need.

Any way, not wanting to block the PR for that :wink:

boskos-release/0.1/samples/service-account.yaml
bobcatfish: Thanks for catching these @PuneetPunamiya !

boskos-acquire/0.1/boskos-acquire.yaml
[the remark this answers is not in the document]
bobcatfish: > but also about user experiences. More step means more containers (if you have things like LimitRange, …, this means more resource reserved), more different images means more images to pull, more image to store on nodes 

I guess, feels like a strange place to be optimizing to me: I also have a dream of having many tiny images which do one thing well and only that thing, which can be cleanly composed together.

anyway maybe we can revisit it as we go - seems like it's not too bad to have the 2 images instead of one for now, so seems like it's okay to leave it for now

bobcatfish:
PTAL @vdemeester @afrittoli @PuneetPunamiya

boskos-acquire/0.1/samples/pipelinerun.yaml
```diff
@@ -0,0 +1,34 @@
...
+        taskRef:
+          name: boskos-acquire
+        params:
+          - name: owner-name
+            value: $(params.owner-name)
+      - name: wait
+        runAfter: [boskos-acquire]
+        taskSpec:
+          steps:
+          - image: ubuntu
```
popcor255, line 22:
```suggestion
          - image: alpine
```
Alpine image size is 5.57mb and Ubuntu is like 74.9mb. They both have /bin/sh and sleep.

boskos-acquire/0.1/boskos-acquire.yaml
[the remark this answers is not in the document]
popcor255: I tried making my own image for this. However, the only way I was able to getting to work was by making a curl request to the k8s cluster directly. If the k8s api changes than it would have to be updated every time. I tried finding a k8s-cli slim but my google-foo failed me :sob:. It would be cool to have a container that had a bunch of kubectl slim that just basic verb/object capabilities. I might even do it as a side-project, tbh.

boskos-acquire/0.1/boskos-acquire.yaml
```diff
@@ -0,0 +1,67 @@
+apiVersion: tekton.dev/v1beta1
+kind: Task
+metadata:
+  name: boskos-acquire
+  labels:
+    app.kubernetes.io/version: "0.1"
+  annotations:
+    tekton.dev/pipelines.minVersion: "0.12.1"
+    tekton.dev/tags: "boskos,test"
+spec:
```
PuneetPunamiya, line 10: It would be great if you could please add the description in the form of summary and description 

For e.g.

```yaml
spec:
     description: >-
       One line summary of the task.         👈 # Summary

       Description of the task 
  ```

File: boskos-release/0.1/boskos-release.yaml [hunk 9]
```diff
@@ -0,0 +1,39 @@
...
+kind: Task
+metadata:
+  name: boskos-release
+  labels:
+    app.kubernetes.io/version: "0.1"
+  annotations:
+    tekton.dev/pipelines.minVersion: "0.12.1"
+    tekton.dev/tags: "boskos,test"
+spec:
+  description: |
```
PuneetPunamiya, line 11: It would be great if you could please add the description in the form of summary and description 

For e.g.

```yaml
spec:
     description: >-
       One line summary of the task.         👈 # Summary

       Description of the task 
  ```

**Comment by PuneetPunamiya** [comment]
@bobcatfish  The tasks are not in the task directory, can you please add the task in the task directory and I think you need to rebase the code as well

File: boskos-acquire/0.1/boskos-acquire.yaml [hunk 10]
```diff
@@ -0,0 +1,67 @@
+apiVersion: tekton.dev/v1beta1
```
sthaha, line 1: @bobcatfish I am aware this was submitted before the TEP - ReOrg but since you are adding the annotations and labels, could you also move the task  under the `task` directory so that it shows up in our hub 👼

boskos-acquire/0.1/boskos-acquire.yaml
```diff
@@ -0,0 +1,67 @@
...
+    script: |
+      RESOURCE=$(boskosctl acquire \
+        --server-url=$(params.server-url) \
+        --owner-name=$(params.owner-name) \
+        --type=$(params.type) \
+        --state=free \
+        --target-state=busy)
+      echo $RESOURCE > /workspace/full-resource-output.json
+      echo $RESOURCE | jq -rj ".name" > /tekton/results/leased-resource
+  - name: create-heartbeat-pod-yaml
```
vdemeester, line 41: I am gonna try one last time :stuck_out_tongue_closed_eyes: . I don't see why the step below is worse than the 2 steps proposed :upside_down_face: 

```yaml
  - name: create-heartbeat-pod-yaml
    image: lachlanevenson/k8s-kubectl
    script: |
      FULL_RESOURCE_OUTPUT=$(cat /workspace/full-resource-output.json)
      LEASED_RESOURCE=$(cat /tekton/results/leased-resource)
      cat <<EOF | kubectl apply -f -
      apiVersion: v1
      kind: Pod
      metadata:
        name: boskos-heartbeat-$LEASED_RESOURCE
      spec:
        containers:
        - name: heatbeat
          image: gcr.io/k8s-staging-boskos/boskosctl@sha256:a7fc984732c5dd0b4e0fe0a92e2730fa4b6bddecd0f6f6c7c6b5501abe4ab105
          args:
          - heartbeat
          - --server-url=$(params.server-url)
          - --owner-name=$(params.owner-name)
          - --resource=$FULL_RESOURCE_OUTPUT
          - --period=5m
      EOF
```

You don't need to write that yaml to `/workspace/heartbeat.yaml` ever, it's not gonna be used :upside_down_face:

vdemeester: btw, because we are using `:latest` tag on the steps, k8s *will* pull the image each time, so 2 steps is gonna cost more than one :upside_down_face: (if we do use digest or tags for images, that's another story)

bobcatfish: Okay, seems like you feel strongly enough that I'll combine the two steps into one. I've created https://github.com/tektoncd/catalog/issues/441 to continue the discussion

> btw, because we are using :latest tag on the steps, k8s will pull the image each time, so 2 steps is gonna cost more than one 🙃 (if we do use digest or tags for images, that's another story)

Oh good point, ill add the digest (I think this is a best practice we should recommend for all catalog tasks?)

(also it depends on the image pull policy right?)

boskos-acquire/0.1/boskos-acquire.yaml
bobcatfish: will do, thanks @sthaha !

bobcatfish:
> here is no way to have the Pod/heartbeat as a task? does the pod need to be keep hearbeating for a while and due cannot be 'plugged' directly in a step?

Exactly - it needs to run the entire time the project is in use. Otherwise, after some period of time (I think it's 30 min by default) boskos will assume the project isn't being used and will clean it and put it back into its pool of available projects. I've also created https://github.com/tektoncd/pipeline/issues/2973 as a possible alternative approach.

bobcatfish:
Should be ready for another look @vdemeester @afrittoli @PuneetPunamiya @popcor255 @sthaha @chmouel  !

boskos-acquire/0.1/boskos-acquire.yaml
vdemeester: > (also it depends on the image pull policy right?)

So, I think, whatever the pull policy is, using `:latest` force kubernetes to pull the image each time (because it has *no* way to know which "latest" you wanted :stuck_out_tongue_closed_eyes: )

chmouel:
/lgtm

Looking Good! 🤙🏽

boskos-acquire/0.1/boskos-acquire.yaml
bobcatfish: Are you sure?

https://kubernetes.io/docs/concepts/configuration/overview/#container-images
https://kubernetes.io/docs/concepts/containers/images/#updating-images

It sounds like that might only be in the case of having on imagePullPolicy?

> imagePullPolicy is omitted and either the image tag is :latest or it is omitted: Always is applied.

vdemeester: Right, with `IfNotPresent` or `Never` it won't try to pull it (if not present for the first one). That said, in the task itself, there is no `imagePullPolicy` so we can "only" assume the default behavior. Overall, the more the catalog grow, the more we should try to follow our recommendation https://github.com/tektoncd/catalog/blob/master/recommendations.md#reference-images-by-digest :upside_down_face: @sthaha validation tool could help there (to warn of tasks that uses latest)

PR #410, opened 2020-07-11
Add Server's User to the mvn

soharaki:
# Changes

Add `servers` tag for the private repository server.

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/helm/helm-upgrade-from-source.yaml
+++ b/helm/helm-upgrade-from-source.yaml
@@ -2,7 +2,16 @@ apiVersion: tekton.dev/v1beta1
 kind: Task
 metadata:
   name: helm-upgrade-from-source
+  labels:
+    app.kubernetes.io/version: "0.1"
+  annotations:
+    tekton.dev/pipelines.minVersion: "0.12.1"
+    tekton.dev/tags: helm
 spec:
+  description: >-
+    These tasks will install / upgrade a helm chart into your Kubernetes /
+    OpenShift Cluster using Helm
+
   params:
     - name: charts_dir
       description: The directory in source that contains the helm chart


task/maven/0.1/maven.yaml
```diff
@@ -24,6 +24,14 @@ spec:
...
       type: string
       default: ""
+    - name: SERVER_USER
+      description: The username for the server
+      type: string
+      default: ""
+    - name: SERVER_PASSWORD
+      description: The password for the server
+      type: string
+      default: ""
```
vinamra28, line 34: Hi @soharaki, can we avoid adding more parameters to the task and instead make use of `ConfigMap` if we need to mount more settings into settings.xml? [Reference](https://github.com/tektoncd/catalog/tree/master/task/maven/0.1#with-custom-m2settingsyaml)

soharaki: Hi @vinamra28 
Here are some of the ways we want to use mirror servers in maven.
1. Communicate with maven by https.
2. Use the cached library in a private network environment.
In the latter example, I supposed that `server` tag is required, along with the mirror tag for the security.

I understand what you are saying.
If we try to represent all the tags in settings.xml in maven's Task parameter, and the we have to add so MANY parameters to the settings.xml.

```xml
    <settings xmlns="http://maven.apache.org/SETTINGS/1.0.0"
      xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
      xsi:schemaLocation="http://maven.apache.org/SETTINGS/1.0.0
                          https://maven.apache.org/xsd/settings-1.0.0.xsd">
      <localRepository/>
      <interactiveMode/>
      <offline/>
      <pluginGroups/>
      <servers/>
      <mirrors/>
      <proxies/>
      <profiles/>
      <activeProfiles/>
    </settings>
```
via [settings.xml](https://maven.apache.org/settings.html)

soharaki:
> Could we just update the parameters in the README as well otherwise looks good to me 🙂

Thanks for the review. I forgot to update the README.md. I'll do it.

task/maven/0.1/README.md
```diff
@@ -12,6 +12,8 @@ kubectl apply -f https://raw.githubusercontent.com/tektoncd/catalog/master/task/
 
 - **GOALS**: Maven `goals` to be executed
 - **MAVEN_MIRROR_URL**: Maven mirror url (to be inserted into ~/.m2/settings.xml)   
+- **SERVER_USER**: Username to to authenticate to the server (to be inserted into ~/.m2/settings.xml)
```
vinamra28, line 15:
```suggestion
- **SERVER_USER**: Username to authenticate to the server (to be inserted into ~/.m2/settings.xml)
```

vinamra28, line 15:
```suggestion
- **SERVER_USER**: Username to authenticate to the server (to be inserted into ~/.m2/settings.xml)
```

soharaki: @vinamra28 
Embarrassing. I fixed the typos.

sthaha:
@vdemeester I agree with you. We don't have to block this PR from progressing. 

On to the versioning part of the discussion
> So, if we follow the TEP-003 closely, this might be worth a 0.2 version

On the version 0.2, I kind of agree it simplifies things and strictly follows `semver`; however since the parameters are new and have default values, the old taskruns should still run the same - right (i.e. no behavioural change to existing usage of tasks and the tests in 0.1 version should still pass against the new version). Given this we may decide to allow this to be `0.1` but it is better to document it in the TEP. 

Having said all this .. I prefer a `0.2` to keep things simple.

vinamra28:
hi @soharaki, since #449 is now merged so you can continue with this PR and add the following changes in `0.2` as suggested above :slightly_smiling_face:

soharaki:
This is a note for myself.

Task:
1. Use `git reset --hard` to undo the fix for https://github.com/soharaki/catalog.
2. Add upstream with `git remote add upstream https://github.com/tektoncd/catalog.git`. 
3.  Create master branch.

```sh
    git fetch upstream
    git checkout -b master
    git merge upstream/master
```   
4. Fix!
5.  `git checkout patch-1`
6.  `git merge master`
7.  `git push`

soharaki:
> hi @soharaki, since #449 is now merged so you can continue with this PR and add the following changes in `0.2` as suggested above 🙂

Excuse me, I cleaned up the cloned respository's history and the PR is closed, so I recreated it as #504, the version for task v2.

PR #433, opened 2020-07-22
Remove implicit "set -x" from git-clone, deleteExisting by default

sbwsg:
# Changes

This commit adds git-clone v0.2. One default behaviour has changed
and one new param has been introduced in this version:

1) Tasks that don't start their script with #! automatically receive
a preamble that includes `set -x`. This results in verbose output
because every command is printed while the Task is running.

This commit adds a new param, "verbose", that - when set to "false" -
prevents the commands run by the task from being printed to logs.
The default value is "true": the commands run by the task will be
printed in the log.

2) The git-clone Task offers a param, deleteExisting, that cleans
any existing directory before performing a clone into it. This
has previously been set to "false" by default since it's potentially
destructive. But it sounds like there's quite often a request for
how to enable this behaviour, so it seems like the more pragmatic
default would be "true".

This commit sets deleteExisting to be defaulted to "true" so that
any existing git checkout is removed prior to a new clone being
performed.

# Submitter Checklist

review them:

  - [x] mandatory `spec.description` follows convention

### DIFF
--- a/task/git-clone/0.2/README.md
+++ b/task/git-clone/0.2/README.md
@@ -1,0 +1,60 @@
+# Git Task
+
+This `Task` is Git task to work with repositories used by other tasks
+in your Pipeline.
+## `git-clone`
+
+**Please Note: this Task is only compatible with Tekton Pipelines versions 0.11-rc1 and greater!**
+
+This `Task` has two required inputs:
+
+1. The URL of a git repo to clone provided with the `url` param.
+2. A Workspace called `output`.
+
+The `git-clone` `Task` will clone a repo from the provided `url` into the
+`output` Workspace. By default the repo will be cloned into the root of
+your Workspace. You can clone into a subdirectory by setting this `Task`'s
+`subdirectory` param. If the directory where the repo will be cloned is
+already populated then by default the contents will be deleted before the
+clone takes place. This behaviour can be disabled by setting the
+`deleteExisting` param to `"false"`.
+
+This `Task` does the job of the legacy `GitResource` `PipelineResource` and
+is intended as its replacement. This is part of our plan to [offer replacement
+`Tasks` for Pipeline Resources](https://github.com/tektoncd/catalog/issues/95)
+as well as
+[document those replacements](https://github.com/tektoncd/pipeline/issues/1369).
+
+### Workspaces
+
+* **output**: A workspace for this Task to fetch the git repository in to.
+
+### Parameters
+
+* **url**: git url to clone (_required_)
+* **revision**: git revision to checkout (branch, tag, sha, ref…) (_default_: master)
+* **refspec**: git refspec to fetch before checking out revision (_default_:refs/heads/master:refs/heads/master)
+* **submodules**: defines if the resource should initialize and fetch the submodules (_default_: true)
+* **depth**: performs a shallow clone where only the most recent commit(s) will be fetched (_default_: 1)
+* **sslVerify**: defines if http.sslVerify should be set to true or false in the global git config (_default_: true)
+* **subdirectory**: subdirectory inside the "output" workspace to clone the git repo into (_default:_ "")
+* **deleteExisting**: clean out the contents of the repo's destination directory if it already exists before cloning the repo there (_default_: true)
+* **httpProxy**: git HTTP proxy server for non-SSL requests (_default_: "")
+* **httpsProxy**: git HTTPS proxy server for SSL requests (_default_: "")
+* **noProxy**: git no proxy - opt out of proxying HTTP/HTTPS requests (_default_: "")
+* **verbose**: log the commands that are executed during `git-clone`'s operation (_default_: true)
+
+### Results
+
+* **commit**: The precise commit SHA that was fetched by this Task
+
+## Usage
+
+### `git-clone`
+
+The following pipelines demonstrate usage of the git-clone Task:
+
+- [Cloning a branch](../0.1/samples/git-clone-checking-out-a-branch.yaml)
+- [Checking out a specific git commit](../0.1/samples/git-clone-checking-out-a-commit.yaml)
+- [Checking out a git tag and using the "commit" Task Result](../0.1/samples/using-git-clone-result.yaml)
+
--- a/task/git-clone/0.2/git-clone.yaml
+++ b/task/git-clone/0.2/git-clone.yaml
@@ -1,0 +1,125 @@
+apiVersion: tekton.dev/v1beta1
+kind: Task
+metadata:
+  name: git-clone
+  labels:
+    app.kubernetes.io/version: "0.2"
+  annotations:
+    tekton.dev/pipelines.minVersion: "0.12.1"
+    tekton.dev/tags: git
+    tekton.dev/displayName: "git clone"
+spec:
+  description: >-
+    These Tasks are Git tasks to work with repositories used by other tasks
+    in your Pipeline.
+
+    The git-clone Task will clone a repo from the provided url into the
+    output Workspace. By default the repo will be cloned into the root of
+    your Workspace. You can clone into a subdirectory by setting this Task's
+    subdirectory param.
+
+  workspaces:
+    - name: output
+      description: The git repo will be cloned onto the volume backing this workspace
+  params:
+    - name: url
+      description: git url to clone
+      type: string
+    - name: revision
+      description: git revision to checkout (branch, tag, sha, ref…)
+      type: string
+      default: master
+    - name: refspec
+      description: (optional) git refspec to fetch before checking out revision
+      default: ""
+    - name: submodules
+      description: defines if the resource should initialize and fetch the submodules
+      type: string
+      default: "true"
+    - name: depth
+      description: performs a shallow clone where only the most recent commit(s) will be fetched
+      type: string
+      default: "1"
+    - name: sslVerify
+      description: defines if http.sslVerify should be set to true or false in the global git config
+      type: string
+      default: "true"
+    - name: subdirectory
+      description: subdirectory inside the "output" workspace to clone the git repo into
+      type: string
+      default: ""
+    - name: deleteExisting
+      description: clean out the contents of the repo's destination directory (if it already exists) before trying to clone the repo there
+      type: string
+      default: "true"
+    - name: httpProxy
+      description: git HTTP proxy server for non-SSL requests
+      type: string
+      default: ""
+    - name: httpsProxy
+      description: git HTTPS proxy server for SSL requests
+      type: string
+      default: ""
+    - name: noProxy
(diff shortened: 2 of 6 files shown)


sbwsg:
/hold

I will add some tests around `deleteExisting` "true" and "false"

bobcatfish:
> Tasks that don't start their script with #! automatically receive a preamble that includes set -x. This results in verbose output because every command is printed while the Task is running.

@sbwsg could you explain more about why we want to remove this? i find being able to have this kind of verbose output really helps with debugging, and one of the difficulties I have with a lot of our e2e bash scripts is that they do not do this and it's very hard to know what they actually executing

sbwsg:
> @sbwsg could you explain more about why we want to remove this?

This stems from a conversation on Slack with @chmouel who suggested reducing the amount of log noise to make things a bit less confusing.  I guess when the script is working users don't really want or need to see the noise. But when the script isn't working it can be helpful to get lengthier log output.

I could add a param like `verbose` which defaults to "false".  Making it "true" could result in `set -x` being added.  WDYT @bobcatfish @chmouel ?

task/git-clone/0.2/git-clone.yaml
```diff
@@ -0,0 +1,118 @@
...
+      default: ""
+  results:
+    - name: commit
+      description: The precise commit SHA that was fetched by this Task
+  steps:
+    - name: clone
+      image: gcr.io/tekton-releases/github.com/tektoncd/pipeline/cmd/git-init:v0.14.2
+      script: |
+        #!/bin/sh
+        set -e
```
sthaha, line 75: as a rule of thumb, I think we should have `set -eu` set by default if not more stricter `set -eu -o pipefail` for all bash scripts.

task/git-clone/0.2/git-clone.yaml
```diff
@@ -0,0 +1,118 @@
...
+          -revision "$(params.revision)" \
+          -refspec "$(params.refspec)" \
+          -path "$CHECKOUT_DIR" \
+          -sslVerify="$(params.sslVerify)" \
+          -submodules="$(params.submodules)" \
+          -depth "$(params.depth)"
+        cd "$CHECKOUT_DIR"
+        RESULT_SHA="$(git rev-parse HEAD | tr -d '\n')"
+        EXIT_CODE="$?"
+        if [ "$EXIT_CODE" != 0 ]
```
sthaha, line 113: it would be nice if we followed the same convention as above 
```
if [[ "$EXIT_CODE" != 0 ]] ; then 
   exit $EXIT_CODE
fi
```

task/git-clone/0.2/git-clone.yaml
```diff
@@ -0,0 +1,118 @@
...
+        /ko-app/git-init \
+          -url "$(params.url)" \
+          -revision "$(params.revision)" \
+          -refspec "$(params.refspec)" \
+          -path "$CHECKOUT_DIR" \
+          -sslVerify="$(params.sslVerify)" \
+          -submodules="$(params.submodules)" \
+          -depth "$(params.depth)"
+        cd "$CHECKOUT_DIR"
+        RESULT_SHA="$(git rev-parse HEAD | tr -d '\n')"
```
sthaha, line 111: isn't `| tr -d '\n'` unnecessary? 🤔

chmouel: @sthaha I think you meant :
```suggestion
        RESULT_SHA="$(echo -n git rev-parse HEAD)"
```

which should do the same as @sbwsg in a slightly more elegant way..

chmouel:
> I could add a param like `verbose` which defaults to "false". Making it "true" could result in `set -x` being added. WDYT @bobcatfish @chmouel ?

I just think that having verbosity by default was confusing, 

this maybe could be a pipeline level sort of option when we have script with no shbang, let the user in some way specify she or he may wants to run it in a verbose way?

+1 on scott's suggestions in the meantime.

task/git-clone/0.2/git-clone.yaml
sthaha: @chmouel no I mean `RESULT_SHA=$(git rev-parse HEAD)` 

The echo was to just show that it doesn't insert a `\n`

bobcatfish:
> I just think that having verbosity by default was confusing,

It's a tough balance to strike: when you don't need the output, it's confusing, however when you do need it (and you often don't know that until it's too late), it's really handy to have it.

The e2e test scripts we have been using across Tekton so far have taken the stance that less is better and have not included -x. I've found this makes debugging them, maintaining them, and updating them extremely hard.

If we want to add a flag to disable this, I suggest we default the functionality to being more verbose by default.

Note that script mode defaults to -x as well https://github.com/tektoncd/pipeline/blob/master/docs/tasks.md#running-scripts-within-steps

task/git-clone/0.2/git-clone.yaml
chmouel: yeah you right either way it's not  needed  to cut the `\n`....

sbwsg:
OK, I've updated with - I think - all suggested modifications.  I've also introduced a verbose param to make the `set -x` behaviour configurable.

/hold cancel

/test pull-tekton-catalog-integration-tests

task/git-clone/0.2/samples/git-clone-checking-out-a-branch.yaml
```diff
@@ -0,0 +1,86 @@
...
+    affinity:
+      podAntiAffinity:
+        requiredDuringSchedulingIgnoredDuringExecution:
+        - labelSelector:
+            matchExpressions:
+            - key: "tekton.dev/pipelineRun"
+              operator: In
+              values:
+              - git-clone-checking-out-a-branch
+          topologyKey: kubernetes.io/hostname
```
chmouel, line 70: I am not sure I understand this part, is it to make sure the pipelineRuns doesn't end up on the same minion ?

vdemeester: I understand the same as @chmouel, is it to make sure we don't run on the same minion of that task because they might "hit on each others with PVC or sthg" ?

sbwsg: this was copied verbatim from the 0.1 version.  Looks like it might be a way to manage the pvcs pre-affinityassistant like you say.  I'll go back and take a better look.

sbwsg: I've removed the affinity, let's see how this runs in integration.

sbwsg:
OK looks like that worked out ok without the affinity rules.  This should be ready for re-review.

sbwsg:
Hey this is just waiting on an approve.  Do any reviewers have more comments before it proceeds?

task/git-clone/0.2/README.md
```diff
@@ -0,0 +1,56 @@
...
+* **revision**: git revision to checkout (branch, tag, sha, ref…) (_default:_ master)
+* **refspec**: git refspec to fetch before checking out revision (_default_:refs/heads/master:refs/heads/master)
+* **submodules**: defines if the resource should initialize and fetch the submodules (_default_: true)
+* **depth**: performs a shallow clone where only the most recent commit(s) will be fetched (_default_: 1)
+* **sslVerify**: defines if http.sslVerify should be set to true or false in the global git config (_default_: true)
+* **subdirectory**: subdirectory inside the "output" workspace to clone the git repo into (_default:_ "")
+* **deleteExisting**: clean out the contents of the repo's destination directory if it already exists before cloning the repo there (_default_: true)
+* **httpProxy**: git HTTP proxy server for non-SSL requests
+* **httpsProxy**: git HTTPS proxy server for SSL requests
+* **noProxy**: git no proxy - opt out of proxying HTTP/HTTPS requests
```
bobcatfish, line 41: are these optional? i think it would make sense to mention a default for them if so? i also think it might be useful to list the required params before the optional ones

bobcatfish: i think "verbose" is missing from this list

(but also i think maintaining this list AND the docs in the Task itself is strange, but that's a larger question!)

task/git-clone/0.2/git-clone.yaml
```diff
@@ -0,0 +1,125 @@
...
+      type: string
+      default: ""
+    - name: noProxy
+      description: git no proxy - opt out of proxying HTTP/HTTPS requests
+      type: string
+      default: ""
+    - name: verbose
+      description: log the commands used during execution
+      type: string
+      default: "true"
```
bobcatfish, line 70: maybe we can add bool types some day

task/git-clone/0.2/README.md
```diff
@@ -0,0 +1,56 @@
...
+### Parameters
+
+* **url**: git url to clone (_required_)
+* **revision**: git revision to checkout (branch, tag, sha, ref…) (_default:_ master)
+* **refspec**: git refspec to fetch before checking out revision (_default_:refs/heads/master:refs/heads/master)
+* **submodules**: defines if the resource should initialize and fetch the submodules (_default_: true)
+* **depth**: performs a shallow clone where only the most recent commit(s) will be fetched (_default_: 1)
+* **sslVerify**: defines if http.sslVerify should be set to true or false in the global git config (_default_: true)
+* **subdirectory**: subdirectory inside the "output" workspace to clone the git repo into (_default:_ "")
+* **deleteExisting**: clean out the contents of the repo's destination directory if it already exists before cloning the repo there (_default_: true)
```
bobcatfish, line 38: i think it might make sense to highlight this one more - i wouldnt expect this behavior personally and now this is the default

sbwsg:
ok if integration fails I'll make the suggested changes in this PR otherwise I'll open a new one.

bobcatfish:
```
{"level":"fatal","ts":1597417297.477715,"logger":"fallback-logger","caller":"imagedigestexporter/main.go:78","msg":"Unexpected error writing message .//image-digested to open .//image-digested: permission denied","commit":"6d12d1e","stacktrace":"main.main\n\tgithub.com/tektoncd/pipeline/cmd/imagedigestexporter/main.go:78\nruntime.main\n\truntime/proc.go:203"}
2020/08/14 15:01:37 Skipping step because a previous step failed
2020/08/14 15:01:40 Exiting...
```

Did we just upgrade the version of tekton we're testing with and hit a bug maybe?

task/git-clone/0.2/README.md
sbwsg: Added an additional bit of prose to the README calling out this default behaviour.

task/git-clone/0.2/README.md
sbwsg: Added "verbose".  Added the default values for the proxy params.  Sort order has required params at top.

Issue #413, opened 2020-07-13
Validation Tool to follow the Catalog Structure


PuneetPunamiya:
# Expected Behavior

1. Each resource should follow the following structure
```
    ./task/         👈 the kind of the resource 
      
      /argocd       👈 definition file must have same name
        /0.1
          /OWNERS     👈 owners of this ressource
          /README.md     
          /argocd.yaml   👈 the file name must be resource name
          /samples/deploy-to-k8s.yaml
        /0.2/...
     
     /golang-build
        /0.1
          /README.md
          /golang-build.yaml
          /samples/golang-build.yaml
```

* **NOTE** : The `resource name` and the `directory name` should match

2. Resource YAML file should include the following changes
  *  Labels include the version of the resource.
  *  Annotations include
     - `minimum pipeline version` supported by the resource 
     - `tags` associated with the resource
     - `displayName` of the resource
    
  ```yaml
    
 labels:
    app.kubernetes.io/version: "0.1"            👈 version of the resource
    
  annotations:
    tekton.dev/pipelines.minVersion: "0.12.1"   👈 will work from x version  of pipeline
    tekton.dev/tags: "ansible, cli"                        👈 Comma separated list of tags 
    tekton.dev/displayName: "Ansible Tower Cli"  👈 displayName can be optional   
    
spec:
  description: |-
    ansible-tower-cli task simplifies 
    workflow, jobs, manage users...   👈 # Summary
    
    Ansible Tower (formerly ‘AWX’) is a ...
  
  ```

# Actual Behavior

  - API breakage 
  - No concept of versions
  - Are the Resources valid 
  - No field to mention Minimum Pipeline Version

wumaxd:
@PuneetPunamiya Regarding 1.
How about creating a list of changed directories (`./task/<dir>`) based on the `CHANGED_FILES` in the `presubmit-tests.sh` (https://github.com/tektoncd/catalog/blob/a0e2d1eb4d8fcdd6c0ff0e9389be8dfbf6124695/vendor/github.com/tektoncd/plumbing/scripts/presubmit-tests.sh#L79).
Based on this list check if for each version a task exists (`./task/<dir>/<version>/<dir>.yml`) and if a README exists (`./task/<dir>/<version>/README.md`). Would you want to see more enforced?

I would do this using a bash if/else like this
```bash
if [ ! -f "$FILE" ]; then
    echo "$FILE does not exist."
fi
```

If this is fine, I would be happy to take this task.

[...]

PR #436, opened 2020-07-22
Add a task for creating a GKE cluster for e2e tests 🧪

bobcatfish:
# Changes

As part of #373 I'm continuing to create Tasks that we can use to run
tests for items in the catalog. This Task creates a cluster within a GKE
project and is based on the way we create clusters for the existing end
to end tests we have in Tekton, which are themselves based on Knative,
when are THEMselves based on k8s. These tests all use a tool called
kubetest, which is responsible for invoking boskos (see #408) and
creating a GKE cluster
(https://github.com/kubernetes/test-infra/tree/master/kubetest).

This Task attempts to create a cluster in the same way, which based on the
output in the end to end logs seems to consist of creating the cluster
and setting up a firewall for it.

I'm not sure if this belongs in the catalog itself since it is specific
to our own end to end tests and isn't as generic as other catalog tests,
so if we want to move this into the test directory that seems fine, but
I also think it could be an okay addition to the catalog on its own (esp
if other k8s based projects that test against GKE want to use it.)

# Submitter Checklist

review them:

          ```

          spec:
            description: >-

          ```

for more details._

---

### DIFF
--- a/task/gke-cluster-create/0.1/README.md
+++ b/task/gke-cluster-create/0.1/README.md
@@ -1,0 +1,43 @@
+# GKE End to End Cluster Create
+
+The Task `gke-e2e-cluster-create` can be used to create a GKE cluster in a GCP
+project and fetch a kubeconfig that can be used (in a context with both kubectl and gcloud
+available) to make requests to the cluster.
+
+The cluster created will have a firewall applied such that the only traffic allowed to instances
+in the cluster will be SSH, TCP services running on 80 or 8080, and services exposed via the
+NodePort default range (https://kubernetes.io/docs/concepts/services-networking/service/#nodeport).
+
+## Parameters
+
+* **project-name**: The name of the GCP project in which to create the GKE cluster. (_required_)
+* **private-key-path**: The path to the private key within the gcp-service-account workspace. (_required_)
+* **identifier**: A string which identifies the purpose for which this cluster is being created. Used to name other resources created. (_required_)
+* **min-nodes**: The minimum number of nodes in the cluster. (_default_:1)
+* **max-nodes**: The maximum number of nodes in the cluster. (_default_:3)
+* **region**: The region to create the cluster in. (_default_:us-central1)
+* **machine-type**: The machine type to create, from
+  https://cloud.google.com/compute/docs/machine-types. (_default_:n1-standard-4)
+* **image-type**: The type of image to create the nodes, from
+  https://cloud.google.com/kubernetes-engine/docs/concepts/node-images. (_default_:cos)
+* **cluster-version**: The GKE version to install, in a format that can be used as the
+    `--cluster-version` argument to https://cloud.google.com/sdk/gcloud/reference/beta/container/clusters/create
+    (_default_: latest)
+
+## Workspaces
+
+* **gcp-service-account**: A Secret or volume containing the private key of a GCP service account
+  that can create GKE clusters in the project
+* **kubeconfig**: A workspace into which a kubeconfig file called `kubeconfig` will be written that
+  will contain the information required to access the cluster. The `kubeconfig` will expect to use
+  gcloud to authenticate, so in order for it to be used it must be run in a container which contains
+  both kubectl and gcloud.
+
+## Results
+
+* **cluster-name** The name of the cluster that was created.
+
+## Usage
+
+See [samples/create-gke-cluster.yaml](samples/create-gke-cluster.yaml) for an example of a TaskRun
+that creates a GKE cluster and writes the kubeconfig to a PVC.
--- a/task/gke-cluster-create/0.1/gke-cluster-create.yaml
+++ b/task/gke-cluster-create/0.1/gke-cluster-create.yaml
@@ -1,0 +1,108 @@
+apiVersion: tekton.dev/v1beta1
+kind: Task
+metadata:
+  name: gke-cluster-create
+  labels:
+    app.kubernetes.io/version: "0.1"
+  annotations:
+    tekton.dev/pipelines.minVersion: "0.12.1"
+    tekton.dev/tags: "gke,test"
+    tekton.dev/displayName: "GKE Cluster Create"
+spec:
+  description: |
+    Create a GKE cluster.
+
+    This Task can be used to create a GKE cluster in a GCP project and fetch a kubeconfig that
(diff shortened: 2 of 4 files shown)


bobcatfish:
@ixdy this might be relevant to your interests! I'm trying to replicate more kubetest functionality in Tekton Tasks. Also if you happen to know of anything else that is done to the clusters kubetest creates that I've missed plz lemme know.

gke-e2e-cluster-create/0.1/README.md
```diff
@@ -0,0 +1,43 @@
+# GKE End to End Cluster Create
```
dlorenc, line 1: Any reason the name and tags all include "e2e" and "end to end"? This seems much more broadly useful than the end to end tests, even though that's what you're using it for.

gke-e2e-cluster-create/0.1/gke-e2e-cluster-create.yaml
```diff
@@ -0,0 +1,107 @@
...
+      UNIQUE_STR=$(head /dev/urandom | tr -dc a-z0-9 | head -c 10 ; echo '')
+      UNIQUE_NAME=$(params.identifier)-$UNIQUE_STR
+
+      # Configure gcloud to use the provided project and service account
+      gcloud auth activate-service-account --key-file=$(workspaces.gcp-service-account.path)/$(params.private-key-path)
+      gcloud config set project $(params.project-name)
+
+      # Create a network and a new cluster
+      gcloud compute networks create $UNIQUE_NAME --project $(params.project-name) --subnet-mode=auto
+      gcloud beta container clusters create \
```
dlorenc, line 69: Why do we need to use beta here?

gke-e2e-cluster-create/0.1/gke-e2e-cluster-create.yaml
```diff
@@ -0,0 +1,107 @@
...
+  steps:
+  - name: gcloud
+    image: google/cloud-sdk:slim@sha256:27b2c22bf259d9bc1a291e99c63791ba0c27a04d2db0a43241ba0f1f20f4067f
+    script: |
+      UNIQUE_STR=$(head /dev/urandom | tr -dc a-z0-9 | head -c 10 ; echo '')
+      UNIQUE_NAME=$(params.identifier)-$UNIQUE_STR
+
+      # Configure gcloud to use the provided project and service account
+      gcloud auth activate-service-account --key-file=$(workspaces.gcp-service-account.path)/$(params.private-key-path)
+      gcloud config set project $(params.project-name)
```
dlorenc, line 65: Is this call really required? you pass --project on every command below anyway.

gke-e2e-cluster-create/0.1/gke-e2e-cluster-create.yaml
```diff
@@ -0,0 +1,107 @@
...
+
+      # This firewall rule allows the cluster to expose SSH, TCP services running on 80 or 8080,
+      # and services exposed via the NodePort default range (https://kubernetes.io/docs/concepts/services-networking/service/#nodeport)
+      gcloud compute firewall-rules create e2e-ports-$UNIQUE_STR \
+        --project=$(params.project-name) \
+        --network=$UNIQUE_NAME \
+        --allow=tcp:22,tcp:80,tcp:8080,tcp:30000-32767,udp:30000-32767 \
+        --target-tags=$INSTANCE_TAG
+
+      printf $UNIQUE_NAME > /tekton/results/cluster-name
```
dlorenc, line 107: nit: newline

gke-e2e-cluster-create/0.1/samples/create-gke-cluster.yaml
```diff
@@ -0,0 +1,34 @@
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: kubeconfig-pvc
+spec:
+  resources:
+    requests:
+      storage: 5Gi
```
dlorenc, line 8: not sure it matters, but the kubeconfig probably won't be 5 gigs :)

gke-e2e-cluster-create/0.1/gke-e2e-cluster-create.yaml
dlorenc, line 107: Oooh one other one - the cluster name by itself isn't actually that useful - you generally need the region/zone to access it as well. Would that make sense as a result too? Or I guess we maybe don't need it since it's a param.

gke-e2e-cluster-create/0.1/README.md
bobcatfish: I wasn't sure what else to call it - i thought it would be doing more e2e specific stuff, but I guess the only thing is the firewall? Maybe that's a reasonable firewall to start a cluster off with? I'm happy to just call it create GKE cluster!

gke-e2e-cluster-create/0.1/gke-e2e-cluster-create.yaml
bobcatfish: i don't know! thats what kubetest was doing but ill remove it and if that still works then 👍

gke-e2e-cluster-create/0.1/gke-e2e-cluster-create.yaml
bobcatfish: oh yeah good point

plus this command always gives me this weird warning about the project not existing which is just not true :D

gke-e2e-cluster-create/0.1/samples/create-gke-cluster.yaml
bobcatfish: but _how do you know_

gke-e2e-cluster-create/0.1/gke-e2e-cluster-create.yaml
bobcatfish: whatever you think is best! it feels a bit weird to replicated something that is passed in as a param but if you think it's worth it, ill do it!

(maybe one day we'll have some kind of "cluster info" result type that will have all of this info?)

bobcatfish:
ps @dlorenc do you wanna be an OWNER of this Task? no pressure :D

bobcatfish:
Ready for another look @dlorenc !

PR #439, opened 2020-07-23
Fixes and improvements to the upload-pypi Task.

dlorenc:
# Changes

I added support for TaskRunResults for this task. This change adds four results:
- the sha256 hash of the bdist package and the sdist package.
- the package name
- the package version

I also fixed a permission issue in the sample - I could not get this to run as a non-root
user (even without my change). The git repo is cloned correctly, but the default user of the
twine image does not have write permissions in this directory.

# Submitter Checklist

review them:

- [X] Complies with [Catalog Orgainization TEP][TEP], see [example]. **Note** [An issue has been filed to automate this validation][validation]

          ```

          spec:
            description: >-

          ```

for more details._

---

### DIFF
--- a/task/upload-pypi/0.1/samples/run.yaml
+++ b/task/upload-pypi/0.1/samples/run.yaml
@@ -54,6 +54,10 @@ kind: PipelineRun
 metadata:
   name: publish-package-pipeline-run
 spec:
+  podTemplate:
+    securityContext:
+      runAsNonRoot: false
+      runAsUser: 0
   pipelineRef:
     name: publish-package-pipeline
   workspaces:
--- a/task/upload-pypi/0.1/README.md
+++ b/task/upload-pypi/0.1/README.md
@@ -49,3 +49,23 @@ spec:
 ```
 
 In this example, the Git repo being used is expected to have a `setup.py` file at the root of the repository. [setup.py](https://packaging.python.org/tutorials/packaging-projects/#creating-setup-py) is build script for [setuptools](https://pypi.org/project/setuptools/)
+
+This TaskRun outputs several `Results`:
+
+- A `sha256` hash for each uploaded file (the bdist and the sdist packages).
+- The name of the uploaded package
+- The version of the uploaded package
+
+This looks like:
+
+```
+  taskResults:
+  - name: bdist_sha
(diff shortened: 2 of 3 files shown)


task/upload-pypi/0.1/samples/run.yaml
```diff
@@ -54,6 +54,10 @@ kind: PipelineRun
 metadata:
   name: publish-package-pipeline-run
 spec:
+  podTemplate:
+    securityContext:
+      runAsNonRoot: false
+      runAsUser: 0
```
chmouel, line 60: Why would you need to run this as root?

Is that because of the python eggs cache trying to write to / ?

What abotu something like this ? https://stackoverflow.com/a/2193746

dlorenc: No, I don't think that's the issue. I tried those and they don't seem to work. In this case, here's the error I see:

```
[upload-pypi-run : build-package] running egg_info
[upload-pypi-run : build-package] error: could not create 'dlorenc_requests.egg-info': Permission denied
```

The directory created by the previous git-clone task is created by root, and is not world-writeable. Adding an `ls -l` before running shows that the directory is root owned:

```
[upload-pypi-run : build-package] + ls -la .
[upload-pypi-run : build-package] total 192
[upload-pypi-run : build-package] drwxr-xr-x 9 root root  4096 Jul 25 14:07 .
```

I actually don't know how this is supposed to work with the new git-clone task - maybe we should add a line there to set the permissions world-writeable?

dlorenc: Trying to run the entire pipeline as non-root doesn't work either:

```
[fetch-repository : clone] + CHECKOUT_DIR=/workspace/output/
[fetch-repository : clone] + '[[' true '==' true ]]
[fetch-repository : clone] + cleandir
[fetch-repository : clone] + '[[' -d /workspace/output/ ]]
[fetch-repository : clone] + rm -rf /workspace/output//lost+found
```

The clone step fails because it tries to clean up the directory first, and that is root owned.

chmouel:
/lgtm

yes you right this seems to be the issue with the git-clone being run as root by default on k8s i was misleaded while running on OpenShift which has a PodSecurityPolicy of forcing the image running as non root 

image

let's talk about what policy we want to set at our next working group, in the meantime let's get this in since the test was broken,

Issue #453, opened 2020-07-29
Reorganise openwhisk to conform with catalog re-org TEP


sthaha:
Creating this issue so that we don't miss the follow up reorganisation of the openwhisk related tasks, conditions and pipelines. 

`---8<---8<---8<---8<---8<---8<---8<---8<---8<---8<---8<---8<-`

@vdemeester @mrutkows Just noting that I am okay with merging the patch since the TEP is still in `proposed` but please be aware that the way it is currently organised doesn't [conform to proposal](https://github.com/tektoncd/community/blob/master/teps/0003-tekton-catalog-organization.md#organization) ([see example](https://github.com/tektoncd/community/blob/master/teps/0003-tekton-catalog-organization.md#example-of-a-resource-from-catalog))  and wouldn't show up in the Hub preview. 
This can be addressed in a separate patch of course.

I am also unsure about the `.tmpl` files. 

Since the pipeline is usable, I don't think we should bikeshed on it. However, we should discuss this further in our WG call. 

cc: @chmouel

_Originally posted by @sthaha in https://github.com/tektoncd/catalog/pull/375#issuecomment-664987708_

mrutkows:
@sthaha Please assign to me... I already have another branch that conforms with the updates made to the TEP 9 days ago...

PR #457, opened 2020-07-30
Reorg. OpenWhisk pipeline to conform to latest proposal

mrutkows:
These changes reflect follow-on work that will reorganize the OpenWhisk pipeline catalog entry to conform to the latest hub layout proposal.  Primarily this involves placing most of the pipeline and knative example files under a samples directory along with assuring the openwhisk entry has the correct tags and labels.

For reference, the TEP being used for informing this work is here:
https://github.com/tektoncd/community/blob/master/teps/0003-tekton-catalog-organization.md

# Submitter Checklist

review them:

          ```

          spec:
            description: >-

          ```

for more details._

---


mrutkows:
Fixes https://github.com/tektoncd/catalog/issues/453

mrutkows:
@chmouel @sthaha Any further advice on this re-org (such as direction on where to place service account and docker secret related YAMLs)?

sthaha:
@mrutkows The changes look fine to me but there may be few things you may want to fix. 
1. The current `displayName` is "OpenWhisk application Pipeline for Knative". Can we make it shorter? perhaps `OpenWhisk for Knative` ? We know its a pipeline from its kind
2. Can we move the `condition` (a Tekton Resource, that may be sharable) to `/condition/openwhisk-runtime-js/0.1/openwhisk-runtime-js` and so on and so forth ? 

3. On a related note, we could make it generic so that others can use it as well  by changing `OW_APP_PATH` to `APP_PATH` ? and add the recommended annotations and labels
4. The same for other Runtime Detection as well. 

Hub can easily add support for displaying conditions

mrutkows:
Hi Sunil, thanks for your comments.  Here are my thoughts against each: 

>     1. move the `deploy.sh`, `workspaces`  to `support/` directory (Guideline: all files required to support running a pipeline/task to in `support` dir)

Fine with moving our top-level script to a subdir.

>     3. `runtime` need to move under `samples` IIUC

Fine with this as well

>     4. The current `displayName` is "OpenWhisk application Pipeline for Knative". Can we make it shorter? perhaps `OpenWhisk for Knative` ? We know its a pipeline from its kind

Is there an imperative to make it shorter (a limit for fitting in the tile?  Since there are several things that we could have OpenWhisk pipelines for besides "app" building (e.g., building providers, building the base images themselves, building the platform components, building IDE plugins, etc.) the inclusion of "application" seems to clarify; removing pipeline is of course fine.

>     2. `tasks` need to move out to `/task/openwhisk-<taskname>/0.1/openwhisk-<taskname>.yaml`
>     5. Can we move the `condition` (a Tekton Resource, that may be sharable) to `/condition/openwhisk-runtime-js/0.1/openwhisk-runtime-js` and so on and so forth ?

I disagree. These tasks and condition(s) are not intended to be reusable catalog entries, but only intended to be uses and maintained within the context of the pipeline itself.  Please do not force me to effectively break apart the pipeline.

>     6. On a related note, we could make it generic so that others can use it as well  by changing `OW_APP_PATH` to `APP_PATH` ? and add the recommended annotations and labels
>     7. The same for other Runtime Detection as well.

I would prefer to keep the "OW_" prefix as it was a conscious decision to indicate the source application is coded to OpenWhisk conventions and paradigms.

sthaha:
> Hi Sunil, thanks for your comments. Here are my thoughts against each:
> > ```
> > 4. The current `displayName` is "OpenWhisk application Pipeline for Knative". Can we make it shorter? perhaps `OpenWhisk for Knative` ? We know its a pipeline from its kind
> > ```
>
> Is there an imperative to make it shorter (a limit for fitting in the UI tile)? Since there are several things that we could have OpenWhisk pipelines for besides "app" building (e.g., building providers, building the base images themselves, building the platform components, building IDE plugins, etc.) the inclusion of "application" seems to clarify; removing pipeline is of course fine.

Good question! I am not sure but we should decide on some arbitrary value independent of how `Hub` is able to render it now and I agree with including `application` if it clarifies the purpose of the pipeline.

>
> > ```
> > 2. `tasks` need to move out to `/task/openwhisk-<taskname>/0.1/openwhisk-<taskname>.yaml`
> > 5. Can we move the `condition` (a Tekton Resource, that may be sharable) to `/condition/openwhisk-runtime-js/0.1/openwhisk-runtime-js` and so on and so forth ?
> > ```
>
> I disagree.

 +1 I think I see your point of view. IIUC, the question here is, where do you
 keep all the supporting resources that are specific to the current resources
 and aren't reusable components.

 We definitely need to cover this in our TEP.

>  These tasks and condition(s) are not intended to be reusable catalog entries, but only intended to be uses and maintained within the context of the pipeline itself. Please do not force me to effectively break apart the pipeline.

+1. How about we keep this under `./support/<kind>/name.yaml` as there won't be
any "version" for these supporting resources. This also makes it easy for a
tool to parse/search for resources, validate and list those.

>
> It is my goal to create the best pipeline for OpenWhisk and evolve the tasks to be better over time, but the reality is I do not at this time wish to maintain, for example a Condition to test for Java, for a larger community at this time. Maybe eventually... but I have seen several other open source projects like Paketo work on app/image detection which would be a better building block... and I have no time now to explore that and feel that eventually someone will come along with a better "Detect" task (perhaps using a Paketo builder image detect) that I might be able to delete mine in favor of.

I think we both agree here :-)
>
> > ```
> > 6. On a related note, we could make it generic so that others can use it as well  by changing `OW_APP_PATH` to `APP_PATH` ? and add the recommended annotations and labels
> > 7. The same for other Runtime Detection as well.
> > ```
>
> I would prefer to keep the "OW_" prefix as it was a conscious decision to indicate the source application is coded to OpenWhisk conventions and paradigms.

+1 based on the idea that these are supporting resources, not sharable and are intended to be used only with this pipeline.

mrutkows:
Fixes https://github.com/tektoncd/catalog/issues/484

PR #459, opened 2020-07-31
Fixes for pylint tasks

chmouel:
# Changes

* Only intall the requirements when the file is here
  (sometime we write pure python script that doesn't need external libs)
* fix installing pylint when needed

# Submitter Checklist

review them:

          ```

          spec:
            description: >-

          ```

### DIFF
--- a/task/pylint/0.1/pylint.yaml
+++ b/task/pylint/0.1/pylint.yaml
@@ -37,9 +37,16 @@ spec:
   steps:
     - name: lint
       image: docker.io/python:$(inputs.params.PYTHON)
-      workingDir: /workspace/source
+      workingDir: $(workspaces.source.path)
       script: |
         export PATH=$PATH:$HOME/.local/bin
-        pip install -r $(inputs.params.SOURCE_PATH)/$(inputs.params.REQUIREMENTS_FILE)
-        pip show pylint || echo "###\nWarning: Pylint is missing in your requirements\n###" && pip install pylint
+        if [ -n "$(inputs.params.REQUIREMENTS_FILE)" ] && [ -e "$(inputs.params.REQUIREMENTS_FILE)" ];then
+           pip install -r $(inputs.params.SOURCE_PATH)/$(inputs.params.REQUIREMENTS_FILE)
+           pip show pylint || {
+              echo "###\nWarning: Pylint is missing in your requirements\n###";
+              pip install pylint
+           }
+        else
+           pip install pylint
+        fi
         pylint $(inputs.params.ARGS) $(inputs.params.MODULE_PATH)


task/pylint/0.1/pylint.yaml
```diff
@@ -37,9 +37,16 @@ spec:
   steps:
     - name: lint
       image: docker.io/python:$(inputs.params.PYTHON)
-      workingDir: /workspace/source
+      workingDir: $(workspaces.source.path)
       script: |
         export PATH=$PATH:$HOME/.local/bin
-        pip install -r $(inputs.params.SOURCE_PATH)/$(inputs.params.REQUIREMENTS_FILE)
-        pip show pylint || echo "###\nWarning: Pylint is missing in your requirements\n###" && pip install pylint
+        if [ -n "$(inputs.params.REQUIREMENTS_FILE)" && -e "$(inputs.params.REQUIREMENTS_FILE)" ];then
```
dlorenc, line 43: What does "-n" do? I've never seen that one before, and am having trouble finding it. https://tldp.org/LDP/abs/html/fto.html doesn't have it listed.

task/pylint/0.1/pylint.yaml
```diff
@@ -37,9 +37,16 @@ spec:
...
+           pip install -r $(inputs.params.SOURCE_PATH)/$(inputs.params.REQUIREMENTS_FILE)
+           pip show pylint || {
+              echo "###\nWarning: Pylint is missing in your requirements\n###";
+              pip install pylint
+           }
+        else
+           pip install pylint
```
dlorenc, line 50: Would it make sense to pin this to a version here?

task/pylint/0.1/pylint.yaml
chmouel: checks if the var is defined  : 

```
$ [[ -n ${FOO} ]] && echo "yes"
chmouel@TheBatman:~
$ FOO=1
chmouel@TheBatman:~
$ [[ -n ${FOO} ]] && echo "yes"
yes
```

see man test : 

```     -n string     True if the length of string is nonzero.```

File: task/pylint/0.1/pylint.yaml (hunk 2 above)
Comment (chmouel) [reply task/pylint/0.1/pylint.yaml:50]: I am not really sure if that's really needed, this is not really common in python world, and if the user really wants to be pinned they can have it in their requirements file..

Comment (chmouel) [reply task/pylint/0.1/pylint.yaml:50]: I think we could use this docker image directl too which has the binary https://hub.docker.com/r/cytopia/pylint

but i believe the original author of this task did it like this so the user can choose her preferred python version at task time (which is actually something done quite often in python having the need for different version of python).

Comment (wumaxd) [reply task/pylint/0.1/pylint.yaml:50]: Yes the idea was here to keep it completely flexible for the user of this task.

As already mentioned by @chmouel we could change the python version for this task (and also the pytest task) to 3.8 or even to latest as it is based on the based python image.

**Comment by wumaxd** [comment]
@chmouel the same fix is needed for the pytest task. I update the task if this change here is approved.

**Comment by chmouel** [comment]
@wumaxd sounds good to me thanks! 

(fyi feel free to add your /lgtm if you think it could have one since even non OWNERS can add a review)

File: task/pylint/0.1/pylint.yaml [hunk 3]
```diff
@@ -37,9 +37,16 @@ spec:
   steps:
     - name: lint
       image: docker.io/python:$(inputs.params.PYTHON)
-      workingDir: /workspace/source
+      workingDir: $(workspaces.source.path)
       script: |
         export PATH=$PATH:$HOME/.local/bin
-        pip install -r $(inputs.params.SOURCE_PATH)/$(inputs.params.REQUIREMENTS_FILE)
-        pip show pylint || echo "###\nWarning: Pylint is missing in your requirements\n###" && pip install pylint
+        if [ -n "$(inputs.params.REQUIREMENTS_FILE)" ] && [ -e "$(inputs.params.REQUIREMENTS_FILE)" ];then
```
vinamra28, line 43: can we set the default value to `""` in `REQUIREMENTS_FILE` params at line no [36](https://github.com/tektoncd/catalog/pull/459/files#diff-4b6a867b7d4f0c4e83dbaa5d9ad0ba25L36) and allow user to pass the `requirements.txt` filepath in the `TaskRun` so that it is more user-controlled :thinking:

chmouel: well you can pass the REQUIREMENTS_FILE already, is what you want to not set it as default ?

vinamra28: Yes

chmouel: it's both way user-controlled, isnt it ?

vinamra28: yes it is now also user-controlled, just presented my thought on it :sweat_smile:

chmouel:
/retest

docker build test seems flaky

PR #462, opened 2020-08-03
Add support for REST services testing

chmouel:
Add support for REST services testing.

We do this using https://github.com/chmouel/go-rest-api-test/. The idea is that
we let the user specify on which URL and method the service will reply with some
outputs.

The way we implement this is to have the e2e-tests detect if we have a directory in the tests
directory called `fixtures` with a yaml file that contains the responder
rules (see github-add-comments task example) it will then add a sidecar task to
the task.

The test can then using http://localhost:8080 to point to it.


bobcatfish:
hey @chmouel ! Seems like a reasonable feature to add. I do wonder if static content will cover enough use cases (e.g. will the tests want more dynamic content)

Given the overall plan to move (slowly) to a Tekton Pipelines based setup for our e2e tests, what would you think implementing this via a Task? (I wonder if this would be another use case for a pipeline level sidecar https://github.com/tektoncd/pipeline/issues/2973)

task/github-add-comment/0.1/github-add-comment.yaml
```diff
@@ -71,7 +71,12 @@ spec:
         print("Sending this data to GitHub: ")
         print(data)
 
-        conn = http.client.HTTPSConnection("$(params.GITHUB_HOST_URL)")
+        # This is for our fake github server
+        if "$(params.GITHUB_HOST_URL)".startswith("http://"):
```
sthaha, line 75: should we also look at https?

task/github-add-comment/0.1/github-add-comment.yaml
```diff
@@ -71,7 +71,12 @@ spec:
         print("Sending this data to GitHub: ")
         print(data)
 
-        conn = http.client.HTTPSConnection("$(params.GITHUB_HOST_URL)")
+        # This is for our fake github server
```
sthaha, line 74: +1

task/github-add-comment/0.1/github-add-comment.yaml
sthaha, line 75: would `fake://` be better (more legible? ) ?

test/e2e-common.sh
```diff
@@ -125,6 +125,19 @@ function test_task_creation() {
             cp ${yaml} ${TMPF}
             [[ -f ${taskdir}/tests/pre-apply-task-hook.sh ]] && source ${taskdir}/tests/pre-apply-task-hook.sh
             function_exists pre-apply-task-hook && pre-apply-task-hook
+
+            [[ -d ${taskdir}/tests/fixtures ]] && {
```
sthaha, line 129: ♥️ , I think the magic needs a bit more explanation as to how this works in the commit message.

task/github-add-comment/0.1/github-add-comment.yaml
chmouel: I think in general it's probably better not to have implementation be 'aware' of test and handle case in case if we are in testing mode. 

And i think here having to use a `HTTP` connexion instead of `HTTPS` connexion....

I have actually spent spend some at first adding HTTPS support to the mockserver, but having to add the generation of the keys and having the tests using those keys for connecting was getting  this review more complicated to understand for a first POC PR, maybe that's something we can add in the future.

chmouel:
> hey @chmouel ! Seems like a reasonable feature to add. I do wonder if static content will cover enough use cases (e.g. will the tests want more dynamic content)

what use case for dynamic content would you have in mind ?  It should be quite flexible and there is way to expand that if there is limitations..

In general I have been trying to avoid go down the path to end up testing APIs, in for example this github-add-comment task we are not trying to validate if the task sent us the right API queries/params to our mockserver, the only thing we want is to have it answering properly so the task can goes on. 

The mockserver is giving the opportunity to be able to have the test testing the task flow itself and not block on external services.

task/github-add-comment/0.1/github-add-comment.yaml
chmouel: > Given the overall plan to move (slowly) to a Tekton Pipelines based setup for our > e2e tests, what would you think implementing this via a Task? (I wonder if this would be another use case for a pipeline level sidecar tektoncd/pipeline#2973)

I don't believe this would be an impediment when tekton pipelines based e2e tests are implemented. We are still launching a `run.yaml` for testing the task. the mais issue is how to figure out to do task injection, i mean, how do we inject a sidecar into a `Task` without having to manipulate the Task yaml (like we do for the builder tasks and here for the mockserver). If there was a way to have this via the `TaskSpec` in the `Run` then things would be quite transparent and portable.

I did go over an iteration where we spin up a `Deployment` and a `Service` for all tasks to use directly instead of using the sidecar injection things, but things would get more complicated with rules, ie: what do we do when two services implement /api/users/1 how can we differentiate between the two.

chmouel:
> Given the overall plan to move (slowly) to a Tekton Pipelines based setup for our > e2e tests, what would you think implementing this via a Task? (I wonder if this would be another use case for a pipeline level sidecar tektoncd/pipeline#2973)

I don't believe this would be an impediment when tekton pipelines based e2e tests are implemented. We are still launching a `run.yaml` for testing the task. the mais issue is how to figure out to do task injection, i mean, how do we inject a sidecar into a `Task` without having to manipulate the Task yaml (like we do for the builder tasks and here for the mockserver). If there was a way to have this via the `TaskSpec` in the `Run` then things would be quite transparent and portable.

I did go over an iteration where we spin up a `Deployment` and a `Service` for all tasks to use directly instead of using the sidecar injection things, but things would get more complicated with rules, ie: what do we do when two services implement /api/users/1 how can we differentiate between the two.

test/e2e-common.sh
chmouel: It is I think in the commit message, but I definitively would update the [CONTRIBUTING](https://github.com/tektoncd/catalog/blob/master/CONTRIBUTING.md#end-to-end-testing) guide on how it works whe we find a consensus on how we want to implement this

chmouel:
/close

I guess there is no interest for this...

chmouel:
resurecting this since that's something very much needed for some tasks to be tested

chmouel:
I have added the tests for the pull request task in 370ef0455583d8b723e764eceada5c3145594a92 to give another example how testing with multiple api request is done.

Still a WIP since i need to finish writing some documentation.

chmouel:
/hold cancel

Ready for review, I made https://github.com/chmouel/go-rest-api-test a bit more robust as well... Documentation should be up to date https://github.com/tektoncd/catalog/pull/462/commits/cf1ca4bf5c237163554bdea3059cdc06edb1ef86

vdemeester (approved):
/meow

PR #473, opened 2020-08-05
Add Tasks to install Tekton Pipelines itself 🤯

bobcatfish:
# Changes

As part of #373 I'm continuing to create Tasks that we can use to run
tests for items in the catalog.

The goal of the 2 Tasks in this commit is to install Tekton Pipelines into a
cluster that was created using the Task added in #436.

There are a couple of options different from what I did in this commit
which I'd like people's opinions on:
1. I went to a lot of trouble to make it so that Pipelines can be
   installed with just a kubectl command, specifically creating a user
   with cluster admin priviledges. If instead I used a container with
   gcloud, I could use the same approach as other GKE specific Tasks,
   but I wanted to make the install Task not tied to GKE. I also had
   to give the prow-account more priviledges in all of our boskos
   projects for this to work 😬 . So my question is: is this worth
   it, or am I better off just creating a GKE specific Task for
   installing Tekton Pipelines that uses gcloud to auth?
2. If we DO want a super generic Task, how useful is this Task that just
   applies a Tekton yaml? Would we be better off with a generic kubectl
   task that I could feed the exact arguments for installing Tekton
   Pipelines?

The other confusion I ran into here (which I am going to follow up with
@sbwsg about!) is around service accounts vs. secrets. The GKE Tasks
I've made so far rely on secrets provided via workspaces, but I feel
like it's not clear if it makes sense to go that route, or to rely on
service accounts. In the meantime this approach is at least consistent
with #436.

# Submitter Checklist

review them:

          ```

          spec:
            description: >-

          ```

for more details._

---

### DIFF
--- a/task/gke-add-cluster-admin/0.1/gke-add-cluster-admin.yaml
+++ b/task/gke-add-cluster-admin/0.1/gke-add-cluster-admin.yaml
@@ -1,0 +1,64 @@
+apiVersion: tekton.dev/v1beta1
+kind: Task
+metadata:
+  name: gke-add-cluster-admin
+  labels:
+    app.kubernetes.io/version: "0.1"
+  annotations:
+    tekton.dev/pipelines.minVersion: "0.12.1"
+    tekton.dev/tags: "gke"
+    tekton.dev/displayName: "GKE Add Cluster Admin"
+spec:
+  description: |
+    Create a cluster admin user in a GKE cluster.
+
+    This Task will authenticate with a GKE cluster using gcloud and then create a new
+    user in the cluster with cluster admin priviledges.
+    The service account provided via gcp-service-account must have `container.clusterRoleBindings.create`
+    permissions in the target GKE project that the cluster lives in.
+
+  params:
+  - name: private-key-path
+    description: The path to the private key within the gcp-service-account workspace.
+  - name: new-username
+    description: The name to use for the new user
+    default: cluster-admin
+  workspaces:
+  - name: gcp-service-account
+    description: |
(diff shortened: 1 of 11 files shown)


bobcatfish:
Interested in your thoughts on the 2 questions above @dlorenc @vdemeester @sbwsg @afrittoli  ^^

bobcatfish:
```
+ echo 'FAILED: docker-build task has failed to comeback properly'
FAILED: docker-build task has failed to comeback properly
```

a flake maybe? 🤔

afrittoli:
@bobcatfish I will look into this once I'm back. Just FYI - this is what we use today to install Tekton (pipeline, trigger and dashboard) to the `robocat` and `dogfooding` clusters today: https://github.com/tektoncd/plumbing/blob/master/tekton/resources/release/install_tekton_release.yaml - it uses the release yaml but it adds the ability to use `kustomize` overlays from a git repo on top. I find that would be useful for test purposes too, since one might want to test different configurations of the tekton services in E2E tests.

bobcatfish:
Thanks @afrittoli ! i probably wont get a change to make sweeping changes to this until next week but it looks like im better off using the Tasks youve already created

when you get a chance could you explain a bit more about the overlays you need? i'm also wondering how the Task is supplied with the credentials it needs, is that via the service account?

bobcatfish:
/hold

want to integrate in the Tasks @afrittoli has already created for this: https://github.com/tektoncd/plumbing/blob/master/tekton/resources/release/install_tekton_release.yaml

bobcatfish:
I'll just close this for now actually to reduce the noise :D

Issue #441, opened 2020-07-23
Guidance for writing Tasks: When is it appropriate to have multiple steps?


bobcatfish:
# Expected Behavior

It'd be nice to provide users with best practices around when it is appropriate to use multiple steps and when it makes sense to combine into one.

## Christie's dream ☁️ 

It'd be great if each step of a Task could do one thing well and one thing only. Each step could be a small image that only does one thing, and the image itself is well tested. A Task then becomes a sequence of highly cohesive tiny images each doing one thing.

# Actual Behavior

In #408 @vdemeester and I disagreed about which of these approaches to take, split between these two options:

Two steps, one of which uses a shell to create yaml configuration:
```yaml
  - name: create-heartbeat-pod-yaml
    image: bash
    script: |
      FULL_RESOURCE_OUTPUT=$(cat /workspace/full-resource-output.json)
      LEASED_RESOURCE=$(cat /tekton/results/leased-resource)
      cat <<EOF > /workspace/heartbeat.yaml
      apiVersion: v1
      kind: Pod
      metadata:
        name: boskos-heartbeat-$LEASED_RESOURCE
      spec:
        containers:
        - name: heatbeat
          image: gcr.io/k8s-staging-boskos/boskosctl@sha256:a7fc984732c5dd0b4e0fe0a92e2730fa4b6bddecd0f6f6c7c6b5501abe4ab105
          args:
          - heartbeat
          - --server-url=$(params.server-url)
          - --owner-name=$(params.owner-name)
          - --resource=$FULL_RESOURCE_OUTPUT
          - --period=5m
      EOF
  - name: start-boskosctl-heartbeat
    image: lachlanevenson/k8s-kubectl
    args:
    - "apply"
    - "-f"
    - "/workspace/heartbeat.yaml"
```

One step, which is able to both use a shell and execute kubectl because the image already contains a shell:
```yaml
  - name: create-heartbeat-pod-yaml
    image: lachlanevenson/k8s-kubectl
    script: |
      FULL_RESOURCE_OUTPUT=$(cat /workspace/full-resource-output.json)
      LEASED_RESOURCE=$(cat /tekton/results/leased-resource)
      cat <<EOF | kubectl apply -f -
      apiVersion: v1
      kind: Pod
      metadata:
        name: boskos-heartbeat-$LEASED_RESOURCE
      spec:
        containers:
        - name: heatbeat
[...]

PR #480, opened 2020-08-10
Add Task to install Tekton components using Released Tekton Operator

vinamra28:
# Changes

The following task uses the released version of tekton operator to install tekton components on a new cluster. By default Tekton pipelines are installed but if we want to install other components as well then we have boolean parameters for respective components

# Submitter Checklist

review them:

- [X] Complies with [Catalog Orgainization TEP][TEP], see [example]. **Note** [An issue has been filed to automate this validation][validation]

          ```

          spec:
            description: >-

          ```

for more details._

---

### DIFF
--- a/task/tekton-operator-install/0.1/tekton-operator-install.yaml
+++ b/task/tekton-operator-install/0.1/tekton-operator-install.yaml
@@ -1,0 +1,217 @@
+apiVersion: tekton.dev/v1beta1
+kind: Task
+metadata:
+  name: tekton-operator-install
+  labels:
+    app.kubernetes.io/version: "0.1"
+  annotations:
+    tekton.dev/pipelines.minVersion: "0.12.1"
+    tekton.dev/tags: operator, tekton, deploy
+    tekton.dev/displayName: "Tekton Operator Install"
+spec:
+  description: >-
+    This task can be used to install Tekton pipelines and also it's
+    components using Tekton Operator on a new cluster.
+  workspaces:
+    - name: kubeconfig
+      description: >-
+        The workspace consisting of kubeconfig file so that operator
+        can install tekton components on fresh cluster
+  params:
+    - name: OPERATOR_VERSION
+      type: string
+      description: Tekton operator version to be installed
+      default: "latest"
+    - name: INSTALL_TRIGGERS
+      type: string
+      description: If triggers needs to be installed
+      default: "false"
+    - name: INSTALL_DASHBOARD
+      type: string
+      description: If Tekton dashboard to be installed
+      default: "false"
+    - name: INSTALL_EXTENSION_WEBHOOKS
+      type: string
+      description: If Tekton Extension Webhooks needs to be installed
+      default: "false"
+  steps:
+    - name: install-operator
+      image: lachlanevenson/k8s-kubectl@sha256:3a5e22a406a109f4f26ec06b5f1f6a66ae0cd0e185bc28499eb7b7a3bbf1fe09
+      workingDir: $(workspaces.kubeconfig.path)
+      script: |
+        #!/bin/sh
+        set -u -o pipefail
+
+        [[ -f ./kubeconfig ]] && \
+        export KUBECONFIG=./kubeconfig || \
+        (echo "please provide kubeconfig file" && exit 1)
+
+        # check whether CR is installed or not
+        checkCodeStatus() {
+          local crd_name=${1}
+          local cr_name=${2}
+          kubectl get "${crd_name}" "${cr_name}" \
+            -o jsonpath='{range .status.conditions[0]}{.code}{"\n"}{end}'
+        }
+
+        # check the version of the installed CR
+        checkVersionStatus() {
+          local crd_name=${1}
+          local cr_name=${2}
+          kubectl get "${crd_name}" "${cr_name}" \
+            -o jsonpath='{range .status.conditions[0]}{.version}{"\n"}{end}'
+        }
+
+        show_failure() {
+          echo "failed to install $1 $2"
+          exit 1
+        }
+
+        main(){
+          kubectl apply -f https://storage.googleapis.com/tekton-releases/operator/$(params.OPERATOR_VERSION)/release.notags.yaml
+          local maxloop=30 # 5 minutes max
+
+          # check operator and pipeline correctly installed or not
+          # display the version of pipeline
+          local count=0
+          while true; do
+            # hard timeout condition
+            [[ ${count} == ${maxloop} ]] && show_failure "tektonpipelines" "cluster"
+
+            local codeStatus=$(checkCodeStatus "tektonpipelines" "cluster")
+            [[ "$codeStatus" == "installed" ]] && \
+            checkVersionStatus "tektonpipelines" "cluster" && break
+            sleep 10
+            count=$((count+1))
+          done
+          echo "successfully installed operator version $(params.OPERATOR_VERSION)"
+        }
+
+        main "$@"
+
+    - name: install-addon
+      image: lachlanevenson/k8s-kubectl@sha256:3a5e22a406a109f4f26ec06b5f1f6a66ae0cd0e185bc28499eb7b7a3bbf1fe09
+      workingDir: $(workspaces.kubeconfig.path)
+      script: |
+        #!/bin/sh
+        set -u -o pipefail
+
+        [[ -f ./kubeconfig ]] && \
+        export KUBECONFIG=./kubeconfig || \
+        (echo "please provide kubeconfig file" && exit 1)
+
+        # create TektonAddon CR manifest and apply that directly
+        createCustomResource() {
+          cr_name=${1}
+          cat <<EOF | kubectl apply -f -
+        apiVersion: operator.tekton.dev/v1alpha1
+        kind: TektonAddon
+        metadata:
+          name: ${cr_name}
+        EOF
+        }
+
+        # check whether CR is installed or not
+        checkCodeStatus() {
+          local crd_name=${1}
+          local cr_name=${2}
+          kubectl get "${crd_name}" "${cr_name}" \
+            -o jsonpath='{range .status.conditions[0]}{.code}{"\n"}{end}'
+        }
+
+        # check the version of the installed CR
+        checkVersionStatus() {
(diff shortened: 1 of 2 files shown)


task/tekton-operator-install/0.1/tekton-operator-install.yaml
```diff
@@ -0,0 +1,79 @@
...
+
+        if [[ -n $(params.TRIGGERS_VERSION) ]]; then
+          kubectl apply -f trigger_cr.yaml
+        elif [[ -n $(params.DASHBOARD_VERSION) ]]; then
+          kubectl create clusterrolebinding tekton-operator-cluster-admin \
+          --clusterrole cluster-admin --serviceaccount tekton-operator:tekton-operator
+          kubectl apply -f dashboard_cr.yaml
+        elif [[ -n $(params.EXTENSION_WEBHOOKS_VERSION) ]]; then
+          kubectl apply -f extension_webhooks_cr.yaml
+        fi
```
savitaashture, line 79: I assume all 3 installation are independent so i feel we should not bind with `if/elif` condition :thinking: 
Even we can use switch instead of `if` condition

savitaashture:
Thank you for the PR finally we can install tekton using task :tada:

task/tekton-operator-install/0.1/tekton-operator-install.yaml
savitaashture, line 79: I assume all 3 installation are independent so i feel we should not bind with if/elif condition :thinking: 

Even we can use switch instead of if condition

vinamra28: But in switch we have two scenarios:-
1. `break` after executing the case means if we want to install two components then after installing one component it will break out from the switch thus not installing other component
2. `fallthrough` to next case which means that if suppose I want to install triggers then dashboard and other components also get's installed
WDYT?? :thinking:

task/tekton-operator-install/0.1/tekton-operator-install.yaml
```diff
@@ -0,0 +1,79 @@
...
+        cat << EOF > extension_webhooks_cr.yaml
+        apiVersion: operator.tekton.dev/v1alpha1
+        kind: TektonAddon
+        metadata:
+          name: extensionwebhooks
+        spec:
+          version: $(params.EXTENSION_WEBHOOKS_VERSION)
+        EOF
+
+        kubectl apply -f https://storage.googleapis.com/tekton-releases/operator/latest/release.notags.yaml
```
vdemeester, line 69: We *may* want to be able to specify the version of the operator release too (defaulting to latest)

nikhil-thomas:
@vinamra28 this looks is looks great. :+1: 

just a few suggestions,

1. **Boolean params (Triggers and Dashboard)**: lets make the Triggers and Dashboard params boolean. We are moving away from supporting older versions of addons.
    so keep the 2 params, make default false. We can create the Addons CRs without providing versions param and it will install the latest.

2. **Validation Step**: this task needs a 2nd step, validation.  A simple approach to validate the install is to check the status of operator CR (name: cluster) which is an instance of `tektonpipeline.operator.tekton.dev`. `status.conditions[0].code == "installed"`. If addons are enabled we can check their status as well.

3. Im not sure whether we have to provide support for extensions through this task. It is experimental. In addition if a user needs it any way the CR can be created after running the task as well (special case).

cc @savitaashture @vdemeester

task/tekton-operator-install/0.1/tekton-operator-install.yaml
JustinKuli: Should they all be separate `if`s?

task/tekton-operator-install/0.1/tekton-operator-install.yaml
```diff
@@ -0,0 +1,79 @@
...
+      type: string
+      description: The version of the Tekton dashboard to be installed
+      default: ""
+    - name: EXTENSION_WEBHOOKS_VERSION
+      type: string
+      description: >-
+        The version of the Tekton Extension Webhooks to be installed
+      default: ""
+  steps:
+    - name: intall-components
```
nikhil-thomas, line 35: typo, `install`

shall we split this step into 2 `setup` and `install`

in setup we can write the manifests in to a dir after checking the params

in install we can call `kubectl apply -f <path to dir> ` after installing the operator

vinamra28: can we avoid writing two steps because I am taking care of the following issue https://github.com/tektoncd/catalog/issues/441. I can understand your point that it will make the code more cleaner :sweat_smile:

task/tekton-operator-install/0.1/tekton-operator-install.yaml
savitaashture: Yes if that's the case i agree with @JustinKuli we can use `if`s

task/tekton-operator-install/0.1/tekton-operator-install.yaml
nikhil-thomas: i see, cool :+1: 

then could you move 
```
kubectl apply -f https://storage.googleapis.com/tekton-releases/operator/latest/release.notags.yaml
```
to the top (right after KUBECONFIG check)

It would be simpler if we move the CR creating part into a function. The function will need only addon name as argument. We can write the manifest to a known path (`addon-manifest-dir`)

this would help us check params, call the function to generate CR manifests. and finally call a `kubectl apply -f` on the addon-manifest-dir.

vdemeester: > can we avoid writing two steps because I am taking care of the following issue #441. I can understand your point that it will make the code more cleaner sweat_smile

I actually think it's a good fit for 2 steps :upside_down_face: one that deploy the thing, one that waits for it to be correctly installed. Steps are relatively cheap (especially if using the same image).

vinamra28:
@nikhil-thomas @vdemeester broke the task into two steps. What I did over here is:-
step 1:- Apply the operator release manifest and wait for it's successful installation
After the operator and pipelines are successfully installed then we will proceed with step 2
step 2:- Apply the addons manifests asked by the user and wait for it's successful installation.
Please let me know your thoughts on it :sweat_smile:

task/tekton-operator-install/0.1/tekton-operator-install.yaml
```diff
@@ -0,0 +1,187 @@
...
+      description: The version of the triggers to be installed
+      default: "false"
+    - name: INSTALL_DASHBOARD
+      type: string
+      description: The version of the Tekton dashboard to be installed
+      default: "false"
+    - name: INSTALL_EXTENSION_WEBHOOKS
+      type: string
+      description: >-
+        The version of the Tekton Extension Webhooks to be installed
```
nikhil-thomas, line 36: :+1: for switching to boolean params
could you update the description(s) as well

task/tekton-operator-install/0.1/tekton-operator-install.yaml
```diff
@@ -0,0 +1,187 @@
...
+            -o jsonpath='{range .status.conditions[0]}{.version}{"\n"}{end}'
+        }
+
+        show_failure() {
+          echo "failed to install $1 $2"
+          exit 1
+        }
+
+        main(){
+          kubectl apply -f https://storage.googleapis.com/tekton-releases/operator/$(params.OPERATOR_VERSION)/release.notags.yaml
```
nikhil-thomas, line 69: :+1: `notags.yaml`

task/tekton-operator-install/0.1/tekton-operator-install.yaml
```diff
@@ -0,0 +1,187 @@
...
+          echo "failed to install $1 $2"
+          exit 1
+        }
+
+        main() {
+          local maxloop=30 # 5 minutes max
+          local cnt=0
+          local trigger_success=true
+          local dashboard_success=true
+          local extension_success=true
```
nikhil-thomas, line 135: it would be better if we can set this to `false` by default

task/tekton-operator-install/0.1/tekton-operator-install.yaml
```diff
@@ -0,0 +1,187 @@
...
+
+          [[ $(params.INSTALL_TRIGGERS) == "false" ]] && \
+          [[ $(params.INSTALL_DASHBOARD) == "false" ]] && \
+          [[ $(params.INSTALL_EXTENSION_WEBHOOKS) == "false" ]] && \
+          echo "No addons to be installed" && exit 0
+
+          [[ $(params.INSTALL_TRIGGERS) == "true" ]] && \
+          createCustomResource "trigger" && trigger_success=false
```
nikhil-thomas, line 143: if the default of `trigger_success` is false we won't have to set it to `false` again here

task/tekton-operator-install/0.1/tekton-operator-install.yaml
```diff
@@ -0,0 +1,187 @@
...
+            checkVersionStatus "tektonaddon" "dashboard" && dashboard_success=true
+
+            #if extension webhooks needs to be installed then check the status
+            [[ $(params.INSTALL_EXTENSION_WEBHOOKS) == "true" ]] && [[ $extension_success == "false" ]] && \
+            [[ $(checkCodeStatus "tektonaddon" "extensionwebhooks") == "installed" ]] && \
+            checkVersionStatus "tektonaddon" "extensionwebhooks" && extension_success=true
+
+            # break if required Addons are successfully installed
+            [[ "$trigger_success" == "true" ]] && [[ "$dashboard_success" == "true" ]] && \
+            [[ "$extension_success" == "true" ]] && break
```
nikhil-thomas, line 180: shall we break this into 3 different checks, 
if the TaskRun param is true, then we can check `<item>_success` 
this could be handled as a function as well.

nikhil-thomas (changes requested):

task/tekton-operator-install/0.1/tekton-operator-install.yaml
vinamra28: addressed :+1:

chmouel:
/lgtm
/approve

Thanks 👍🏼🔥

Issue #481, opened 2020-08-10
docker-build test task is failing


chmouel:
# Expected Behavior

No failure :) 

# Actual Behavior

The test are sometime failing : 

`FAILED: docker-build task has failed to comeback properly`

The main issue seems to be with this : 

```
+ docker build --no-cache -f ./Dockerfile -t localhost:5000/nocode .
unable to resolve docker endpoint: open /certs/client/ca.pem: no such file or directory
```

# Steps to Reproduce the Problem

1. Check the log [here](https://storage.googleapis.com/tekton-prow/pr-logs/pull/tektoncd_catalog/479/pull-tekton-catalog-integration-tests/1292831095561129984/build-log.txt) from this [issue](https://github.com/tektoncd/catalog/pull/479#issuecomment-671397322)
2.
3.

# Additional Info

chmouel:
cc @popcor255 @PuneetPunamiya @ImJasonH @vdemeester 

(people in the owners file for docker-build task)

it is breaking the catalog CI fyi,

ImJasonH:
This is unexpected, since the Task's sidecar shouldn't report as ready until the certs are generated and available: https://github.com/tektoncd/catalog/blob/cb2b319b20b99bf48fd7d636fbe35afb170976b3/task/docker-build/0.1/docker-build.yaml#L100-L103

popcor255:
I can not reproduce the error. @chmouel :\

vinamra28:
@popcor255 could you please clone https://github.com/tektoncd/catalog/pull/485 and then try reproducing the problem. For now I have renamed the `tests` dir to `samples-1`.

chmouel:
This is weird, maybe there is some race going on, i can def reproduce it every time when running the test manually :

<details>
  <summary>Manual run output: </summary>
<pre>
         name: output
kind: List
metadata:
  resourceVersion: ""
  selfLink: ""
+ echo '--- Container Logs'
--- Container Logs
++ kubectl get pod -o name -n docker-build-0-1
+ for pod in $(kubectl get pod -o name -n ${tns})
+ kubectl logs --all-containers -n docker-build-0-1 pod/affinity-assistant-f5130924a6-0
Error from server (BadRequest): container "affinity-assistant" in pod "affinity-assistant-f5130924a6-0" is waiting to start: ContainerCreating
+ true
+ for pod in $(kubectl get pod -o name -n ${tns})
[...]

PR #483, opened 2020-08-10
[WIP] Fix sidecar not generating certs.

popcor255:
# Changes
The end to end test attaches a sidecar with a image registry for tasks to push into to. This task is optional and is encouraged to be used for testing. However, the function add_sidecar_registry sets the sidecar registry instead of appending it. Allowing the sidecar-registry to run alongside with other tasks will fix this bug. #481

# Submitter Checklist

review them:

- [ ] Complies with [Catalog Orgainization TEP][TEP], see [example]. **Note** [An issue has been filed to automate this validation][validation]

          ```

          spec:
            description: >-

          ```

for more details._

---

### DIFF
--- a/task/jib-maven/0.1/jib-maven.yaml
+++ b/task/jib-maven/0.1/jib-maven.yaml
@@ -46,6 +46,8 @@ spec:
     env:
     - name: HOME
       value: /workspace
+    - name: "DOCKER_CONFIG"
+      value: $(credentials.path)/.docker/
     volumeMounts:
     - name: $(params.CACHE)
       mountPath: /workspace/.m2


ImJasonH:
Thanks for this fix! Can you add some more detail to the PR description about why this is necessary, and what the behavior was before? My understanding of `readinessProbe` is that it would check every second (`periodSeconds`) until the command exited successfully. Is that not the case? What additional behavior does `sleep 10` add?

popcor255:
Yeah, this is not a real fix. Delaying the sleep worked on my local machine. Im just trying to see if it passes the test in this case as well. If so, then the readiness probe isnt working properly. I will update the description as well. :+1:

popcor255:
:facepalm: They updated the latest tag

> `docker:19` and `docker:latest` currently point to the same image.

I have no idea why these tests are passing now.

PR #491, opened 2020-08-12
[Fixes #481] remove add registry sidecar from docker-build tests

### LINKED ISSUE #481: docker-build test task is failing

popcor255:
# Changes

## Overview
* Removed docker-build from skip task
* Removed add_sidecar_registry from tests
* Added a deployment with an image registry instead of sidecar.

## Summary
The end to end test attaches a sidecar with a image registry for tasks to push into to. This task is optional and is encouraged to be used for testing. However, the function add_sidecar_registry sets the sidecar registry instead of appending it. Not using the tasks will fix the test https://github.com/tektoncd/catalog/issues/481

## TLDR;
Script sets sidecar equal to an object instead of appending it. So, Im not going to use it in my test.

### DIFF
--- a/task/docker-build/0.1/docker-build.yaml
+++ b/task/docker-build/0.1/docker-build.yaml
@@ -11,11 +11,9 @@ metadata:
 spec:
   description: >-
     This task will build and push an image using docker.
-
     The task will build an out image out of a Dockerfile.
     This image will be pushed to an image registry.
     The image will be built and pushed using a dind sidecar over TCP+TLS.
-
   params:
   - name: image
     description: Reference of the image docker will produce.
@@ -33,6 +31,9 @@ spec:
     default: ""
   - name: push_extra_args
     description: Extra parameters passed for the push command when pushing images.
+    default: ""
+  - name: insecure_registry
+    description: Allows the user to push to an insecure registry that has been specified
     default: ""
   workspaces:
   - name: source
@@ -86,6 +87,7 @@ spec:
       - --storage-driver=vfs
       - --userland-proxy=false
       - --debug
+      - --insecure-registry=$(params.insecure_registry)
     securityContext:
       privileged: true
     env:
--- a/task/docker-build/0.1/tests/pre-apply-task-hook.sh
+++ b/task/docker-build/0.1/tests/pre-apply-task-hook.sh
@@ -1,8 +1,4 @@
 #!/usr/bin/env bash
 
-# Add an internal registry as sidecar to the task so we can upload it directly
-# from our tests without having to go to an external registry.
-add_sidecar_registry ${TMPF}
-
 # Add git-clone
 kubectl -n ${tns} apply -f ./task/git-clone/0.1/git-clone.yaml
--- a/task/docker-build/0.1/tests/resources.yaml
+++ b/task/docker-build/0.1/tests/resources.yaml
@@ -1,4 +1,3 @@
----
 apiVersion: v1
 kind: PersistentVolumeClaim
 metadata:
@@ -9,3 +8,39 @@ spec:
   resources:
     requests:
       storage: 500Mi
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: registry-docker-build-local
+  namespace: docker-build-0-1
+spec:
+  selector:
+    matchLabels:
+      app: registry-docker-build-local
+  template:
+    metadata:
+      labels:
+        app: registry-docker-build-local
+    spec:
+      containers:
+      - name: registry-docker-build-local
+        image: registry:2
+        resources:
+          limits:
+            memory: "128Mi"
+            cpu: "500m"
+        ports:
+        - containerPort: 5000
+---
+apiVersion: v1
+kind: Service
+metadata:
+  name: registry
+  namespace: docker-build-0-1
+spec:
+  selector:
+    app: registry-docker-build-local
+  ports:
+  - port: 5000
+    targetPort: 5000
--- a/task/docker-build/0.1/tests/run.yaml
+++ b/task/docker-build/0.1/tests/run.yaml
@@ -1,4 +1,3 @@
----
 apiVersion: tekton.dev/v1beta1
 kind: Pipeline
 metadata:
@@ -30,7 +29,9 @@ spec:
       workspace: shared-workspace
     params:
     - name: image
-      value: localhost:5000/nocode
+      value: registry.docker-build-0-1.svc.cluster.local:5000/nocode
+    - name: insecure_registry
+      value: registry.docker-build-0-1.svc.cluster.local:5000
 ---
 apiVersion: tekton.dev/v1beta1
 kind: PipelineRun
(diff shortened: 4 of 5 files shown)


test/e2e-common.sh
```diff
@@ -27,8 +27,7 @@ source $(dirname $0)/../vendor/github.com/tektoncd/plumbing/scripts/e2e-tests.sh
 # from our tests withouth having to go to an external registry.
 function add_sidecar_registry() {
     cp ${1} ${TMPF}.read
-
-    cat ${TMPF}.read | python3 -c 'import yaml;f=open(0, encoding="utf-8"); data=yaml.load(f.read(), Loader=yaml.FullLoader);data["spec"]["sidecars"]=[{"image":"registry", "name": "registry"}];print(yaml.dump(data, default_flow_style=False));' > ${TMPF}
+    cat ${TMPF}.read | python3 -c 'import yaml;f=open(0, encoding="utf-8"); data=yaml.load(f.read(), Loader=yaml.FullLoader);sidecars=data["spec"].get("sidecars"); sidecars = [[], data["spec"].get("sidecars")][sidecars != None]; sidecars.append({"image":"registry", "name": "registry"}); data["spec"]["sidecars"]=sidecars;print(yaml.dump(data, default_flow_style=False));' > ${TMPF}
```
ImJasonH, line 30: This is frankly indecipherable. Can we make this more readable somehow?

chmouel: yes we can.. let's maybe add this fix and work on refactoring this?

popcor255: I 110% percent agree. There is tool already designed to do this kind of substitution called yq. I don't know why there is an arbitrary line of python in the first place. However, I feel like this someone else has had to say something before. This had to be put there for some reason. idk, why tho...

popcor255:
:facepalm: This why its one line...
```bash
++ python3 utils.py add_sidecar_to_yaml_file
python3: can't open file 'utils.py': [Errno 2] No such file or directory
```

chmouel:
This sounds good, thanks for doign this refactoring, any chances if you can squash the commits ? (preferrably if you can separate the refactoring from the fix but that's not absolutely necessary). thanks!

vinamra28:
probably fetching the latest commits from upstream and rebasing the PR would help fix the CI

test/utils.py
```diff
@@ -0,0 +1,32 @@
...
+def add_sidecar_to_yaml_file():
+    # This function is expecting a tekton task from stdin
+    with sys.stdin as file:
+        # Load yaml file passed by stdin
+        payload = yaml.load(file, Loader=yaml.FullLoader)
+        # If sidecars doesnt have any entries create an array for payload manipulation
+        if 'sidecars' not in payload['spec']:
+            payload['spec']['sidecars'] = []
+        # Append the local image registry to the yaml file
+        payload['spec']['sidecars'].append({"image": "registry", "name": "registry"})
```
ImJasonH, line 14: Can we just specify the registry in the original YAML, instead of injecting in this way?

test/e2e-common.sh
```diff
@@ -43,7 +42,7 @@ function install_pipeline_crd() {
   if [[ -n ${RELEASE_YAML} ]];then
 	latestreleaseyaml=${RELEASE_YAML}
   else
-    latestreleaseyaml=$(curl -s https://api.github.com/repos/tektoncd/pipeline/releases|python -c "import sys, json;x=json.load(sys.stdin);ass=x[0]['assets'];print([ x['browser_download_url'] for x in ass if x['name'] == 'release.yaml'][0])")
```
ImJasonH, line 46: Could this just default to https://storage.googleapis.com/tekton-releases/latest/release.yaml, instead of querying GitHub for the latest release?

popcor255: :+1: nice catch

test/utils.py
popcor255: I agree that injecting sidecars is not a good idea. I also think it would be a mistake to change the yaml for the sake of passing a test. The local image registry was only there to help us test `docker push`. Now, honestly its hurting.  I would perfer to push to my docker hub account then add an ephemeral image registry that provides no purpose for the end-user. Also, there are about 13 other tasks that use the add registry sidecar. Ignoring this method in docker-build does not fix the overall problem that this function does perform it's expected behavior. 
```bash
    cat ${TMPF}.read | python3 -c 'import yaml;f=open(0, encoding="utf-8"); data=yaml.load(f.read(), Loader=yaml.FullLoader);data["spec"]["sidecars"]=[{"image":"registry", "name": "registry"}];print(yaml.dump(data, default_flow_style=False));' > ${TMPF}
```
I do think having a namespace with a local image registry would be a good solution. Openshift provides an image registry that can be accessed in its own namespace. I think creating an uninsecured registry in its own namespace that gets cleaned up at the end of testing would be useful. Pods would be able to access it through it's svc. This would also help CI times cause we wouldn't have to spin up 14 image registry pods.

ImJasonH: Oh I see what you're saying, the registry shouldn't be specified in the actual catalog task. That definitely makes sense, I wasn't paying enough attention. 🤦 

+1 to having e2e tests start by running a registry, run all the relevant tasks against it, and then delete the registry. This would let us test the tasks without modifying them, which might also hide breakages.

popcor255: The only thing left that needs to be decided is a standard for naming and tagging images with a shared registry. (I.E we cant have 10 tasks build and push foo:latests to the registry without breaking the test).

popcor255: Since, tasks already have a version naming and tagging it with those labels would work.
example:
<task-name>:<task-version>
localhost:5000/docker-build:0.1

popcor255:
@ImJasonH @afrittoli @chmouel YAY! YAML! :tada: (could you take another look) :wink:

popcor255:
how do i remove hold :crying_cat_face:

task/docker-build/0.1/docker-build.yaml
```diff
@@ -34,6 +34,21 @@ spec:
   - name: push_extra_args
     description: Extra parameters passed for the push command when pushing images.
     default: ""
+  - name: docker_host
```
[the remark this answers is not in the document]
popcor255: Oh, big derp. That makes alot of sense because changing the parameters like I did, required me to change the docs anyway. This way is must better. +1

chmouel:
/lgtm
/approve

Let's see how it goes!

PR #506, opened 2020-09-09
Parallelize calls to copy from/to GCS

coryrc:
# Changes

The -m flag copies from multiple threads and makes a huge
performance difference for large trees.

# Submitter Checklist

review them:

- [ na] Includes [docs](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if user facing)
- [ x] Commit messages follow [commit message best practices](https://github.com/tektoncd/community/blob/master/standards.md#commit-messages)
- [na ] Complies with [Catalog Organization TEP][TEP], see [example]. **Note** [An issue has been filed to automate this validation][validation]
  - [ na] File path follows  `<kind>/<name>/<version>/name.yaml`
  - [ na] Has `README.md` at `<kind>/<name>/<version>/README.md`
  - [ na] Has mandatory `metadata.labels` - `app.kubernetes.io/version` the same as the `<version>` of the resource
  - [ na] Has mandatory `metadata.annotations` `tekton.dev/pipelines.minVersion`
  - [ na] mandatory `spec.description` follows the convention

### DIFF
--- a/task/gcs-download/0.1/gcs-download.yaml
+++ b/task/gcs-download/0.1/gcs-download.yaml
@@ -55,7 +55,7 @@ spec:
 
       if [[ "$(params.typeDir)" == "true" ]]; then
         mkdir -p "$DESTINATION"
-        gsutil rsync -d -r "$(params.location)" "$DESTINATION"
+        gsutil -m rsync -d -r "$(params.location)" "$DESTINATION"
       else
         gsutil cp $(params.location) "$DESTINATION"
       fi
--- a/task/gcs-upload/0.1/gcs-upload.yaml
+++ b/task/gcs-upload/0.1/gcs-upload.yaml
@@ -50,7 +50,7 @@ spec:
       fi
 
       if [[ -d "$SOURCE" ]]; then
-        gsutil rsync -d -r "$SOURCE" "$(params.location)"
+        gsutil -m rsync -d -r "$SOURCE" "$(params.location)"
       else
         gsutil cp "$SOURCE" "$(params.location)"
       fi


task/gcs-download/0.1/gcs-download.yaml
```diff
@@ -55,7 +55,7 @@ spec:
 
       if [[ "$(params.typeDir)" == "true" ]]; then
         mkdir -p "$DESTINATION"
-        gsutil rsync -d -r "$(params.location)" "$DESTINATION"
+        gsutil -m rsync -d -r "$(params.location)" "$DESTINATION"
```
chmouel, line 58: Is there drawback to multiplex the connexions? Should it be optional?

coryrc: Yes-ish; from `gsutil help options`:

Using the -m option may make your performance worse if you
are using a slower network, such as the typical network speeds
offered by non-business home network plans. It can also make
your performance worse for cases that perform all operations
locally (e.g., gsutil rsync, where both source and destination
URLs are on the local disk), because it can "thrash" your local
disk.

The latter case does not apply here. The former case if a possibility, but even on my slow home American "broadband" connection I see improved performance.

chmouel: Sounds good to me! thanks, that doesn't sound that could happen in our case (i.e from a task)

chmouel:
/lgtm

Looking Good! 🤙🏽

vinamra28 (approved):
/lgtm

PR #516, opened 2020-09-23
Add the ability to specify secret name and key

afrittoli:
# Changes

Rather than forcing task users to create a secret with a fixed
name, allow setting the secret name and key as parameter.

The new parameters are optional, in the sense that they default
to the previous hardcoded values.

# Submitter Checklist

review them:

          ```

          spec:
            description: >-

          ```

for more details._

---

### DIFF
--- a/task/github-set-status/0.2/README.md
+++ b/task/github-set-status/0.2/README.md
@@ -1,0 +1,79 @@
+# GitHub Set Status
+
+GitHub Set Status is part of a collection of GitHub tasks to help working
+with the [GitHub API](https://docs.github.com/en/rest/reference).
+
+## GitHub token
+
+This task expects a secret set in the kubernetes secret `github`
+with a GitHub token in the key `token`; you can easily create it on the
+command line with `kubectl` like this :
+
+```
+kubectl create secret generic github --from-literal token="MY_TOKEN"
+```
+
+## Set Status on a Commit/PR
+
+The `github-set-status` task uses the [status api](https://docs.github.com/en/rest/reference/repos#statuses)
+to mark GitHub commits with an `error`, `failure`, `pending`, or `success`
+state, which is then reflected in pull requests involving those commits.
+
+Statuses include as well a `description` and a `target_url`, to give the user
+informations about the CI statuses or a direct link to the full log.
+
+### Install the Task
+
+```
+kubectl apply -f https://raw.githubusercontent.com/tektoncd/catalog/master/task/github-set-status/0.2/github-set-status.yaml
+```
+
+### Parameters
+
+* **REPO_FULL_NAME**: The GitHub repository full name, _e.g:_ `tektoncd/catalog`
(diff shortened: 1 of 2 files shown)


afrittoli:
I created this as a `v2` because of the change of interface. Note that `v2` is backward compatible with `v1` so perhaps the change is not needed, or it should be a `v1.1`?

task/github-set-status/0.2/README.md
```diff
@@ -0,0 +1,79 @@
...
+to mark GitHub commits with an `error`, `failure`, `pending`, or `success`
+state, which is then reflected in pull requests involving those commits.
+
+Statuses include as well a `description` and a `target_url`, to give the user
+informations about the CI statuses or a direct link to the full log.
+
+### Install the Task
+
+```
+kubectl apply -f https://raw.githubusercontent.com/tektoncd/catalog/master/task/github-set-status/0.1/github-set-status.yaml
```
vinamra28, line 28:
```suggestion
kubectl apply -f https://raw.githubusercontent.com/tektoncd/catalog/master/task/github-set-status/0.2/github-set-status.yaml
```

task/github-set-status/0.2/github-set-status.yaml
```diff
@@ -0,0 +1,127 @@
...
+  steps:
+    - name: set-status
+      env:
+        - name: GITHUBTOKEN
+          valueFrom:
+            secretKeyRef:
+              name: $(params.GITHUB_TOKEN_SECRET_NAME)
+              key: $(params.GITHUB_TOKEN_SECRET_KEY)
+
+      image: registry.access.redhat.com/ubi8/ubi:latest
```
vinamra28, line 92: can we avoid using latest tag over here :sweat_smile:

vinamra28 (commented):
Just a few nitpick :sweat_smile:

vinamra28:
> I created this as a `v2` because of the change of interface. Note that `v2` is backward compatible with `v1` so perhaps the change is not needed, or it should be a `v1.1`?
> 
> If this is accepted, I would update the other github tasks to match.

@afrittoli `v2` would be fine I guess since there is an enhancement in task by specifying the secrets

task/github-set-status/0.2/README.md
afrittoli: Good catch, thank you!

task/github-set-status/0.2/github-set-status.yaml
afrittoli: Yeah... this is the same as the original task, but I agree we should not use latest.
I'll use ubi8/python-38:1-34.1599745032 instead

vinamra28 (approved):
/lgtm

Divyansh42 (approved):
/lgtm

chmouel:
/lgtm
/approve

I think it would be nice if we can have an issue created to track down the updates of the other github task

PR #517, opened 2020-09-23
gcs-generic: Make credential file optional.

wlynch:
# Changes

When running on GCP, gsutil will try to infer identity from the GKE
metadata service instead of an explicitly defined credential file. The
easiest way to support this is to only set the credentials iff the file
is provided, otherwise just fallback to gsutils default behavior.

Also adds a simpler example that doesn't require credentials.

Provides working example for issue described in https://github.com/tektoncd/pipeline/issues/3208.

/cc @dansiviter (confirmed this works for a private GCS file w/ a Workload Identity enabled cluster)

# Submitter Checklist

review them:

  - [x]  Has `README.md` at `<kind>/<name>/<version>/README.md`

          ```

          spec:
            description: >-

          ```

for more details._

---

### DIFF
--- a/task/gcs-generic/0.1/gcs-generic.yaml
+++ b/task/gcs-generic/0.1/gcs-generic.yaml
@@ -26,11 +26,11 @@ spec:
    description: |
       The path inside the credentials workspace to the GOOGLE_APPLICATION_CREDENTIALS key file.
    type: string
-   default: service_account.json
+   default: ''
  - name: image
    description: google cloud image that will be used in steps.
    type: string
-   default: google/cloud-sdk
+   default: gcr.io/google.com/cloudsdktool/cloud-sdk:slim
  steps:
  - name: credential-setup
    image: "$(params.image)"
--- a/task/gcs-generic/0.1/samples/ls-taskrun.yaml
+++ b/task/gcs-generic/0.1/samples/ls-taskrun.yaml
@@ -1,0 +1,15 @@
+apiVersion: tekton.dev/v1beta1
+kind: TaskRun
+metadata:
+  generateName: gcs-ls-
+spec:
+  workspaces:
+    - name: credentials
+      emptyDir: {}
+  params:
+    - name: "command"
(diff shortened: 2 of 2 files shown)


task/gcs-generic/0.1/gcs-generic.yaml
```diff
@@ -26,11 +26,11 @@ spec:
    description: |
       The path inside the credentials workspace to the GOOGLE_APPLICATION_CREDENTIALS key file.
    type: string
-   default: service_account.json
+   default: ''
  - name: image
    description: google cloud image that will be used in steps.
    type: string
-   default: google/cloud-sdk
+   default: gcr.io/google.com/cloudsdktool/cloud-sdk
```
vinamra28, line 33: can you please also add a tag or a digest for the image? :slightly_smiling_face:

vinamra28 (changes requested):
just one change required otherwise looks good

task/gcs-generic/0.1/gcs-generic.yaml
wlynch: This is intentional - this pulls in the latest version by default, and users can pin to specific versions by setting the image param.

Set the tag to `slim`, but this is to pull in a minimized version, not for any version pinning.

vinamra28: my reason for specifying the tag was that in future we are planning to add catlin(catalog linter) to catalog CI which has a requirement that the image should be tagged

vinamra28 (commented):
/lgtm

task/gcs-generic/0.1/gcs-generic.yaml
chmouel: I think from a security point of view we don't want to have by default pulling the `latest`, isnt it ? cc @dlorenc

chmouel:
/lgtm
/approve

Thanks 👍🏼🔥

PR #519, opened 2020-09-25
Fix Catlin error by adding Digest to the Images

vinamra28:
# Changes

- Previously catalog's task has some images that are not
  tagged properly,this patch fixes these images by adding
  the digest
- Soon once Catlin is added to the CI it will throw error
  for those images which don't have proper tags

co-authored by:- @PuneetPunamiya

# Submitter Checklist

review them:

          ```

          spec:
            description: >-

          ```

for more details._

---


Divyansh42 (approved):
/lgtm
Thanks :+1:

Divyansh42 (approved):
/lgtm

vdemeester (commented):
@vinamra28 one question, the image that were previously using `:latest` (or not tag at all), how did we get the digest ? 
Asking that because if the digest used is the *current* master, we're gonna end up with a inexistent image at some point. The registries are usually garbage collecting any image that is not tagged ; and as `:latest` is a moving target, a digest point to a `:latest` *might* end up pointing to a non-tagged image and thus might refer a garbage-collected image.
/hold

vinamra28:
> @vinamra28 one question, the image that were previously using `:latest` (or not tag at all), how did we get the digest ?
> Asking that because if the digest used is the _current_ master, we're gonna end up with a inexistent image at some point. The registries are usually garbage collecting any image that is not tagged ; and as `:latest` is a moving target, a digest point to a `:latest` _might_ end up pointing to a non-tagged image and thus might refer a garbage-collected image.

I inspected the registries and chose the digest of the non-latest tag.

vdemeester:
> @vdemeester I inspected the registries and chose the digest of the non-latest tag.

Should we add a comment on which `tag` it represent ? (or even better, use `:tag@digest` notation ? although might not work on all container runtime :thinking:)

vinamra28:
> > @vdemeester I inspected the registries and chose the digest of the non-latest tag.
> 
> Should we add a comment on which `tag` it represent ? (or even better, use `:tag@digest` notation ? although might not work on all container runtime )

I would prefer adding a comment which `tag` it represent or instead of using the digest shall I switch back to the tags?

vdemeester:
> I would prefer adding a comment which `tag` it represent or instead of using the digest shall I switch back to the tags?

- `digest` is immutable, but it's hard to know the "version"
- `tag` allow to see the "version", but it is mutable
- `tag`+`digest` bring the best of both world :upside_down_face: 

So, digest is fine (better than tag only), tag+digest is also fine and might be better, to *convey* the version of the image used (with digest, it's hard to know which version it is, if it got updated, …). I am fine with either the tag+digest notation or a comment with the tag name. In an ideal world, we could automate the update of a digest (with the comment that come with it).

vinamra28:
> > I would prefer adding a comment which `tag` it represent or instead of using the digest shall I switch back to the tags?
> 
> * `digest` is immutable, but it's hard to know the "version"
> * `tag` allow to see the "version", but it is mutable
> * `tag`+`digest` bring the best of both world upside_down_face
> 
> So, digest is fine (better than tag only), tag+digest is also fine and might be better, to _convey_ the version of the image used (with digest, it's hard to know which version it is, if it got updated, …). I am fine with either the tag+digest notation or a comment with the tag name. In an ideal world, we could automate the update of a digest (with the comment that come with it).

`tag`+`digest` would make sense. Will update the PR in that case :slightly_smiling_face:

task/buildpacks-phases/0.1/buildpacks-phases.yaml
```diff
@@ -59,7 +59,7 @@ spec:
   steps:
     # Ensure the builder's user has read/write permissions for needed directories.
     - name: prepare
-      image: alpine
+      image: docker.io/library/alpine:3.12
```
PuneetPunamiya, line 62: @vdemeester Since this image is available for various os architecture and each os architecture we have different digest, any particular architecture we need to choose or the default one i.e `amd64` and also is it fine to use this particular tag i.e `3.12`

task/blue-green-deploy/0.1/blue-green-deploy.yaml
```diff
@@ -29,7 +29,7 @@ spec:
       description: The newer version of the deployment to be deployed in another zone
   steps:
     - name: change-deployment
-      image: quay.io/openshift/origin-cli:latest
+      image: quay.io/openshift/origin-cli:4.4
```
PuneetPunamiya, line 32: @chmouel  @vdemeester  Using this particular tag and not the digest because with quay if a tag is updated the previous digest becomes invalid ?

So are we fine with using this specific tag ?

task/gcloud/0.1/gcloud.yaml
```diff
@@ -14,7 +14,7 @@ spec:
   params:
   - name: gcloud-image
     description: gcloud CLI container image to run this task
-    default: google/cloud-sdk:slim
+    default: gcr.io/google.com/cloudsdktool/cloud-sdk:slim@sha256:70d3c9ef711c704259b06485f9ab3bab6e1b5c99e4a5c1ed37f9338004664c17 #tag: slim, 312.0.0-slim
```
PuneetPunamiya, line 17: @vdemeester  @chmouel  Are we fine with image registery change?

task/blue-green-deploy/0.1/blue-green-deploy.yaml
chmouel: yes that is fine

task/gcloud/0.1/gcloud.yaml
chmouel: all good yes,

task/buildpacks-phases/0.1/buildpacks-phases.yaml
vdemeester: @PuneetPunamiya the manifest digest should be the same, so you "should" have only one digest. In any case, it's fine using the tag for this :wink:

task/gcloud/0.1/gcloud.yaml
vdemeester: yes :wink:

task/ansible-tower-cli/0.1/ansible-tower-cli.yaml
```diff
@@ -38,7 +38,7 @@ spec:
       default: "tower-creds"
   steps:
     - name: config
-      image: quay.io/rcmendes/ansible-tower-cli:latest
+      image: quay.io/rcmendes/ansible-tower-cli@sha256:3a61778f410526db8d6e02e87715d58ee770c4a4faf57ac408cb5ec1a025ef2c #tag: latest
```
sthaha, line 41: would updating the `latest` tag (when the author pushes a new image) overwrite this `sha256` and cause the task to fail?

PuneetPunamiya: Yes the task would fail, but for now we don't have any other tag available for it on which we can rely on

vdemeester: So, it will not "fail" per-say — the digest will point to a non tagged image which still works. The main problem with that is, we do not have any guarantee that the upstream registry won't garbage collect the image (docker hub does garbage collect those untagged image at some point).

vinamra28:
can we get this in if the PR looks good to you :sweat_smile: 
/cc @vdemeester @chmouel @sbwsg

sthaha (commented):
/lgtm

vdemeester (approved):
Because we use the tag in the reference, we don't need the comment, but this can be handled later on :wink:

so0k:
for some of these, the tag was pointing to latest of a month ago, how does is this kept up to date?

vinamra28:
hi @so0k, for now we are not keeping those images up-to-date. But probably in future we can add an automation to update the images digests

PR #534, opened 2020-10-06
Rework the triggers-jenkins-job

chmouel:
# Changes

Rewrite the triggers-jenkins-job, the arguments and params are about to same.

Starting from Jenkins LTS 2.176.2 we need to create a crumb and capture the
cookie jar to make the subsequent requests.

https://support.cloudbees.com/hc/en-us/articles/219257077-CSRF-Protection-Explained

Add tests :

We spins up a deployment with a jenkins lts and expose a service to it so our
task can access it.

We do some ninja execing into the pods to create a jenkins job in there and
setup the configmap secret.

We probably should drop the CRUMB setting since it will not work anymore but it
is harmless since we are regenerating it anyway.

Todo: need to test with jenkins image that is older than 2.176.2, but it
probably should work

# Submitter Checklist

### DIFF
--- a/task/trigger-jenkins-job/0.1/tests/pre-apply-task-hook.sh
+++ b/task/trigger-jenkins-job/0.1/tests/pre-apply-task-hook.sh
@@ -1,0 +1,69 @@
+#!/bin/bash
+# This will create a Jenkins deployment, a jenkins service, grab the password,
+# gets a 'crumb' and generate a secret out of it for our task to run against to.
+
+cat <<EOF | kubectl apply -f- -n "${tns}"
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: jenkins
+spec:
+  selector:
+    matchLabels:
+      run: jenkins
+  replicas: 1
+  template:
+    metadata:
+      labels:
+        run: jenkins
+    spec:
+      containers:
+      - name: jenkins
+        image: jenkins/jenkins:lts
+        ports:
+        - containerPort: 8080
+        volumeMounts:
+          - name: jenkins-home
+            mountPath: /var/jenkins_home
+      volumes:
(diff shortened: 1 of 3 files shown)


task/trigger-jenkins-job/0.1/tests/pre-apply-task-hook.sh
```diff
@@ -0,0 +1,63 @@
...
+  template:
+    metadata:
+      labels:
+        run: jenkins
+    spec:
+      containers:
+      - name: jenkins
+        image: jenkins/jenkins:lts
+        ports:
+        - containerPort: 8080
```
vinamra28, line 25: CI is failing on OpenShift cluster. Probable fix would be adding a `volumeMount` in `containerSpec`
```suggestion
      containers:
      - name: jenkins
        image: jenkins/jenkins:lts
        ports:
        - containerPort: 8080
        volumeMounts:
          - name: jenkins-home
            mountPath: /var/jenkins_home
      volumes:
        - name: jenkins-home
          emptyDir: {}
```

vinamra28 (commented):
Thanks for fixing the script and adding the tests for it :+1: . The overall PR looks good :star_struck: just one concern is that the test is failing on OpenShift Cluster

task/trigger-jenkins-job/0.1/tests/pre-apply-task-hook.sh
chmouel: Nicely found! thanks!

vinamra28 (commented):
tested with older version of Jenkins ie `jenkins/jenkins:2.176.1` and working all good. Thanks :+1: 
/lgtm

vdemeester (approved):
/meow

PR #535, opened 2020-10-06
Add a test for the kubeconfig-creator task

chmouel:
# Changes

Add a test for the kubeconfig-creator task

### DIFF
--- a/task/kubeconfig-creator/0.1/tests/run.yaml
+++ b/task/kubeconfig-creator/0.1/tests/run.yaml
@@ -1,0 +1,27 @@
+apiVersion: tekton.dev/v1beta1
+kind: PipelineRun
+metadata:
+  name: kubeconfig-creator-creator
+spec:
+  pipelineSpec:
+    workspaces:
+      - name: output
+    tasks:
+      - name: kubeconfig-creator-test
+        taskRef:
+          name: kubeconfig-creator
+        workspaces:
+          - name: output
+            workspace: output
+        params:
+          - name: name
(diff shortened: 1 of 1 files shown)


piyush-garg:
Test is failing

```sh
{"level":"fatal","ts":1602013552.0897691,"logger":"fallback-logger","caller":"kubeconfigwriter/main.go:115","msg":"Error writing kubeconfig to file: open /workspace/output/kubeconfig: permission denied","commit":"4e23d50","stacktrace":"main.createKubeconfigFile\n\tgithub.com/tektoncd/pipeline/cmd/kubeconfigwriter/main.go:115\nmain.main\n\tgithub.com/tektoncd/pipeline/cmd/kubeconfigwriter/main.go:52\nruntime.main\n\truntime/proc.go:203"}
```

chmouel:
yep it's weird, it doesnt fail when running it locally... not sure why the volume is not writable here..

Divyansh42 (commented):
Can we modify the kubeconfig-creator task? as after the change task will not produce [this](https://github.com/tektoncd/catalog/pull/535#issuecomment-704720088) error.

May be we can add something like this to resolve write permission issue
```
securityContext:
        runAsUser: 0
```

Then probably we can write some better tests, because presently we are not testing much things :sweat_smile:

Divyansh42 (approved):
I think running the task as root user will not be a good approach.
So this PR is fine for me.

piyush-garg (approved):
/lgtm

PR #536, opened 2020-10-07
Add add_task helper function for tests

chmouel:
We were previously using the outdated git_clone version 0.1 in the
add_git_clone_task function.

So let's introduce a more generic function :

```
add_task ${task} ${version}
```

if version is 'latest' it will always install the latest version of the task.

Change all pre-apply-task-hook to use that function instead.

### DIFF
--- a/test/e2e-common.sh
+++ b/test/e2e-common.sh
@@ -32,9 +32,20 @@ function add_sidecar_registry() {
     rm -f ${TMPF}.read
 }
 
-# Add the git_clone task
-function add_git_clone_task() {
-    kubectl -n ${tns} apply -f ./task/git-clone/0.1/git-clone.yaml
+function add_task() {
+    local array path_version task
+    task=${1}
+    if [[ "${2}" == "latest" ]];then
+        array=($(echo task/${task}/*|sort -u))
+        path_version=${array[-1]}
+	else
+		path_version=task/${task}/${2}
+        if [[ ! -d ${path_version} ]];then
+            echo "I could not find version '${2}' for the task '${task}' in ./task/"
+            exit 1
+        fi
+	fi
+    kubectl -n "${tns}" apply -f "${path_version}"/"${task}".yaml
 }
 
 function install_pipeline_crd() {
@@ -93,7 +104,7 @@ function show_failure() {
 
 }
 function test_task_creation() {
-    for runtest in ${@};do
+    for runtest in "${@}";do
         # remove task/ from beginning
         local runtestdir=${runtest#*/}
         # remove /0.1/tests from end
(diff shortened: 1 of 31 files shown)


test/e2e-common.sh
```diff
@@ -32,9 +32,15 @@ function add_sidecar_registry() {
...
 
-# Add the git_clone task
-function add_git_clone_task() {
-    kubectl -n ${tns} apply -f ./task/git-clone/0.1/git-clone.yaml
+function add_task() {
+    local array version task
+    task=${1}
+    version=${2}
+    [[ "${version}" == "latest" ]] && {
+        array=($(echo task/${task}/*|sort -u))
```
piyush-garg, line 40: should we handle the scenario of wrong task name

test/e2e-common.sh
```diff
@@ -32,9 +32,15 @@ function add_sidecar_registry() {
...
+        version=${array[-1]}
```
piyush-garg, line 41: nit: can we have a better name for this, as it will be a path like `task/git-clone/0.2` it avoids confusion in the next apply step

chmouel: better name of what ? the name of the function?

test/e2e-common.sh
chmouel: why not, sure but it will fail anyway since we run under -e :-)

test/e2e-common.sh
piyush-garg: name of the var `version`, it confused me, maybe it's only me :smile: 

Because in the next step we do `kubectl -n "${tns}" apply -f "${version}"/"${task}".yaml`

chmouel: ah yeah you want to be named path_version or something... sure i'll be happy to rename so you are not confused if you reread this code :-p

chmouel: there is actually a bug... :)

test/e2e-common.sh
```diff
@@ -32,9 +32,19 @@ function add_sidecar_registry() {
     rm -f ${TMPF}.read
 }
 
-# Add the git_clone task
-function add_git_clone_task() {
-    kubectl -n ${tns} apply -f ./task/git-clone/0.1/git-clone.yaml
+function add_task() {
```
chmouel, line 35: @piyush-garg updated with your suggestions! thanks

vdemeester (approved):
nice :+1:

vinamra28:
@chmouel just one more task needs a fix and this PR is good to go
https://github.com/tektoncd/catalog/blob/e38be678d1ec494d57316ea949b5bf92d0e01b9d/task/mypy-lint/0.1/tests/pre-apply-task-hook.sh#L4

vinamra28 (approved):
/lgtm

piyush-garg (approved):
/lgtm

PR #546, opened 2020-10-14
Improve e2e tests runner output

chmouel:
We temporary disable the debug output while on the loop so we don't have the
gazillions debug messages while digging into the log. We have added the 'timing'
when it start and ends so we can easily spots the very slow tests (see #548 for investigation)

Aditionally we have introduced the new variable `CATALOG_TEST_SKIP_CLEANUP` if
set it will skip deleting the namespace. This is useful when running in
conjuncton with the `./test/run-test.sh` script to debug when there is a
failure.

### DIFF
--- a/test/e2e-common.sh
+++ b/test/e2e-common.sh
@@ -163,6 +163,10 @@ EOF
         local cnt=0
         local all_status=''
         local reason=''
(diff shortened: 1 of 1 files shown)


vdemeester (approved):
/lgtm

Issue #555, opened 2020-10-23
Ability to specify secret name and key in github tasks


Divyansh42:
# Expected Behavior
Rather than forcing task users to create a secret with a fixed
name, allow setting the secret name and key as parameter.

The new parameters will be optional, in the sense that they default
to the previous hard-coded values.

Check [this](https://github.com/tektoncd/catalog/blob/master/task/github-set-status/0.2/github-set-status.yaml#L88-L90) for the same.
# Actual Behavior
Presently in few of the tasks we force the user to create a secret with a fixed name.
```
env:
    - name: GITHUBTOKEN
      valueFrom:
        secretKeyRef:
          name: github
          key: token
```

Issue #452, opened 2020-07-28
Document Minimum Permissions Needed for GitHub Access Tokens for GitHub Tasks


danielhelfand:
As part of using Tasks around interacting with the GitHub API, it's common to need to create a personal access token to send as an `Authorization` header as part of a request.

An example of this is the [github-add-comment](https://github.com/tektoncd/catalog/tree/master/task/github-add-comment/0.1#github) Task. Showing something along the lines of access token permissions for working with a public repository/private repository, such as below, would make using these Tasks much simpler:

![image](https://user-images.githubusercontent.com/34258252/88723446-7f043f00-d0ee-11ea-9325-93fd5f87b7bd.png)

While there is no perfect way to document exact permissions, there should be more information available in the README, and some simple examples of personal access tokens for all GitHub Tasks.

PR #559, opened 2020-10-27
Modify GitHub tasks to add ability to specify secret name and key

Divyansh42:
# Changes

Rather than forcing task users to create a secret with a fixed
name, allow setting the secret name and key as parameter.

The new parameters are optional, in the sense that they default
to the previous hardcoded values.

Fixes : https://github.com/tektoncd/catalog/issues/555

# Submitter Checklist

review them:

          ```

          spec:
            description: >-

          ```

for more details._

---

### DIFF
--- a/task/github-add-comment/0.2/README.md
+++ b/task/github-add-comment/0.2/README.md
@@ -3,25 +3,17 @@
 The `github-add-comment` task let you add a comment to a pull request or an
 issue.
 
-### Install the Task
+## Install the Task
 
 ```
-kubectl apply -f https://raw.githubusercontent.com/tektoncd/catalog/master/task/github-add-comment/0.1/github-add-comment.yaml
+kubectl apply -f https://raw.githubusercontent.com/tektoncd/catalog/master/task/github-add-comment/0.2/github-add-comment.yaml
 ```
-
-### Parameters
-
-* **GITHUB_HOST_URL:**: The GitHub host domain (_default:_ `api.github.com`)
-* **API_PATH_PREFIX:**: The GitHub Enterprise has a prefix for the API path. _e.g:_ `/api/v3`
-* **REQUEST_URL:**: The GitHub pull request or issue url, _e.g:_
-  `https://github.com/tektoncd/catalog/issues/46`
-* **COMMENT:**: The actual comment to add _e.g:_ `don't forget to eat your vegetables before commiting.`.
 
 ## Secrets
 
-This tasks would expect to have a secret set in the kubernetes secret `github`
-with a GitHub token in the key `token`, you can easily create it on the
-command line with `kubectl` like this :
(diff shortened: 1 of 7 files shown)


vinamra28 (commented):
can we try resolving the issue https://github.com/tektoncd/catalog/issues/452 over here?

task/github-close-issue/0.1/README.md
```diff
@@ -1,30 +1,29 @@
-# GitHub
+# Close an issue or a pull request
 
-A collection of tasks to help working with the [GitHub
-API](https://developer.github.com/v3/).
+The `github-close-issue` task let you close a pull request or an
```
Divyansh42, line 3: @chmouel can we change the name of the task, as this task can close an issue or a pull request.
But from the task name`github-close-issue` it seems we can only close issues.

chmouel: I am pretty sure that API endpoint work for issues and pull request.. Since I have used it like that  previously.

Divyansh42: Yeah it works fine for both issue and pull request, I just mean to say may be we can only change the name of the task so that user can get to know that this task can be used to close issues as well as pull request.

Presently if user reads the name of the task, it would appear to him this task is only to close issue not pull request.

chmouel: Yes but what do you want to call it ?

vinamra28: I guess changing the name of the task would be a breaking change and instead of bumping the new version of the task we'll have to add a new task instead which I guess would not be appropriate as there would be two tasks in catalog performing the same work just the name is different

Divyansh42: Agreed with @vinamra28 , I think now it's not good to rename the task.

chmouel: IMO, one task to one directoy is a bit too cumbersome, if we were going wild we could end up with hundred of github- tasks to implement just one simple call.

Why is is that the hub make it a requirement ?

Could we not have github/close-pull-request/0.1 task/github/close-issue/0.1 etc.. ?

sthaha: @chmouel the current constraint `<kind>/<name>/<version>/<name>.yaml` prevents name collision. since a system/tool can validate that the content of the `name.yaml` has 

```yaml

kind: <kind>
metadata:  
  name: <name>
  labels: 
    app.kubernetes.io/version: <version>
```

In your proposal 
>  Could we not have github/close-pull-request/0.1 task/github/close-issue/0.1 etc.. ?

Any thoughts how you would avoid a name collision and enforce these constraints ? 
* no name collision
* has a version
* The path of resource is computable . i.e we can compute that path `<kind>/<name>/<version>/<name>.yaml` from the metadata. E.g. a tool can compute the location of a `task` named `git-cli` that has a version `0.1`

chmouel:
/lgtm

Looking Good! 🤙🏽

vinamra28 (approved):
/lgtm

vdemeester (approved):

PR #569, opened 2020-10-29
 Detect propelry changed files when there is multiple yaml in there

chmouel:
Do not quote "$@" too (as shellcheck was telling me) because we would not get a
proper array to get into.

Detect propelry changed files when there is multiple yaml in there

/hold


chmouel:
/hold cancel

ready to review/merge,

vdemeester (approved):

PR #593, opened 2020-11-26
buildah: add a new 0.3 version removing privileged

vdemeester:
# Changes

Using the `vfs` storage driver in a container is similar to using
`overlay` but has the benefit to not require the `privileged`
securityContext.

# Submitter Checklist

review them:

          ```

          spec:
            description: >-

          ```

for more details._

---

### DIFF
--- a/task/buildah/0.3/README.md
+++ b/task/buildah/0.3/README.md
@@ -1,0 +1,63 @@
+# Buildah
+
+This Task builds source into a container image using Project Atomic's
+[Buildah](https://github.com/projectatomic/buildah) build tool. It uses
+Buildah's support for building from
+[`Dockerfile`](https://docs.docker.com/engine/reference/builder/)s, using its
+`buildah bud` command. This command executes the directives in the `Dockerfile`
+to assemble a container image, then pushes that image to a container registry.
+
+## Install the Task
+
+```
+kubectl apply -f https://raw.githubusercontent.com/tektoncd/catalog/master/task/buildah/0.3/buildah.yaml
+```
+
+## Parameters
+
+* **IMAGE**: The name (reference) of the image to build.
+* **BUILDER_IMAGE:**: The name of the image containing the Buildah tool. See
+  note below.  (_default:_ quay.io/buildah/stable:v1.15.1)
+* **DOCKERFILE**: The path to the `Dockerfile` to execute (_default:_
+  `./Dockerfile`)
+* **CONTEXT**: Path to the directory to use as context (_default:_
+  `.`)
+* **TLSVERIFY**: Verify the TLS on the registry endpoint (for push/pull to a
+  non-TLS registry) (_default:_ `true`)
+* **FORMAT**: The format of the built container, oci or docker (_default:_
+ `oci`)
+* **BUILD_EXTRA_ARGS**: Extra parameters passed for the build command when
+  building images. (_default:_ `""`)
+* **PUSH_EXTRA_ARGS**: Extra parameters passed for the push command when
+  pushing images. (_default:_ `""`)
+* **STORAGE_DRIVER**: Set build storage driver to use (_default:_ `"vfs"`_)
+
+## Workspaces
+
+* **source**: A [Workspace](https://github.com/tektoncd/pipeline/blob/master/docs/workspaces.md) containing the source to build.
+* **sslcertdir**: An [*optional* Workspace](https://github.com/tektoncd/pipeline/blob/v0.17.0/docs/workspaces.md#optional-workspaces) containing your custom SSL certificates to connect to the registry. Buildah will look for files ending with *.crt, *.cert, *.key into this workspace. See [this sample](./samples/openshift-internal-registry.yaml) for a complete example on how to use it with OpenShift internal registry.
+
+## Usage
+
+This TaskRun runs the Task to fetch a Git repo, and build and push a container
+image using Buildah.
+
+```yaml
+apiVersion: tekton.dev/v1beta1
+kind: TaskRun
+metadata:
+  name: buildah-build-my-repo
+spec:
+  taskRef:
+    name: buildah
+  params:
+  - name: IMAGE
+    value: gcr.io/my-repo/my-image
+  workspaces:
+  - name: source
+    persistentVolumeClaim:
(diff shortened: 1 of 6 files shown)


task/buildah/0.3/README.md
```diff
@@ -0,0 +1,63 @@
...
+  `.`)
+* **TLSVERIFY**: Verify the TLS on the registry endpoint (for push/pull to a
+  non-TLS registry) (_default:_ `true`)
+* **FORMAT**: The format of the built container, oci or docker (_default:_
+ `oci`)
+* **BUILD_EXTRA_ARGS**: Extra parameters passed for the build command when
+  building images. (_default:_ `""`)
+* **PUSH_EXTRA_ARGS**: Extra parameters passed for the push command when
+  pushing images. (_default:_ `""`)
+* **STORAGE_DRIVER**: Set build storage driver to use (_default:_ `"vfs"`_)
```
vdemeester, line 33: I wonder if we should have this parameter at all (and if we should document it or not)

vdemeester:
:thinking: 
/hold

task/buildah/0.3/README.md
chmouel: Isnt there some drawbacks using vfs ? Being much slower? (maybe i am confused tho which option is the really slow one here)

The user may want to adjust it depend of their build workload and not being bothered about running it as Root

vdemeester: > Isnt there some drawbacks using vfs ? Being much slower? (maybe i am confused tho which option is the really slow one here)

Well, usually there is, but in a container in a cluster, the difference is very very very small (cc @gabemontero). It is small enough that we cane consider using `vfs` by default, and thus allowing us to not required `privileged`.

chmouel: I mean the drawback to don't give the choise the user with a  params.

vdemeester: Ah :sweat_smile: well keeping it means it might or might not work (if the user provides `overlay` for `STORAGE_DRIVER` without being able to run it as `privileged` it will fail)

gabemontero: there have also been enhancements in buildah wrt vfs since this example was first constructed

some recent experimentation on our part in OCP showed the difference now was not discernable.

IIRC I believe historically this example was using 1.11

but I see for example @vdemeester 's update moves to `quay.io/buildah/stable:v1.15.1`

also, IIRC, @nalind from the buildah team suggested 1.16 for even more improvements ... I would consider that change as well @vdemeester 

And moving forward, I think we should engage with him directly on these types of questions, which I've attempted to do here of course :-) ... please consider this the "formal introduction" @vdemeester @chmouel .... I'll mention this to him in the next meeting I'm in with him as well.

nalind: 1.16 should handle ADD and COPY better (faster) than it used to, and we've made correctness fixes as we've gone along.

Keep in mind that most of the gains we're seeing with vfs have happened because it tries to take advantage of the FICLONE and FICLONERANGE ioctls when copying the contents of one layer to create the basis for another layer, to which it then applies a layer diff to finish initializing the new layer's contents.  That ioctl() call will only succeed if the underlying filesystem supports it.  For OpenShift nodes, we can generally expect it to be XFS formatted with "-m reflink=1", which provides that ability, but I can't make guarantees about clusters in general.

gabemontero: Ah .... very informative / useful @nalind  - thanks

so @vdemeester @chmouel @sbose78 @siamaksade .... vs. how we went with downstream openshift pipelines, perhaps we want to consider exposing the "performance vs. security" nuances around buildah in our upstream examples (as the performance impact is possibly more likely on non-openshift nodes)?

@sbose78 I know this was on your radar in particular recently ... thoughts?

chmouel: thanks @nalind and @gabemontero, yeah maybe a documentation update mentionning nalind comment would be great for our users,

sbose78: I'm good as long as the default is vfs. If an advanced user wishes to flip that to "overlay", I would expect the advanced user to read documentation where the nuances (incl need for higher privileges) would be explained. :)

chmouel:
/lgtm

diff for people who needs it https://paste.gg/p/anonymous/e355c7d360f547ef9b5b309dfc46c8b6

chmouel:
test error seems legit : 

```
STEP 2: COMMIT registry:5000/nocode
level=error msg="Error while applying layer: ApplyLayer exit status 1 stdout:  stderr: remount /, flags: 0x44000: permission denied"
error committing base container: error copying layers and metadata for container "191af5fd2f25b48cfb087f36e88a01877df816ab4998ef3eed29b6018bd22f9e": Error committing the finished image: error adding layer with blob "sha256:5f70bf18a086007016e948b04aed3b82103a36bea41755b6cddfaf10ace3c6ef": ApplyLayer exit status 1 stdout:  stderr: remount /, flags: 0x44000: permission denied
```

/hold

chmouel:
@vinamra28 no need to serial test, i think the CI error is legit error... not a fluke..

vinamra28:
> @vinamra28 no need to serial test, i think the CI error is a legit error... not a fluke..

@chmouel I was just running the catlin CI :sweat_smile: 
`/test .*` runs only the catlin job

chmouel:
Ah cool cool thanks 👍

chmouel:
Related : 

https://github.com/containers/buildah/issues/2554#issuecomment-790124388

mdraijer:
Is there any progress on this issue? It is quite a problem running privileged containers just because Buildah needs them.
A solution was proposed a long time ago. This might be driving people towards other tools, such as Kaniko, that do not have this issue.

vdemeester:
> Is there any progress on this issue? It is quite a problem running privileged containers just because Buildah needs them.
> A solution was proposed a long time ago. This might be driving people towards other tools, such as Kaniko, that do not have this issue.

Updated, but as for kaniko, this still needs to run as root, so I've added the `securityContext` for running as root to be explicit.

vdemeester:
/close
Someone will need to carry this I think.

PR #595, opened 2020-11-30
Change runAsUser and Group to parameter

zhangtbj:
# Changes

The `runAsUser` and `runAsGroup` are still using 1000 as a hardcode value.
They should also be updated to use the parameters: `params.USER_ID` and `params.GROUP_ID`

# Submitter Checklist

review them:

          ```

          spec:
            description: >-

          ```

for more details._

---

### DIFF
--- a/task/buildpacks/0.2/buildpacks.yaml
+++ b/task/buildpacks/0.2/buildpacks.yaml
@@ -107,8 +107,8 @@ spec:
         - name: $(params.PLATFORM_DIR)
           mountPath: /platform
       securityContext:
-        runAsUser: 1000
-        runAsGroup: 1000
+        runAsUser: $(params.USER_ID)
+        runAsGroup: $(params.GROUP_ID)
 
   volumes:
     - name: empty-dir


task/buildpacks/0.2/buildpacks.yaml
```diff
@@ -107,8 +107,8 @@ spec:
         - name: $(params.PLATFORM_DIR)
           mountPath: /platform
       securityContext:
-        runAsUser: 1000
-        runAsGroup: 1000
+        runAsUser: $(params.USER_ID)
+        runAsGroup: $(params.GROUP_ID)
```
vinamra28, line 111: AFAIK this doesn't works as currently Tekton doesn't supports integer type values and only supports `string` and `array` type values whereas `runAsUser` and `runAsGroup` requires integer values.

/cc @vdemeester @sbwsg 
Please correct me if I am wrong :sweat_smile:

vdemeester: Yeah… this is something we need to fix at some point

zhangtbj: Hi @vdemeester ,

Do you mean fix to support integer type or fix like what I did in this PR?

chmouel: I don't know very well how kubernetes behaves here in regard with this task, but what about just setting up : 

`.spec.securityContext.runAsNonRoot`

to `True` 

we would not need those runAsUser and Group anymore, right ?

zhangtbj: Yes, I think I see the problem:
`cannot unmarshal string into Go struct field SecurityContext.spec.steps.securityContext.runAsGroup of type int64`

I cannot modify it like this way. :(.

I think it cannot be fixed until Tekton fix the limitation.

Sorry for that, I will close this PR in my side.

chmouel: @zhangtbj I think you could just quote it with "" but i like to know if setting runAsNonRoot may be a better solution, unless you really want a fixed uuid

zhangtbj: Yes, I quote them like:
```
      securityContext:
        runAsUser: "$(params.USER_ID)"
        runAsGroup: "$(params.GROUP_ID)"
```
The same result. Yes, I think we like to fix it but `runAsNonRoot` may be not enough, it should use a dedicated user or group like 1000 in this step.

PR #596, opened 2020-12-08
Add parallelisation of tests

chmouel:
# Changes

We were running previously the tests sequentially,

Since the catalogs tests are growing exponentially, let's add parallelisation to
speed things up (default: 5 tests)

Moreover we add another related patch to be able to redefine KUBECONFIG for downstream CIs.

/area testing

### DIFF
--- a/test/e2e-common.sh
+++ b/test/e2e-common.sh
@@ -113,6 +113,8 @@ function show_failure() {
 }
 function test_task_creation() {
     local runtest
+    declare -A task_to_wait_for
+
     for runtest in $@;do
         # remove task/ from beginning
         local runtestdir=${runtest#*/}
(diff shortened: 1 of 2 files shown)


chmouel:
TIL the KUBECTL is already defined in other part of the scripts

```
++ KUBECTL='./cluster/kubectl.sh --match-server-version'
```

but we can't quite use it it seems because humm it's humm non existent 🤦 

using `KUBECTL_CMD` variable instead,

chmouel:
Added  7bd7cb8  to make sure this PR it works (since it would skip things up if it cannot detect any test changes)

chmouel:
Maven task is again flaky ! but this works and run tasks 5 by 5

vdemeester (approved):

vinamra28 (approved):
/lgtm

PR #597, opened 2020-12-08
Bump tekton pipeline version to 0.18.1

chmouel:
# Changes

This update the pipeline version to latest => 0.18.1

# Submitter Checklist

All good 👍🏼

/kind cleanup

### DIFF
--- a/task/git-batch-merge/0.2/git-batch-merge.yaml
+++ b/task/git-batch-merge/0.2/git-batch-merge.yaml
@@ -72,7 +72,7 @@ spec:
       default: "false"
     - name: gitInitImage
       description: The image used where the git-init binary is.
-      default: "gcr.io/tekton-releases/github.com/tektoncd/pipeline/cmd/git-init:v0.18.0"
+      default: "gcr.io/tekton-releases/github.com/tektoncd/pipeline/cmd/git-init:v0.18.1"
       type: string
   results:
     - name: commit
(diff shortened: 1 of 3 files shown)


task/git-batch-merge/0.2/git-batch-merge.yaml
```diff
@@ -72,7 +72,7 @@ spec:
       default: "false"
     - name: gitInitImage
       description: The image used where the git-init binary is.
-      default: "gcr.io/tekton-releases/github.com/tektoncd/pipeline/cmd/git-init:v0.18.0"
```
Divyansh42, line 75: Is the bump of image version required in these two tasks? because I couldn't find the change in image in [v0.18.1](https://github.com/tektoncd/pipeline/releases/tag/v0.18.1) pipeline release.

chmouel: it's just about sanity.. it's not needed but it makes easy next time when we sed the version out.

vdemeester (approved):

vinamra28:
can you please squash the commits :sweat_smile:

Divyansh42 (approved):
/lgtm

task/git-batch-merge/0.2/git-batch-merge.yaml
bobcatfish: I still think it's a confusing user experience to bump these without bumping the version of the Tasks - when we do this it means that users can pull the same Task at the same version and get different behavior depending on when they pull it, see https://github.com/tektoncd/pipeline/pull/3258

PR #618, opened 2021-01-19
Removing knctl-deploy task

### LINKED ISSUE #430: Remove "knctl-deploy" task from catalog
`knctl` has been sunset since 1.5 years and only supports Knative up to 0.2 (current version: 0.16). The successor is `kn` for which there are also entries in the catalog.

To avoid confusing for the user I suggest to remove `knctl` as it is not maintained anymore and I'm pretty sure does not work with current Knative installations (as it refers to apiversions which are not present anymore in any modern Knative installation).

@cppforlife (author of knctl) would this be ok for you ?

rhuss:
knctl has been sunset for 2 years and only supports Knative up to 0.2 (current version: 0.20).
The successor is kn for which there are also entries in the catalogue.

Fixes #430


navidshaikh (approved):
lgtm

but there seems to be another issue, see https://github.com/tektoncd/catalog/issues/619

vinamra28:
I agree to what @navidshaikh said related to CI issue but instead of removing this can we just deprecate this task by adding an annotation `tekton.dev/deprecated: "true"`? Probably you can find this [TEP-0003](https://github.com/tektoncd/community/blob/master/teps/0003-tekton-catalog-organization.md#deprecation--removal-strategy)

rhuss:
Signed the CLA with rhuss@redhat.com. Anything I need to do to retrigger the CLA check ?

rhuss:
@vinamra28 I see, deprecation is probably the better way according to the referenced document (although I can't see that this task will reliably work with any decent Knative installation, as all of the API version that is referenced in the knctl task has been already removed). I'm pretty sure it's broken these days, so naively I would just remove it (since knctl is abandoned also). 

How is this task tested ?

vinamra28:
> @vinamra28 I see, deprecation is probably the better way according to the referenced document (although I can't see that this task will reliably work with any decent Knative installation, as all of the API version that is referenced in the knctl task has been already removed). I'm pretty sure it's broken these days, so naively I would just remove it (since knctl is abandoned also).
> 
> How is this task tested ?

hi @rhuss, since the author of the task hasn't provided any tests thus this task is not tested. Only the validation of the task is tested.

vinamra28:
@rhuss, can you repush so that CLA becomes green?

rhuss:
Sorry, was totally distracted from this PR :)

So, you I read the task and just add a deprecation or is removing the task the right way ? (not sure again what we have decided)

vinamra28:
> Sorry, was totally distracted from this PR :)
> 
> So, you I read the task and just add a deprecation or is removing the task the right way ? (not sure again what we have decided)

Hi @rhuss, yes we can remove the task and proceed with the PR.
I am wondering why the EasyCLA is not green. Can you please just try to repush once?

rhuss:
Added a minor fix to a doc file so that a push triggers a rebuild. CLA bot is happy now :)

vinamra28 (approved):
/meow

PR #625, opened 2021-01-25
Proposed recommendations for authoring tasks

mogsie:
# Changes

I added some recommendations after some months of working with Tekton Pipelines, as recommended by @ImJasonH in the Tekton slack.

# Submitter Checklist

review them:

(removed N/A checkboxes)

### DIFF
--- a/recommendations.md
+++ b/recommendations.md
@@ -60,3 +60,204 @@ On the catalog, this means that you should, where possible:
         runAsUser: 0 # root uid == 0
         privileged: true
   ```
+
+## Don't use interpolation in shell scripts
+
+Using `$(tekton.task)` interpolation in the `script` or as a `sh -c`
+string is extremely fragile.  The interpolation done by tekton is not
+aware of the context in which the interpolation happens.  A space, a
+quote sign, a backslash or newline could easily thwart an otherwise
+beautiful script.
+
+```
+  steps:
+  - name: foo
+    image: myimage
+    script: |
+      echo $(params.one)
+```
+
+If `params.one` happens to contain a quote, then the resultin shell
+script might look like this:
+
+```
+echo '
+```
+
+This script is not valid, and the task will fail.
+
+Instead, use environment variables or arguments, which are not
+interpolated into the script
+
+```
+  steps:
+  - name: foo
+    image: myimage
+    env:
+      - name: PARAM_ONE
+        value: $(params.one)
+    script: |
+      echo "$PARAM_ONE"
+```
+
+The script will now correctly print out the value of $PARAM_ONE,
+regardless of what it contains.
+
+
+## Extract task code (scripts) to their own files
+
+As a task grows in complexity, it becomes harder and harder to
+maintain it in-line.  Because you have already avoided interpolation
+in the script, there is no real need for the script to be in-lined
+into the Task.
+
+Use a ConfigMap for your script, and mount the configmap in the task.
+Use a single _command:_ that executes your script (optionally with
+environment variables and with parameters):
+
+
+```
+  volumes:
+  - name: scripts
+    configMap:
+      name: my-task-scripts
+      defaultMode: 0755
+  steps:
+  - name: foo
+    image: myimage
+    volumeMounts:
+    - name: scripts
+      mountPath: /mnt/scripts
+    volume
+    command: 
+    - /mnt/scripts/my-command.sh
+```
+
+This allows you to create and update your configmap from a real
+script.  An external script-file allows you to edit it stand-alone as
+a proper script file, rather than in-lined into a tekton task yaml
+file.  This gives you better completion, syntax checking, and many
+other benefits.  You can even run the script locally.
+
+
+```
+#!/bin/bash
+# my-command.sh
+echo "${PARAM_ONE-Hello world}"
+```
+
+To create the configmap:
+
+```
+kubectl create configmap my-task-scripts --from-file=my-command.sh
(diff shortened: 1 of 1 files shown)


recommendations.md
```diff
@@ -60,3 +60,204 @@ On the catalog, this means that you should, where possible:
...
+```
+
+If `params.one` happens to contain a quote, then the resultin shell
+script might look like this:
+
+```
+echo '
+```
+
+This script is not valid, and the task will fail.
```
chmouel, line 87: I think :

```bash
var=$(cat <<EOF
$(params.one)
EOF
)
```

should work too

recommendations.md
```diff
@@ -60,3 +60,204 @@ On the catalog, this means that you should, where possible:
...
+For bonus points, use a kustomize generator to create your task; this
+allows you simple "kubectl apply -k" 
+
+## Remember that there are other languages than sh and bash
+
+Yes, sh and bash are DSLs for running processes, but sometimes there
+are other languages more suited for what you're trying to do.  Tekton
+Pipelines' main positive attribute is the ability to have the right
+tool available for every step, and that includes the _interpreter_.
+Use python or other scripting languages when that is warranted.
```
chmouel, line 164: FYI: this is how we do it in Python : 

https://github.com/tektoncd/catalog/blob/master/task/github-add-comment/0.2/github-add-comment.yaml#L76-L78

It has the same issue as shell, if you use `var = "$(params.foo)"` and params.foo has a " then you run into problems

chmouel (approved):
Sounds good to me, maybe we want to add that other languages can have this issue too!

@Divyansh42 @vinamra28 we have that problem too in the catalog, for example this task https://github.com/tektoncd/catalog/blob/master/task/send-to-channel-slack/0.1/send-to-channel-slack.yaml#L32 is using shell and params, so if you have a " in there (which is probably in a message) it would fail. We may want to fix it as recommended here

recommendations.md
mogsie: Yes, I'll 
* add a python example of the escaping issue
* add a pyton example on how to invoke a python script!

recommendations.md
mogsie: Almost.  Shell escape rules are treacherous:
* If I send in `$(cat)` as the parameter, the script will evaluate `cat` which will likely fail because there is nothing in stdin
* If I send in "EOF" then the parameter will fail, and have the value EOF.
I'm pretty sure that there are no "good" interpolation rules for shell (or any language for that matter); because tekton just doesn't know (and rightly so) about the chosen language's escaping rules.

So in my opinion, the rule of thumb should be to avoid it unless you're very certain about the data being passed in.

mogsie:
Sorry about the noise, I used the wrong e-mail address when committing, causing the EasyCLA to misfire.  I can probably expand on some sections, or remove some if you don't feel they should be recommendations.

* I moved the bit about using other languages _first_ so that I could use python examples further down.
* I expanded on the interpolation bit by adding a python example
* I also added a section about security issues, since that was my main concern with interpolation.

recommendations.md
chmouel: that's right 👍🏽

recommendations.md
```diff
@@ -60,3 +60,252 @@ On the catalog, this means that you should, where possible:
...
+Use python or another scripting languages when that is warranted.
+
+A python example:
+
+```
+  steps:
+    - name: foo
+      image: python:alpine
+      script: |
+        #!/bin/env python
```
chmouel, line 79: nit: maybe python3 :)

(/usr/bin/python is not found by default on fedora/ubuntu)

chmouel:
/lgtm

Thanks for updating this document 👍🏽

chmouel:
I didn't meant to +2A directly and not waiting for another reviewer review, sorry about this

recommendations.md
```diff
@@ -60,3 +60,252 @@ On the catalog, this means that you should, where possible:
...
+    volume
+    command: 
+    - /mnt/scripts/my-command.sh
+```
+
+This allows you to create and update your configmap from a real
+script.  An external script-file allows you to edit it stand-alone as
+a proper script file, rather than in-lined into a tekton task yaml
+file.  This gives you better completion, syntax checking, and many
+other benefits.  You can even run the script locally.
```
vinamra28, line 196: hi @mogsie, I agree with this point that it becomes easy to write a script and being able to test that locally but what I was thinking was will this not make the task mutable and as per [TEP-0003](https://github.com/tektoncd/community/blob/master/teps/0003-tekton-catalog-organization.md) we should make the task immutable so that we don't break experience of the users using the tasks from the catalog.
Also the other question which I have in my mind is where to keep the script that ?

mogsie: To be honest, I submitted these as general Task authoring guidelines; not guidelines for the Catalog per se.  Maybe these guidelines should be moved somewhere more appropriate (but it makes sense that the catalog espouses the recommendations, wherever they are).

But as for immutability, checking in a Task and ConfigMap to the catalog won't gain any of the benefits of extracting the script to its own file; not easily tested, edited, shellchecked, etc.  Committing it as a _kustomize_ manifest is of course an alternative.  Internally we use kustomize for these things, but I felt it was a stretch to have a recommendation to use kustomize.  It would of course be possible to provide a kustomize manifest in the catalog, and have simple instructions to create the ConfigMap by hand (if kustomize is not desirable) `--from-file` as described in the task.

PR #684, opened 2021-03-31
Go back to appending sidecars with python

### LINKED ISSUE #678: Using sed rather than python has potential downsides
I've realized that the changes made in #665 could cause issues if it's not the first function ran against the file to inject the sidecar.
The method it's using is to add `sidecar: blah blah` in a place that's valid but the problem arises when there is already a `sidecar:` block either there already or a modification/injection was run before and already added it.

Sorry :facepalm: 

There are a couple options for resolving this:

* Add some bash to check if there's an existing `sidecar:` block and inject in there (or continue as now if it doesn't find a `sidecar:` block), with the downside of being a bit brittle
* Go back to using the python, with the downside of needing python as a dependency again
  * optionally could clean it up so it's more readable
  * or function-ize it to make it easier for other sidecar injection
* Use `yq-go` to manipulate the yaml easier and safer in scripts, with the downside of arguably just going from one dependency to another
  * Handles the complexity of yaml for you (similar to `jq`)
  * tends to be pretty readable
  * I personally prefer yq to python for this use-case
    * There is the matter of future complex scripting where python might be needed/preferred and then you end up with both python + yq required

06kellyjac:
# Changes

Resolves #678

- Go back to using python to append yaml
  - I dislike python but after discussion in #678 it's clear the sensible choice is to go back to using it
- Check for libraries in a pip agnostic way
  - I install my python dependencies through [nix](https://search.nixos.org/packages?channel=unstable&show=python39Packages.pyyaml&from=0&size=50&sort=relevance&query=pyyaml) when I have to so the pip method doesn't work. This is likely the same if pyyaml is installed through apt/yum/etc
- allow for adding multiple sidecars with a reusable `add_sidecars` function
- fix overwriting sidecars issue for previous python
  - the previous python implementation would replace sidecars resulting in a _similar_ issue to the bash solution if `add_sidecar_registry` wasnt ran first
- make it more readable
  - There isn't really a cost to having the python multi-line & it's much more readable/maintainable

# Submitter Checklist

review them:

          ```

          spec:
            description: >-

          ```

for more details._

---

### DIFF
--- a/test/e2e-common.sh
+++ b/test/e2e-common.sh
@@ -42,21 +42,51 @@ function require_command() {
     fi
 }
 
-require_command ${KUBECTL_CMD}
+require_command ${KUBECTL_CMD} python3
+
+## Python Modules
+
+function require_python_module() {
+    for arg in "$@"; do
+        if ! python3 -c "import ${arg}" &> /dev/null;then
+            echo "required 'python3' module '${arg}' not be found"
+            echo "NOTE: the library name may be different but you you can try installing it via 'python3 -m pip install ${arg}'"
+            exit 1
+        fi
+    done
+}
+
+require_python_module yaml
 
 source $(dirname $0)/../vendor/github.com/tektoncd/plumbing/scripts/e2e-tests.sh
+
+# Add sidecar resources, first argument is the file
+# following arguments are string formed json/yaml processed by yaml.safe_load
+# e.g. '{"image": "registry", "name": "registry"}'
+function add_sidecars() {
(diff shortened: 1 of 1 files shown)


06kellyjac:
Force pushed a note about the possibility of python library names being different on pypi and importing within python

chmouel:
/lgtm

Looking Good! 🤙🏽

chmouel:
can't see the logs? 🤔 

/retest

vdemeester (approved):
/meow

PR #709, opened 2021-04-27
Raise the visibility of authoring recommendations

bittrance:
# Changes

Writing my first contribution to the catalog, I failed to find the authoring recommendations, resulting in unnecessary Slack noise. This adds references to parts of the documentation that I did read in the hope that others will see them in time.

I removed the recommendations list that was inlined in CONTRIBUTING.md as it is covered by the linked recommendations.md. The "portability" item was only partially covered by the recommendations, so I took the liberty of adding a section about backwards version compatibility, as that is the most important portability concern I can think of in this context.

# Submitter Checklist

These are the criteria that every PR should meet, please check them off as you review them:

### DIFF
--- a/recommendations.md
+++ b/recommendations.md
@@ -61,6 +61,13 @@ On the catalog, this means that you should, where possible:
         privileged: true
   ```
 
+## Be as portable and compatible as possible
+
+Make use of recent Kubernetes and Tekton features only when a user
+will expect it from the task's purpose. Your task may be of great use
+to users that have good reason not to upgrade right now. Your task
+should include the `tekton.dev/pipelines.minVersion`.
+
 ## Remember that there are other languages than sh and bash
 
 Yes, sh and bash are DSLs for running processes, but sometimes there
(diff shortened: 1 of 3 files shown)


recommendations.md
```diff
@@ -61,6 +61,13 @@ On the catalog, this means that you should, where possible:
         privileged: true
   ```
 
+## Be as portable and compatible as possible
+
+Make use of recent Kubernetes and Tekton features only when a user
+will expect it from the task's purpose. Your task may be of great use
+to users that have good reason not to upgrade right now. Your task
+should include the `tekton.dev/pipelines.minVersion`.
```
sbwsg, line 69: nit: `should` -> `must`.  The field is mandatory according to the PR template.

afrittoli:
Nice, thank you!
/lgtm

recommendations.md
afrittoli: nice catch!

vdemeester (approved):

PR #711, opened 2021-04-27
Pass git-clone Params as EnvVars Instead

sbwsg:
# Changes

Previously we passed params directly into the `script` block of
the clone step. Because Pipelines doesn't escape variable values
when interpolating them the behaviour of the task could be changed
by the contents of a param.

This commit changes the way params are provided to the script;
instead of being interpolated directly they are converted first
into environment variables and accessed through shell's `${}`
construct.

# Submitter Checklist

There are no behaviour changes to the git-clone tasks themselves.

### DIFF
--- a/task/git-clone/0.1/git-clone.yaml
+++ b/task/git-clone/0.1/git-clone.yaml
@@ -74,8 +74,33 @@ spec:
   steps:
     - name: clone
       image: $(params.gitInitImage)
+      env:
+      - name: PARAM_URL
+        value: $(params.url)
+      - name: PARAM_REVISION
+        value: $(params.revision)
+      - name: PARAM_REFSPEC
+        value: $(params.refspec)
+      - name: PARAM_SUBMODULES
+        value: $(params.submodules)
+      - name: PARAM_DEPTH
+        value: $(params.depth)
+      - name: PARAM_SSL_VERIFY
+        value: $(params.sslVerify)
+      - name: PARAM_SUBDIRECTORY
+        value: $(params.subdirectory)
+      - name: PARAM_DELETE_EXISTING
+        value: $(params.deleteExisting)
+      - name: PARAM_HTTP_PROXY
+        value: $(params.httpProxy)
+      - name: PARAM_HTTPS_PROXY
+        value: $(params.httpsProxy)
+      - name: PARAM_NO_PROXY
+        value: $(params.noProxy)
+      - name: WORKSPACE_OUTPUT_PATH
+        value: $(workspaces.output.path)
       script: |
-        CHECKOUT_DIR="$(workspaces.output.path)/$(params.subdirectory)"
+        CHECKOUT_DIR="${WORKSPACE_OUTPUT_PATH}/${PARAM_SUBDIRECTORY}"
 
         cleandir() {
           # Delete any existing contents of the repo directory if it exists.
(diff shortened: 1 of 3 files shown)


chmouel:
Nice trick, using bash variable expansion mechanism to do the quoting for us, maybe that's something we can do in pipeline controller internally? 🤔 

/lgtm

sbwsg:
> Nice trick, using bash variable expansion mechanism to do the quoting for us, maybe that's something we can do in pipeline controller internally? 🤔

Definitely possible!  This would expose all params to all steps. Today we don't do that but it might be a worthwhile trade-off to make so task authors don't have to manually do this.

bobcatfish:
hey all! a couple questions @sbwsg :

1. Since this is a change to the existing Tasks, I would think we'd want to bump the version? e.g. the purpose of the versions on the Tasks would be to communicate when things change and avoid consuming changes you dont want - i think this would be the case even when fixing (e.g. for releasing the controllers, we bump the patch version for fixes vs. re-releasing existing versions) - maybe it would help if we had a policy around removing old versions so that in case like this, we could delete the previous versions and add a new one?
2. Is providing the values as env variable enough to guard against changing the task behavior? I'm under the impression that nothing would prevent the script from using the variables in such a way that commands could still be injected - or maybe im wrong? (ive been trying to reproduce an example of this and failing so far so maybe im wrong
3. I imagine this is going to hit every single Task that uses script mode - does it make sense that we make this easier in some way?

vdemeester:
> 1. Since this is a change to the existing Tasks, I would think we'd want to bump the version? e.g. the purpose of the versions on the Tasks would be to communicate when things change and avoid consuming changes you dont want - i think this would be the case even when fixing (e.g. for releasing the controllers, we bump the patch version for fixes vs. re-releasing existing versions) - maybe it would help if we had a policy around removing old versions so that in case like this, we could delete the previous versions and add a new one?

The behaviour doesn't change, there is no need to bump the version, thus this change is, *or appears to be*, not a change you wouldn't want.

> 2. Is providing the values as env variable enough to guard against changing the task behavior? I'm under the impression that nothing would prevent the script from using the variables in such a way that commands could still be injected - or maybe im wrong? (ive been trying to reproduce an example of this and failing so far so maybe im wrong)

The script is written by the author, and thus is *immutable* from the user standpoint. If the current changes guard against injection (which I think does) then, it should be enough.

> 3. I imagine this is going to hit every single Task that uses script mode - does it make sense that we make this easier in some way?

Yes and no 🙃. The script feature comes with that problem, but it's more of a *"with great power comes great responsability"* than anything. We don't know which script language the user might use with the script feature ; could be something that doesn't exist today or that we don't know, … could be something that is not subject to the "bash injection" type, etc.. Trying to reduce those "bash type injection" would only result in limiting the usefulness of the script feature I think.

sbwsg:
> Is providing the values as env variable enough to guard against changing the task behavior? I'm under the impression that nothing would prevent the script from using the variables in such a way that commands could still be injected - or maybe im wrong? (ive been trying to reproduce an example of this and failing so far so maybe im wrong)

The script can still use the env var as code, but would need to be more explicit about it - by calling "eval" on it or writing and executing it from a file.

> I imagine this is going to hit every single Task that uses script mode - does it make sense that we make this easier in some way?

> We don't know which script language the user might use with the script feature ; could be something that doesn't exist today or that we don't know, … could be something that is not subject to the "bash injection" type, etc.. 

Pipelines could convert all params into environment variables in all Steps by default. This would be a language-independent change that puts the params' values inside any script's environment.

bobcatfish:
> The behaviour doesn't change, there is no need to bump the version, thus this change is, or appears to be, not a change you wouldn't want.

@vdemeester the behavior DID change tho - you can't inject anything into the script with env vars anymore. this seems to be on par with a bug fix, which in any other software we'd communicate via a version bump

vdemeester:
> @vdemeester the behavior DID change tho - you can't inject anything into the script with env vars anymore. this seems to be on par with a bug fix, which in any other software we'd communicate via a version bump

Right, but this is fixing an issue, and people shouldn't have to be able to do that in the first place. Anyone using env var to inject anything into the script was mis-using it.

PR #740, opened 2021-05-26
Add a script to help managing embedded scripts externally

chmouel:
The idea of this script is if you have inside your script something like this :

It will replace this #include with the content of python/script.py

This make it easier to write your tasks and use linters/code checkers from your
editor on your included script.

This sparked from @afrittoli  discussion here https://github.com/tektoncd/catalog/discussions/715

Currently this is a helper so this would be up to the contributor to uses it before commit but i think we can maybe think on how to do this post merge so to be transparent.

### DIFF
--- a/test/embed-script-externally/README.md
+++ b/test/embed-script-externally/README.md
@@ -1,0 +1,42 @@
+# Embed your script in external files
+
+This is a small script that helps you manage your embedded [scripts
+](https://github.com/tektoncd/pipeline/blob/main/docs/tasks.md#running-scripts-within-steps)
+inside your Tekton tasks externally.
+
+## Install
+
+You need the [ruamel.yaml](https://yaml.readthedocs.io/en/latest/) library, it
+should be availabe in fedora/debian default repositories or you can simply use
+pip to install it :
+
+```shell
+pip3 install ruamel.yaml
+```
+
+## Usage
+
+If for example you have a task like this :
+
+```yaml
+apiVersion: tekton.dev/v1beta1
+kind: Task
+metadata:
+  name: git-clone
+spec:
+  steps:
+    - name: step1
+      script: "#include script.sh"
+```
+
+The "embed_script" script will see it and includes the `script.sh` in place of
+your `"#include script.sh"`.
+
+At the end it will print every yaml documents even the ones who have not been
+substituted from includes. Each document will be separated with a `"---"`
+separator.
+
+## CAVEAT
+
+It doesn't support embedded `TaskSpec` or `PipelineSpec` yet but there is no
+reason this cannot be supported.
--- a/test/embed-script-externally/__init__.py
+++ b/test/embed-script-externally/__init__.py
@@ -1,0 +1,14 @@
+# -*- coding: utf-8 -*-
+# Author: Chmouel Boudjnah <chmouel@redhat.com>
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License. You may obtain
+# a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
--- a/test/embed-script-externally/embed_script.py
+++ b/test/embed-script-externally/embed_script.py
@@ -1,0 +1,72 @@
+#!/usr/bin/env python3
+#
+# Copyright 2021 The Tekton Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import argparse
+import io
+import os.path
+import re
+import sys
+import typing
+
+from ruamel.yaml import YAML
+
+REGEXP = r"^#include\s*([^$]*)"
+
+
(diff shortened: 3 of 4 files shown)


test/embedded-script-external/README.md
```diff
@@ -0,0 +1,37 @@
...
+kind: Task
+metadata:
+  name: git-clone
+spec:
+  steps:
+    - name: step1
+      script: "#include script.sh"
+```
+
+The script will see it and includes the `script.sh` in place of your "#include script.sh"
```
afrittoli, line 32: It would be nice to document that the result is sent to stdout and that if more than one YAML docs is in the source, all be parsed and separated by `---`.

test/embedded-script-external/embedded-script-external.py
```diff
@@ -0,0 +1,69 @@
...
+import re
+import sys
+import typing
+
+from ruamel.yaml import YAML
+
+REGEXP = r"^#include\s*([^$]*)"
+
+
+def replace(filename: str) -> typing.List:
```
afrittoli, line 27: Eventually we should have unit tests 🙏

afrittoli (commented):
Thank you for starting this!

I think it's fine to have it under a `test` folder at least for now - eventually we might want some way to make it easier for contributors to include this in their workflow... perhaps a `make` target or a shell script.

Eventually we'll need
- unit tests for this python
- a script to trigger this and write the output in the right place
- a CI job to run the script and check that generated YAMLs are up to date

But not in this PR :)

/lgtm

test/embedded-script-external/README.md
chmouel: Yep 👍🏻

test/embedded-script-external/embedded-script-external.py
chmouel: Yep I added some, i didn't bother because we don't have ways to run them yet 😅

I usually do pytest but i guess [python builtin unittest](https://docs.python.org/3/library/unittest.html) should be enough for this

chmouel:
> a script to trigger this and write the output in the right place

how would that work tho?

In main we have the "#include script.py" template files? But users (including ides/tkn hub or even curl people) would expect to be able to run the task directly from main branch, isnt it ?

maybe we can have both, we could have a comment above the `script: ` which would tell how to replace the script field to the CI jobs at post merge time, it would not matter the script content because the source of truth would be the #include script content.

Maybe that's a bit confusing, so let me rephrase my thinking : 

- Dev have task/foo/0.1/task.yaml
- task.yaml has something like this : 	
```yaml
steps:
   - name: foo
     #include foo.py
     script : | 
       #!/usr/bin/env python
       print("helo world")
```
- dev do some changes inside the foo.py and submit as PR,
- test-runner sees the #include and automatically test it with the new content from foo.py in script
- We merge the PR
- A post merge job would replace the script portion with the content from foo.py and commit back to catalog.

This allows user to have the latest task working directly and devs being able to develop into a real file for the script portion.

Let me know if that make sense 🙇🏻

afrittoli:
> > a script to trigger this and write the output in the right place
> 
> how would that work tho?
> 
> In main we have the "#include script.py" template files? But users (including ides/tkn hub or even curl people) would expect to be able to run the task directly from main branch, isnt it ?
> 
> maybe we can have both, we could have a comment above the `script: ` which would tell how to replace the script field to the CI jobs at post merge time, it would not matter the script content because the source of truth would be the #include script content.
> 
> Maybe that's a bit confusing, so let me rephrase my thinking :
> 
> * Dev have task/foo/0.1/task.yaml
> * task.yaml has something like this :
> 
> ```yaml
> steps:
>    - name: foo
>      #include foo.py
>      script : | 
>        #!/usr/bin/env python
>        print("helo world")
> ```
> 
>  * dev do some changes inside the foo.py and submit as PR,
> * test-runner sees the #include and automatically test it with the new content from foo.py in script
> * We merge the PR
> * A post merge job would replace the script portion with the content from foo.py and commit back to catalog.

I don't like this approach too much because it means that we need two commits to do a change, and in between commits the repo is in an inconsistent state.

> 
> This allows user to have the latest task working directly and devs being able to develop into a real file for the script portion.
> 
> Let me know if that make sense 🙇🏻

The workflow I had in mind is very much similar to what we have to generated golang files in `pipeline` and other repos.

* Dev creates / updates the task in `task/foo/0.1/task.yaml.template`
* task.yaml.template has something like this :

```yaml
steps:
   - name: foo
     #include foo.py
     script : | 
       #!/usr/bin/env python
       print("helo world")
```

* Dev creates / updates the script in `task/foo/0.1/foo.py`
* Before committing, dev runs `tools/embed-scripts.sh`, which generates `task/foo/0.1/task.yaml`

* CI runs `tools/embed-scripts.sh` and diffs. If differences are found it fails. Differences could happen if:
   * if the dev updated the script and did not run the embed tool
   * if the dev made changes into the generated `task.yaml` instead of the template / script 
* `task/foo/0.1/task.yaml` alone is published to the bundle
* users only ever refer to `task/foo/0.1/task.yaml`

The paths / naming might change, we could have the template defined in a subfolder for instance, to "hide" it from users.

afrittoli (commented):
I think this is a great start, thank you!
We'll need to create an infra image that embeds the script, so we can use it to run a CI job.
For that we'll need a docker file and a requirements.txt, but we can add those in a follow up.
/lgtm

chmouel:
/remove-lifecycle stale

i didnt see your review @vinamra28 🙇🏻

vinamra28:
/remove-lifecycle rotten
/lifecycle frozen
/reopen
@chmouel any updates on this?

chmouel:
I don't think it's trivial to do this, this  will broke the "kuebctl apply -f https://github.com/tektoncd/catalog/path/to/task.yaml" maybe with this TEP which split the structure from the implementation we may be able to achive that but this probably need something extras on the hub end...

chmouel:
> Not necessarily, it depends on how it's implemented. My suggestion would be:
> 
>     store in the catalog the template (YAML without the scripts), the script and the rendered version
>     the rendered version would be stored in the same location as today
>     add a CI job that renders the final task (template + script) and checks that it matches the rendered version in the repo

I think it may be a bit cumbersome for non frequent contributor, we have generation on cli for mans/golden etc.. and it's already a source of annoyances since it's easy to forget to run the command before sending the iteration. 

What about CI job commiting to the repo the final rendered version directly to a branch?

alternatively if hub could do this transparently and we say that hub is the "only" trusted source, then we would not have to store the yaml in repo.

>     provide catalog authors with a script to run to update all rendered tasks

PR #742, opened 2021-05-27
feature: add gitignore

shaowenchen:
Signed-off-by: shaowenchen <mail@chenshaowen.com>

# Changes

Ignore some unnecessary files from the repository.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,11 @@
+# Emacs garbage
+*~
+
+# VSCode config
+.vscode
+
+# JetBrains IDE config
+.idea
+
+# Python
+*.pyc


chmouel:
Thanks for your contribution but it's probably no the best way to do this. 

The editor configs are better addressed with a global exclusion in your ~/.gitconfig. At the same time you won't have to send a PR for a gitignore on every projects you work on and hasn't one :) 

See this article about the reasoning behind it and how to do it : 

https://julien.danjou.info/properly-managing-your-gitignore/

/hold

shaowenchen:
> Thanks for your contribution but it's probably no the best way to do this.
> 
> The editor configs are better addressed with a global exclusion in your ~/.gitconfig. At the same time you won't have to send a PR for a gitignore on every projects you work on and hasn't one :)
> 
> See this article about the reasoning behind it and how to do it :
> 
> https://julien.danjou.info/properly-managing-your-gitignore/
> 
> /hold

The majority of repositories will have a .gitignore file, including https://github.com/tektoncd.

This is just for developers and not every developer knows to configure a global .gitinore file.

chmouel:
Yep but where does it stop, if the user starts to use a new editor, is he or she going to add everything to the .gitignore every time because they didn't know how git exclude works ? Should we assume every one is using the same editors like vscode/golang, probably not, isnt it ? I would rather help and educate the user  on how to do configure global gitignore properly

shaowenchen:
OK.

PR #746, opened 2021-06-03
Add option to use Basic Auth

GijsvanDulmen:
# Changes

Add the option to influence the authorization type so a user could use Basic Auth with prefabricated b64 encoded token.
Not the most secure option, but it's another option for setting commit status.

# Submitter Checklist

### DIFF
--- a/task/github-set-status/0.2/README.md
+++ b/task/github-set-status/0.2/README.md
@@ -44,12 +44,14 @@ kubectl apply -f https://raw.githubusercontent.com/tektoncd/catalog/main/task/gi
   from the status of other systems. _e.g:_ `continuous-integration/tekton`
 * **STATE**: The state of the status. Can be one of the following `error`,
   `failure`, `pending`, or `success`.
+* **AUTH_TYPE**: The type of authentication to use. You could use the less secure "Basic"
+      for example. See https://docs.github.com/en/rest/overview/other-authentication-methods for more information.
 * **GITHUB_TOKEN_SECRET_NAME** \[optional\]: The name of the kubernetes secret that
   contains the GitHub token. Default value: `github`
 * **GITHUB_TOKEN_SECRET_KEY** \[optional\]: The key within the kubernetes secret that
   contains the GitHub token. Default value: `token`
 
-## Usage
+## Usage for Bearer authentication
 
 This TaskRun sets a commit on GitHub to `pending` getting tested by the CI.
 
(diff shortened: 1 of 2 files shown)


vinamra28 (commented):
can we add an example for basic auth as well so that user can know how to use basic auth with this task?
/ok-to-test

GijsvanDulmen:
> can we add an example for basic auth as well so that user can know how to use basic auth with this task?
> /ok-to-test

Added example and README about usage of Basic authentication!

vinamra28 (approved):

task/github-set-status/0.2/README.md
```diff
@@ -44,12 +44,14 @@ kubectl apply -f https://raw.githubusercontent.com/tektoncd/catalog/main/task/gi
   from the status of other systems. _e.g:_ `continuous-integration/tekton`
 * **STATE**: The state of the status. Can be one of the following `error`,
   `failure`, `pending`, or `success`.
+* **AUTH_TYPE**: The type of authentication to use. You could use the less secure "Basic"
+      for example
```
afrittoli, line 48: It would be nice to have a list of valid values or link to a doc with that list

task/github-set-status/0.2/README.md
```diff
@@ -77,3 +79,45 @@ spec:
...
+  name: github-set-status-run-on-commit-bd93
+spec:
+  taskRef:
+    kind: Task
+    name: github-set-status-with-basic-auth
+  params:
+    - name: REPO_FULL_NAME
+      value: tektoncd/catalog
+     - name: SHA
+      value: bd93869b489258cef567ccf85e7ef6bc0d6949ea
```
afrittoli, line 114: broken indentation

task/github-set-status/0.2/README.md
```diff
@@ -77,3 +79,45 @@ spec:
...
+## Usage for Basic authentication
+
+Make sure the token is fabricated by base64 encoding the username and password with a semicolon in between.
+Example:
+```
+Username: githubuser
+Password: reallyinsecure
+Pseudo-code: base64(username+":"+password)
+Result: Z2l0aHVidXNlcjpyZWFsbHlpbnNlY3VyZQ==
+```
```
afrittoli, line 93: Perhaps a small shell script would be better than pseudo-code? 
Quotes around the colon could break things depending on the language used...

afrittoli (commented):
Thank you!
A couple of NITs if you don't mind, then it looks ready to merge.

task/github-set-status/0.2/README.md
GijsvanDulmen: Link added!

task/github-set-status/0.2/README.md
GijsvanDulmen: Added a small script.

afrittoli:
I just realised that your PR changes the interface of the task, but it does so in a backward compatible way (a default is provided for the new params which matches the previous behaviour) so it’s probably ok have the same version number. 
I wonder however how someone would pull in the new version if they have v0.2 installed already and checked if a new version is available.
@tektoncd/catalog-maintainers what is the policy for these cases?

afrittoli:
> I just realised that your PR changes the interface of the task, but it does so in a backward compatible way (a default is provided for the new params which matches the previous behaviour) so it’s probably ok have the same version number.
> I wonder however how someone would pull in the new version if they have v0.2 installed already and checked if a new version is available.
> @tektoncd/catalog-maintainers what is the policy for these cases?

From the TEP

> "Changes to an existing version may however be permitted to non-functional parts like the metadata.annotations, labels of the resource, README, samples, but we should strive to keep that minimal so that the resource definition is independent of the time it was applied by the user. Also, new fields can be added to existing version with default values if it is not going to break pipeline created with existing version. This can be ensured by running tests on the updated resource." 

This suggests that it's ok to add this to the existing version.
I wonder though it this would not cause things like `tkn hub task update` to not pull in the new version?

For folks using OCI bundles, the tag will be overwritten, and a new sha will be generated.

GijsvanDulmen:
> > I just realised that your PR changes the interface of the task, but it does so in a backward compatible way (a default is provided for the new params which matches the previous behaviour) so it’s probably ok have the same version number.
> > I wonder however how someone would pull in the new version if they have v0.2 installed already and checked if a new version is available.
> > @tektoncd/catalog-maintainers what is the policy for these cases?
> 
> From the TEP
> 
> > "Changes to an existing version may however be permitted to non-functional parts like the metadata.annotations, labels of the resource, README, samples, but we should strive to keep that minimal so that the resource definition is independent of the time it was applied by the user. Also, new fields can be added to existing version with default values if it is not going to break pipeline created with existing version. This can be ensured by running tests on the updated resource."
> 
> This suggests that it's ok to add this to the existing version.
> I wonder though it this would not cause things like `tkn hub task update` to not pull in the new version?
> 
> For folks using OCI bundles, the tag will be overwritten, and a new sha will be generated.
> 
> I wonder if a better policy would be to create a new minor version in this case, to indicate that the API is backward compatible, but it's not the same exact task.

Just wondering. Could we merge this and open up a separate issue for this broader discussion? I guess having some minor versioning would be needed (implementation changes, not the stable interface of the task).

vdemeester (approved):
/meow
/lgtm

PR #750, opened 2021-06-08
[tasks/maven] Support to repository cache

crmarques:
# Changes

Support to use a persistent volume to hold downloaded maven artifacts, avoiding unnecessary resource consumption.


chmouel:
I think you mean 0.4 in your commit message

the diff https://paste.gg/p/anonymous/0659c8df83044ad4aefe97deab3c52d5

/ok-to-test

chmouel:
I see you are using a volume, what about a optional workspace instead?

crmarques:
Hi, @chmouel . Thank you for your response. I agree with you about using workspaces. It was my initial thought. But I saw this volume/podTemplate implementation in `jib-maven` and  `buildpacks:0.1` Tasks and imagined that was the recommendation.

I'll change it to workspace and update this PR.

tuananh:
if we want to use `git-clone` task with this, it will fail because there's more than 1 PVC right? `more than one PersistentVolumeClaim is bound`

tuananh:
to get it to work, you will need to set `disable-affinity-assistant` to `true`.

PR #762, opened 2021-06-22
fix argocd sync and wait task

pritidesai:
# Changes

Logging in from one container and having sync and wait run in the separate containers does not work. Authenticate once and reuse the connection across the containers in a pod is not supported. Sync and wait fails with the authentication error. Either we login at each step or collapse sync and wait along with the login step. Or we could authenticate once and
find a way to how to reuse the same connection across the containers.

For now, combining everything into a single step.

# Submitter Checklist

review them:

          ```

          spec:
            description: >-

          ```

for more details._

---

### DIFF
--- a/task/argocd-task-sync-and-wait/0.1/argocd-task-sync-and-wait.yaml
+++ b/task/argocd-task-sync-and-wait/0.1/argocd-task-sync-and-wait.yaml
@@ -35,14 +35,8 @@ spec:
     - name: login
       image: docker.io/argoproj/argocd:$(params.argocd-version)
       script: |
-        if [ -z $ARGOCD_AUTH_TOKEN ]; then
-          yes | argocd login $ARGOCD_SERVER --username=$ARGOCD_USERNAME --password=$ARGOCD_PASSWORD;
+        if [ -z "$ARGOCD_AUTH_TOKEN" ]; then
+          yes | argocd login "$ARGOCD_SERVER" --username="$ARGOCD_USERNAME" --password="$ARGOCD_PASSWORD";
         fi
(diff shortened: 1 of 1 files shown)


chmouel (commented):
/lgtm

Looking Good! 🤙🏽 thanks!

and FYI I am not sure how those images are made but if you had #!/bin/bash instead of #!/bin/sh available in the image you could easily avoid those quote issue with double brackets (ie https://stackoverflow.com/a/669486/145125)

vinamra28 (commented):
I wonder if we should deprecate 0.1 version of this task and add this as version 0.2 since this is a major change in this task ?
/cc @tektoncd/catalog-maintainers

chmouel:
If i understand correctly it's a fix of something that wasn't working at all, so i guess bumping does not change much, but let's see what @pritidesai thinks about this!

pritidesai:
> If i understand correctly it's a fix of something that wasn't working at all, so i guess bumping does not change much, but let's see what @pritidesai thinks about this!

Right, the current task definition is not working at all, this PR is just the fix so let's not bump the version right now.
Ideally I would like to add an automation `tests/` as well, we can bump the version with the `tests/` later.

vinamra28:
> > If i understand correctly it's a fix of something that wasn't working at all, so i guess bumping does not change much, but let's see what @pritidesai thinks about this!
> 
> Right, the current task definition is not working at all, this PR is just the fix so let's not bump the version right now.
> Ideally I would like to add an automation `tests/` as well, we can bump the version with the `tests/` later.

Makes sense. Will merge this for now then
/approve

PR #766, opened 2021-06-27
feature: added cargo task for rust projects

RobertLemmens:
# Changes
Added a cargo task, usefull when you need to compile rust applications. I couldn't find any so a while back created my own. Hoping it might be useful for others I decided to see if it could be added to the catalog.

I tried to keep it as simple as possible. Commands passed to cargo and compiler flags can be added by the user, but provide sensible defaults. This way you can also run just tests with cargo test, generate new projects in pipeline with cargo new, etc.

First time contributing, tried to read up and follow the guidelines but if anything is missing ill make sure to add it. Also a big kudos for the `run-test.sh`, It's small things like this that make contributing much more accessible!

# Submitter Checklist

review them:

          ```

          spec:
            description: >-

          ```

for more details._

---

### DIFF
--- a/task/cargo/0.1/cargo.yaml
+++ b/task/cargo/0.1/cargo.yaml
@@ -3,7 +3,7 @@ kind: Task
 metadata:
   name: cargo
   labels:
-    app.kubernets.io/version: "0.1"
+    app.kubernetes.io/version: "0.1"
   annotations:
     tekton.dev/pipelines.minVersion: "0.12.1"
     tekton.dev/tags: "build-tool, rust"
--- a/task/cargo/0.1/cargo.yaml
+++ b/task/cargo/0.1/cargo.yaml
@@ -1,0 +1,36 @@
+apiVersion: tekton.dev/v1beta1
+kind: Task
+metadata:
+  name: cargo
+  labels:
+    app.kubernets.io/version: "0.1"
+  annotations:
+    tekton.dev/pipelines.minVersion: "0.12.1"
+    tekton.dev/tags: "build-tool, rust"
+    tekton.dev/displayName: "Cargo"
+spec:
+  description: >-
+    This task can be used to execute cargo commands for rust projects.
+  workspaces:
+    - name: source
+      description: The workspace that contains your cargo.toml and rust code. This will be the working directory.
+  params:
+    - name: COMMAND
+      type: string
+      description: The cargo command that gets executed. Passed before compile flags. Results in 'cargo $COMMAND' being executed.
+      default: build
+    - name: COMPILE_FLAGS
+      type: array
+      description: Compilation flags passed to cargo. Results in 'cargo $COMMAND $COMPILE_FLAG1 $COMPILE_FLAG2 $COMPILE_FLAG3...' being executed.
+      default: []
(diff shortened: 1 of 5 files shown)


bobcatfish:
Thanks for the submission!

Reminder to check off the submitter checklist in the PR description 🙏  Are you planning to own and maintain this task going forward? If so you can add an OWNERS file with your username to make sure you get notified about changes ( @vinamra28 @jerop i'm surprised to not see this listed as a requirement 🤔  )

task/cargo/0.1/cargo.yaml
```diff
@@ -0,0 +1,40 @@
...
+      description: The image that contains cargo.
+        Defaults to the official rust image on the docker hub.
+        When using a custom image, make sure cargo is on the path.
+      default: docker.io/rust@sha256:ea81624bfff9a2d46fc181532acdeafda3cf8c31d25536d443ef902d4894f72a
+  steps:
+    - name: cargo-command
+      image: $(params.CARGO_IMAGE)
+      workingDir: $(workspaces.source.path)
+      command: [cargo]
+      args: ["$(params.COMMAND)", "$(params.COMPILE_FLAGS[*])"]
```
bobcatfish, line 40: a note that providing args directly as args like this makes it so that it might be possible to inject commands you don't want  - might make sense to pass these as env vars to make it harder to do this like in https://github.com/tektoncd/catalog/pull/711 - but that might be tough since COMPILE_FLAGS is an array so mayeb you should just ignore me for now 🤔 

(+ @jerop who has been looking into this)

task/cargo/0.1/tests/run.yaml
```diff
@@ -0,0 +1,79 @@
...
+      taskRef:
+        name: cargo
+      params:
+        - name: COMPILE_FLAGS
+          value: ['$(params.COMPILE_FLAGS[*])']
+      runAfter:
+        - git-clone
+      workspaces:
+        - name: source
+          workspace: source
```
bobcatfish, line 36: is it possible to add a task that runs after cargo-build that can verify that the cargo build worked? (np if not)

RobertLemmens: Thanks for taking a look! 

Good idea, that's a helpfull addition indeed. I've added a new task to the test pipeline that checks for the expected compiled binaries as a simple smoke test for now.

task/cargo/0.1/cargo.yaml
RobertLemmens: Yes, good catch. I am however unsure how to handle the array aswell. Command could be switched to env but doing just the command and not the compile flags would make the task definition a bit confusing so i'd rather do both at the same time. 


If someone from the team here has a good idea on how to do this I'd happily switch it around to env, if not, maybe this could be added in a later stage?

RobertLemmens:
Thanks for reviewing @bobcatfish . I've added an OWNERS file with my username and will maintain it, but see it as a _hopefully_ community driven task, so i hope more rust members will eventually come around to improve upon and co own it 😁.

tekton-robot tells me i need to be a member of tekton cd org in order to be mentioned in an owners file however, is there some action i need to take here?

PR #775, opened 2021-07-01
Add diff-task-version script

chmouel:
Using github.com/cli/cli it takes a PR URL as first argument and check
it out.

It will then detect automatically the old and new version and post a
comment to the PR with the diff and copy it in your clipboard (with
xclip on Linux and pbcopy on OSX)

### DIFF
--- a/hack/diff-task-version
+++ b/hack/diff-task-version
@@ -1,0 +1,114 @@
+#!/usr/bin/env bash
+# Chmouel Boudjnah <chmouel@chmouel.com>
+#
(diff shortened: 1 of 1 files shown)


vdemeester (approved):
Next step is to use this as part of the CI and have the bot do it for us :)

chmouel:
Next step would be to handle the edge cases, it's not robust enough for automatic ci yet

chmouel:
@tektoncd/catalog-maintainers  +1 or -1?

PR #778, opened 2021-07-04
test/e2e: handle fixtures configmap volume correctly

zhouhaibing089:
When adding `fixtures` volume, we need to consider whether the task
already has a volumes section. If the volumes is already defined, then
we should add `fixtures` rather than replacing the whole volumes.

The bug causes integration test failure for #776 

# Changes

test/e2e: handle fixtures configmap volume correctly

# Submitter Checklist

review them:

### DIFF
--- a/test/e2e-common.sh
+++ b/test/e2e-common.sh
@@ -233,7 +233,8 @@ function test_task_creation() {
         yaml=$(printf  ${taskdir}/*.yaml)
         started=$(date '+%Hh%M:%S')
         echo "${started} STARTING: ${testname}/${version} "
-        cp ${yaml} ${TMPF}
+        # dry-run this YAML to validate and also get formatting side-effects.
+        ${KUBECTL_CMD} -n ${tns} create -f ${yaml} --dry-run=client -o yaml >${TMPF}
         [[ -f ${taskdir}/tests/pre-apply-task-hook.sh ]] && source ${taskdir}/tests/pre-apply-task-hook.sh
         function_exists pre-apply-task-hook && pre-apply-task-hook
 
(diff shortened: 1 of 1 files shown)


vinamra28:
@zhouhaibing089 have you tested this PR on #776 ? If yes then will be happy to merge in that case

zhouhaibing089:
@vinamra28: I'm now testing in CI with cherry-picking #776 here.

zhouhaibing089:
Hmm, all right. The output from `sed` gives:

```yaml
  volumes:
    - name: fixtures
      configMap:
        name: fixtures
  - name: githubtoken
    secret:
      secretName: $(params.GITHUB_TOKEN_SECRET_NAME)
```

The indentation is different and fails the test.

GijsvanDulmen:
> Hmm, the output from `sed` gives:
> 
> ```yaml
>   volumes:
>     - name: fixtures
>       configMap:
>         name: fixtures
>   - name: githubtoken
>     secret:
>       secretName: $(params.GITHUB_TOKEN_SECRET_NAME)
> ```
> 
> The indentation is different and fails the test.

Is this a problem of the `sed` choice or a problem of the indenting of the other PR?

zhouhaibing089:
@GijsvanDulmen: In my opinion, the `sed` command needs to honor the current number of indentation spaces. This should be straightforward if I can match two consecutive lines, but I'm not sure how to do that yet.

GijsvanDulmen:
> @GijsvanDulmen: In my opinion, the `sed` command needs to honor the current number of indentation spaces. This should be straightforward if I can match two consecutive lines, but I'm not sure how to do that yet.

Probably using `yq` would make it more robust as well.

zhouhaibing089:
> Probably using `yq` would make it more robust as well.

This requires `yq` to be present. I can take that as next action item on how this tool can be included.

At the same time, I figured out another way of doing it - I'm going to run `kubectl create -f ${yaml} --dry-run=client -o yaml` to get a well indented file(as well as some validations).

zhouhaibing089:
@vinamra28 @GijsvanDulmen: I believe this change is now ok to proceed. The integration test is passing.

GijsvanDulmen:
> @vinamra28 @GijsvanDulmen: I believe this change is now ok to proceed. The integration test is passing.

Nice! When this is merged I will rebase the related PR and get it merged :-)

vinamra28:
@zhouhaibing089 can you please remove the commit 9e32792fff7222fc2e831d5aa00e76f06458b91b since that is the part of #776 PR. sorry for the late reviews :sweat_smile:

zhouhaibing089:
@vinamra28: No problem at all. :) - I've removed the second commit.

PR #825, opened 2021-09-23
Fix default user home git cli

frerikandriessen:
# Changes
I made the default user home `/root`, because that was the only way my task could work when I used ssh-credentials. 
This is because the git ssh client looks into /root/.ssh for your config&credentials, but they are by default copied to /tekton/home.

Maybe this change brings other problems that I haven't thought of, I'd be happy to hear them.

# Submitter Checklist

review them:

          ```

          spec:
            description: >-

          ```

for more details._

---


task/git-cli/0.3/README.md
```diff
@@ -0,0 +1,154 @@
...
+        git pull origin master
+```
+
+## Using SSH credentials
+
+This Task supports fetching private repositories using SSH credentials.
+
+1. Bind an `ssh-directory` workspace to this Task.
+The workspace should contain private keys (e.g. `id_rsa`), `config`
+and `known_hosts` files - anything you need to interact with your git remote
```
vinamra28, line 74: Can you please add a guide how to obtain `id_rsa`, `config` and `known_hosts`? (I know that it was not there earlier but if possible can you add? :sweat_smile: )

task/git-cli/0.3/git-cli.yaml
```diff
@@ -0,0 +1,139 @@
...
+    Git command that needs to be run can be passed as a script to
+    the task. This task needs authentication to git in order to push
+    after the git operation.
+
+  workspaces:
+    - name: source
+      description: A workspace that contains the fetched git repository.
+
+    - name: input
+      description: A workspace that contains file that needs to be added to git.
```
vinamra28, line 27: can we make this as optional as it may not be necessary that user may bind a workspace containing the file which needs to be committed to git?

task/git-cli/0.3/git-cli.yaml
```diff
@@ -0,0 +1,139 @@
...
+        A .ssh directory with private key, known_hosts, config, etc. Copied to
+        the user's home before git commands are executed. Used to authenticate
+        with the git remote when performing the clone. Binding a Secret to this
+        Workspace is strongly recommended over other volume types.
+
+    - name: basic-auth
+      optional: true
+      description: |
+        A Workspace containing a .gitconfig and .git-credentials file. These
+        will be copied to the user's home before any git commands are run. Any
```
vinamra28, line 41: Can we make this as a common workspace and in the script check for the corresponding credentials ? On one side it can be easy for the user to remember that for ssh there is a different workspace and for basic auth the workspace is different but on one hand the workspaces list increases :sweat_smile:

vinamra28 (commented):
just adding a couple of thoughts

task/git-cli/0.3/README.md
frerikandriessen: Hm, isn't a primer on SSH a bit out of scope for this task? Or do you mean just a link to an article like https://www.digitalocean.com/community/tutorials/ssh-essentials-working-with-ssh-servers-clients-and-keys?

task/git-cli/0.3/git-cli.yaml
frerikandriessen: Great idea!

task/git-cli/0.3/README.md
vinamra28: yeah link to the article sounds good :+1:

task/git-cli/0.3/git-cli.yaml
frerikandriessen: I'm not sure if that would make it more maintainable. Both the authorization workspaces and the code are now cleanly separated. The entire ssh workspace even gets copied in its entirety, so I don't know on what files to filter (`id_rsa`, `config` and `known_hosts` are just suggestions). I mean we could probably do it we really want to, but I think it's for little gain.

task/git-cli/0.3/git-cli.yaml
frerikandriessen: I've added this explanation
```
* **input**: An optional workspace that contains the files that need to be added to git. You can
    access the workspace from your script using `$(workspaces.input.path)`, for instance:

      cp $(workspaces.input.path)/file_that_i_want .
      git add file_that_i_want
      # etc
```
but I'm not actually sure if that is correct 😅. _Is_ that the correct order of evaluation? Or will the reference inside the script not be evaluated anymore?

task/git-cli/0.3/git-cli.yaml
vinamra28: sure! makes sense. Thanks

task/git-cli/0.3/git-cli.yaml
vinamra28: yeah this will work.

vinamra28:
can you please squash the commits as well?

frerikandriessen:
Commits squashed

PR #827, opened 2021-09-28
Fail github-open-pr task if http error has occured

frerikandriessen:
# Changes

Previously, if you tried to create a PR on github but were unsuccesful (no 2XX result), the task would still pas. To me that's strange, because you intended to do something with the task and it was (for whatever reason) unable to. So this fixes that. Couple of questions I still have:

1. How can we write tests to ensure a task fails on certain input? I think we can only write passing-task tests right now, correct?

2. We could support letting the user choose wether or not the task should fail. I think one should eliminate the underlying issues of why the task is failing, but ok. However, we promise both a 'number' and 'url' of the created PR; naturally, we cannot give that when the POST was unsuccesful, so not letting the task fail feels like a break of contract.

3. There are several other similar tasks interacting with the GitHub API that don't fail on unsuccesful POST.
	3.1 Should I open PR's on those tasks for the same changes?
	3.2 Are there possibilities for us to support certain core/utility methods shared between tasks, such as in this case HTTP posting and error handling? I guess not, and that we want to keep al tasks completely unrelated to each other (they just happen to have the same code copied+pasted in some places).
# Submitter Checklist

review them:

          ```

          spec:
            description: >-

          ```

for more details._

---


chmouel:
/lgtm

👍🏻

vinamra28:
can you please squash the commits and then we can merge this ?

frerikandriessen:
Commits squashed

PR #847, opened 2021-10-19
git-clone Add documentation on short lived token

chmouel:
add a section in Readme about short lived tokens that can be used in
lieu of password with a note about bitbucket server workaround.


sbwsg:
Cheers!

/lgtm

PR #868, opened 2021-11-22
update (npm): add v0.2 with PACKAGE_MANAGER param support

waveywaves:
# Changes

npm task's behavior defaults to using the npm command. Users
would like to use the same task with yarn or some other package manager
as their node env based package manager if npm is not their choice of
pkg manager. To enable this, this commit allows the user to specify the
PACKAGE_MANAGER param (which default to npm) where the user can specify some
other package manager available in the nodejs environment.

# Submitter Checklist

review them:

          ```

          spec:
            description: >-

          ```

for more details._

---

### DIFF
--- a/task/npm/0.2/npm.yaml
+++ b/task/npm/0.2/npm.yaml
@@ -1,0 +1,44 @@
+apiVersion: tekton.dev/v1beta1
+kind: Task
+metadata:
+  name: npm
+  labels:
+    app.kubernetes.io/version: "0.1"
+  annotations:
+    tekton.dev/pipelines.minVersion: "0.17.0"
+    tekton.dev/categories: Build Tools
+    tekton.dev/tags: build-tool
+    tekton.dev/platforms: "linux/amd64,linux/s390x,linux/ppc64le"
+spec:
+  description: >-
+    This task can be used to run npm goals on a project.
+
+    This task can be used to run npm goals on a project
+    where package.json is present and has some pre-defined
+    npm scripts.
+  workspaces:
+    - name: source
+  params:
+    - name: PATH_CONTEXT
+      type: string
+      default: "."
+      description: The path where package.json of the project is defined.
+    - name: ARGS
+      type: array
+      default: ["version"]
+      description: The npm goals you want to run.
+    - name: IMAGE
+      type: string
+      default: "docker.io/library/node:12-alpine@sha256:dfbebf17bfb014e1e7068e76325a117bccf8679c68aec6a28514184a209c8bae"
+      description: The node image you want to use.
+  steps:
+    - name: npm-run
+      image: $(params.IMAGE)
+      command:
+        - "npm"
+      args:
+        - $(params.ARGS)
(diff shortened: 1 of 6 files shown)


piyush-garg:
Hey @waveywaves 

Can you please follow this guideline https://github.com/tektoncd/catalog/blob/main/CONTRIBUTING.md#how-to-contribute-a-new-version-of-a-task-or-pipeline for bumping the version of a task? It will help to review.

Also `COMMAND` param looks confusing to me, we can have param like `PACKAGE_MANAGER` or something else, but command looks like user need to provide full command not only package manager.

waveywaves:
@piyush-garg thanks for the help :D PACKAGE_MANAGER does make more sense and is less ambiguous
I have done the needful, lmk if there are any other changes I am missing out

task/npm/0.2/npm.yaml
```diff
@@ -0,0 +1,48 @@
...
+      description: The goals you want to run for the given command.
+    - name: IMAGE
+      type: string
+      default: "docker.io/library/node:12-alpine@sha256:dfbebf17bfb014e1e7068e76325a117bccf8679c68aec6a28514184a209c8bae"
+      description: The node image you want to use.
+  steps:
+    - name: npm-run
+      image: $(params.IMAGE)
+      command:
+        - $(params.PACKAGE_MANAGER)
```
vinamra28, line 42: hey @waveywaves, I am bit confused over here :sweat_smile:. Instead of adding a new version of npm task should we add this as a new task in catalog since the current task name is `npm` which means that it will mostly be running `npm` commands and users might get confused. Any thoughts?

waveywaves: I had the same thought initially, I am glad you are addressing it. 

I was initially going to create another task for yarn itself as I thought it might need another image which installs yarn. But, the nodejs image provides both yarn and npm as choices of package manager, hence it only made sense at the time to allow users to specify the package manager instead of creating an entirely different task for it. 

I guess it boils down to whether it makes sense to create another task for yarn or provide the flexibility for the user to not have to install an almost identical task and use the existing one by changing a param (which might be more usable).

Users might get confused only if the documentation around the param is not succinct enough or if they haven't read it. I can update the documentation accordingly if that is the change required also.

It's your call, I can create another PR for the yarn task if that is something that makes more sense to the catalog team.

piyush-garg: Hey @waveywaves 

Let me clear here, @vinamra28 is not asking for adding a new task for `yarn`. He is asking to add a new task maybe by the name of `nodejs` or something else which is flexible to use both `yarn` and `npm`. Adding the support for `yarn` in the task named `npm` does not look good. Also creating the two tasks doing the same thing is also not ideal.

I would say let's deprecate the `npm` task and add a new one with both `npm` and `yarn` support.

Otmane-Guo: Hi guys, can you also update the nodejs image in the existing npm task, it looks like node16 has better performance

PuneetPunamiya:
Closing this pr as @waveywaves  to open a new pr with name `nodejs`
/close

Issue #878, opened 2021-12-09
git-cli: Mounted workspace ssh-directory doesn't take effect then service account has attached SSH key secret


juggernaut:
# Expected Behavior
The ssh-directory workspace configuration should be able to take effect even when the associated service account has an attached SSH key secret as documented in https://tekton.dev/vault/pipelines-v0.14.3/auth/ .

# Actual Behavior
The SSH key attached to the service account (in /tekton/.creds) is always chosen to connect to the remote instead of the specified key in ssh-directory. Note that ssh-directory is configured to point to the correct key in the SSH config file, but it doesn't take effect.

# Steps to Reproduce the Problem

1. Service account associated with a SSH key as documented in https://tekton.dev/vault/pipelines-v0.14.3/auth/#ssh-authentication-git
2. Run a pipeline with the git-cli task. Example:
```
    - name: clone-repo
      taskRef:
        name: git-cli
      workspaces:
        - name: ssh-directory
          workspace: git-manifests-repo-ssh-creds
        - name: source
          workspace: cd-ws
.....
```
git-manifests-repo-ssh-creds contains an SSH key different than that associated with the service account.
3. You'll see the following error in the logs:
```
[clone-repo : git] git@github.com: Permission denied (publickey).
```

# Additional Info

Basically, I'm running multiple pipelines under the same service account, some of which are using deprecated `PipelineResource`s. However, it looks like even "newer" pipelines that don't use `PipelineResource` run the credentials initialization process. I want the newer pipelines to use git-cli with (different) SSH keys mounted in the ssh-directory.
There are a couple of problems: [this cp command](https://github.com/tektoncd/catalog/blob/cebb548d561259fef7239f9c1e56db9b241c16f0/task/git-cli/0.3/git-cli.yaml#L128) looks like is intended to replace /root/.ssh with ssh-directory. However, when /root/.ssh already exists, it instead copies ssh-directory as a subdirectory within it.
See the following log snippet for clarity:

```
[commit-staging-image-bump : git] + cp -R /workspace/ssh-directory /root/.ssh
[commit-staging-image-bump : git] + chmod 700 /root/.ssh
[commit-staging-image-bump : git] + chmod -R 400 /root/.ssh/config /root/.ssh/id_ssh-key /root/.ssh/known_hosts /root/.ssh/ssh-directory
```

[...]

PR #879, opened 2021-12-10
Fix git-cli recursive copy command and support GIT_SSH_COMMAND

juggernaut:
# Changes

 * Fixes problem with cp -R described in https://github.com/tektoncd/catalog/issues/878
 * Support GIT_SSH_COMMAND to be able to override SSH command

# Submitter Checklist

review them:

          ```

          spec:
            description: >-

          ```

for more details._

---

### DIFF
--- a/task/git-cli/0.4/git-cli.yaml
+++ b/task/git-cli/0.4/git-cli.yaml
@@ -1,0 +1,146 @@
+---
+apiVersion: tekton.dev/v1beta1
+kind: Task
+metadata:
+  name: git-cli
+  labels:
+    app.kubernetes.io/version: "0.3"
+  annotations:
+    tekton.dev/pipelines.minVersion: "0.21.0"
+    tekton.dev/categories: Git
+    tekton.dev/tags: git
+    tekton.dev/displayName: "git cli"
+    tekton.dev/platforms: "linux/amd64,linux/s390x,linux/ppc64le"
+spec:
+  description: >-
+    This task can be used to perform git operations.
+
+    Git command that needs to be run can be passed as a script to
+    the task. This task needs authentication to git in order to push
+    after the git operation.
+
+  workspaces:
+    - name: source
+      description: A workspace that contains the fetched git repository.
+
+    - name: input
+      optional: true
+      description: |
+        An optional workspace that contains the files that need to be added to git. You can
+        access the workspace from your script using `$(workspaces.input.path)`, for instance:
+
+          cp $(workspaces.input.path)/file_that_i_want .
+          git add file_that_i_want
+          # etc
+
+    - name: ssh-directory
+      optional: true
+      description: |
+        A .ssh directory with private key, known_hosts, config, etc. Copied to
+        the user's home before git commands are executed. Used to authenticate
+        with the git remote when performing the clone. Binding a Secret to this
+        Workspace is strongly recommended over other volume types.
+
+    - name: basic-auth
+      optional: true
+      description: |
+        A Workspace containing a .gitconfig and .git-credentials file. These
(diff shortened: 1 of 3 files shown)


task/git-cli/0.4/git-cli.yaml
```diff
@@ -125,8 +133,9 @@ spec:
         fi
 
         if [ "${WORKSPACE_SSH_DIRECTORY_BOUND}" = "true" ] ; then
-          cp -R "${WORKSPACE_SSH_DIRECTORY_PATH}" "${PARAM_USER_HOME}"/.ssh
+          mkdir -p "${PARAM_USER_HOME}"/.ssh
           chmod 700 "${PARAM_USER_HOME}"/.ssh
```
chmouel, line 137: usually we use `install -m0700 -D` for that to get the two mkdir and chmod as one command.

(but no need to change since I am not even sure install is avail and it's just the same)

chmouel:
I am good to merge this but I wonder how do you use GIT_SSH_COMMAND ? I am guessing you are using this to do pass some magic switch to the ssh binary ? if you have a special need for another ssh flags, maybe other users would have this need, is there any chance to describe it in the README? that would greatly help others.

a more general comment about git-clone : 

we get plenty of people wanting to pass environment variable, i wonder if instead of having to list one by one (and bump a new version for it which is not totally free)  to have a catch all parameters, for examples : 

```ssh
params: 
 - name: GIT_EXTRA_ENVS
   type: array
```

where the array would have : 

`["GIT_SSH_COMMAND=foo", "GIT_OTHER_ENV=bar"]`

and we export all env variables in that array from the script 
  
may be more scalable

juggernaut:
> I am good to merge this but I wonder how do you use GIT_SSH_COMMAND ?

I'm passing in the specific SSH key I want to use to connect to the remote git repo (e.g `ssh -i /root/.ssh/id_ed25519 -o IdentitiesOnly=yes` , but it's useful in general to override ssh command parameters. 

> we get plenty of people wanting to pass environment variable, i wonder if instead of having to list one by one

Isn't this a problem of with catalog tasks in general? There doesn't seem to be an easy way to pass in environment variables to tasks from the pipeline spec without forking the task and explicitly declaring it as a parameter. If there was a way to do that, I would likely not have needed to rev this task at all (just pass in GIT_SSH_COMMAND to the existing git-cli task would work). 

I have no problem changing it to `GIT_EXTRA_ENVS` like you suggest, but if there's something on the roadmap already to address this concern, it may not make sense to add such a thing to git-cli in particular. Let me know either way.

chmouel:
I think there was a proposal for something like this for pipeline as to be able to set arbitrary environments to pipelines/tasks from runs

I think it would be great to have GIT_EXTRA_ENVS as for the other env variable needed in the future and i agree it's a bit hacky and should probably better suited in core pipeline.

Thanks

/lgtm

task/git-cli/0.4/git-cli.yaml
```diff
@@ -0,0 +1,159 @@
...
+        the gitInitImage param with an image containing custom user configuration.
+      type: string
+      default: "/root"
+
+    - name: VERBOSE
+      description: Log the commands that are executed during `git-clone`'s operation.
+      type: string
+      default: "true"
+
+    - name: GIT_EXTRA_ENV_VARS
```
juggernaut, line 89: @chmouel I've added a generic way to pass env vars per your suggestion. However, I couldn't get it to work as type "array" -- it seems array expansion is only allowed in fields that accept array parameters. Expanding an array param like "$(params.ARRAY[*])" in a script results in a validation error.

PR #907, opened 2022-01-19
Buildah 0.4: Make buildah image run by default as user to build images

chmouel:
Create a new version of the buildah task to run by default as user for building images which should be much  more secure.

We are not using upstream buildah since it has some issues when running on kind, the UBI based version from catalog.redhat.com works fine on OpenShift4.x and Kind.

We are merging the two steps together since the sharing of /var/lib/containers doesn't seem to work as non root but that's should not change anything from a user pov.

### DIFF
--- a/task/buildah/0.4/README.md
+++ b/task/buildah/0.4/README.md
@@ -1,0 +1,67 @@
+# Buildah
+
+This Task builds source into a container image using Project Atomic's
+[Buildah](https://github.com/projectatomic/buildah) build tool. It uses
+Buildah's support for building from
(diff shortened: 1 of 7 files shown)


vdemeester (approved):
/meow

chmouel:
there is something fishy going on GKE

```
"time="2022-01-19T13:52:38Z" level=error msg="Error while applying layer: ApplyLayer exit status 1 stdout:  stderr: remount /, flags: 0x44000: permission denied""
```

works fine on kind

Issue #784, opened 2021-07-19
Update policy to bump Task/Pipeline versions whenever a change is made


bobcatfish:
# Expected Behavior

As a user of Tasks in the catalog, when I use Task (or Pipeline) and I specify a specific version X, I should be confident that the Task (or Pipeline) I use at time Y is the same as the version at time Y+1, etc.

That is: if anything that is user facing in a Task (or Pipeline) changes, the version of should be bumped.

# Actual Behavior

We make changes to Tasks without bumping the version, for example:
* https://github.com/tektoncd/catalog/pull/763: Adds a new variable to git-clone 0.1 (not 0.2 or 0.3, not sure why not), and changes the default value of that variable for 0.4
* https://github.com/tektoncd/catalog/pull/711: Converts params to env vars before using them

Anyone grabbing the git-clone 0.1 or 0.4 Task BEFORE these changes would get something different AFTER these changes (and which version will be available in our OCI registry? I'm assuming the latest one, but not sure?)

My understanding of why these changes being made without bumping the version is b/c we view these as additive; however:

1. without knowing exactly how users are using the Tasks/Pipelines, we can't know if these changes will break them (see steps to reproduce)
2. with everything else we version (e.g. releases of Tekton Pipelines), we create a new version even if the changes are additive, or even in the case of bug fixes: we don't update v0.14.0 to contain a fix, we make a new release, v0.14.1, so that we can clearly communicate to our users that there has been a change

# Steps to Reproduce the Problem

I want to demonstrate an example of how this can impact a user by describing how this impacted me recently.

I've been working on a custom task that runs a Pipeline as a TaskRun (https://github.com/tektoncd/experimental/pull/770) and I created an example Pipeline which used git-clone 0.3.

1. I installed `git-clone` with `tkn hub install task git-clone` (i didnt explicitly specify the version i needed but let's pretend i said `0.3` which is what i ended up working with)
2. I created a Pipeline that used `git-clone` and used it to experiment
3. I decided to create an end to end test that used the same `git-clone` - I wrote the TaskRun I expected (with the git-clone steps, params, etc. embedded in it) based on the `git-clone` task i had installed
4. The last part of my test setup was to grab the `git-clone` 0.3 yaml from the catalog repo and add it to a `testdata` folder
5. My tests failed! because the Task had changed since I grabbed it (the params -> env var change), my resulting TaskRun was now different

[...]

PR #918, opened 2022-01-31
Versioning Resources

jerop:
# Changes

Today, we don't always bump versions of resources when updating them. The guideline has been to bump versions only when behavior changes, but it's hard to figure out when the behavior has changed (a change that could be trivial to one user could be meaningful to another).

Not bumping resource versions when changing them causes issues where the resource definition becomes dependent on the time when it was applied by the user - which causes unexpected failures as described in https://github.com/tektoncd/catalog/issues/784.

This issue also came up as an issue where users cannot depend on the Step indices because they can change: https://github.com/tektoncd/community/pull/572#pullrequestreview-836578705.

In TEP-0003, we already proposed that a policy for versioning of resources:
https://github.com/tektoncd/community/blob/main/teps/0003-tekton-catalog-organization.md#versioning-resources

In Catalog Working Group on 01/13/2022, we revisited that policy and:
- agreed to follow the versioning policy
- make it easier to bump resources (see https://github.com/tektoncd/plumbing/pull/994 - thanks @sbwsg 🎉)

This change documents the versioning policy in the contributions guide.

Fixes https://github.com/tektoncd/catalog/issues/784

# Submitter Checklist

review them:

- [n/a] Follows the [authoring recommendations](https://github.com/tektoncd/catalog/blob/main/recommendations.md)

_See [the contribution guide](https://github.com/tektoncd/catalog/blob/master/CONTRIBUTING.md) for more details._

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -30,6 +30,12 @@ The process for contributing looks like this:
 
 ## How to Contribute a new version of a Task or Pipeline
 
+To provide stability of resources, we require that resources have to be bumped
+to new versions when making changes instead of modifying existing versions. This
+ensures that the resource definition is independent of the time it was applied to
+the cluster. Changes to existing versions may be permitted only for non-functional
+parts of the resources, e.g. labels and annotations, but we try to keep that minimal.
+
 If you are planning to add a new version of a Task or Pipeline make sure to
 separate your changes from the copied task. This makes it easy for reviewers to
 review the changes and not the actual copy.


vinamra28 (approved):
/cc @tektoncd/catalog-maintainers

chmouel:
if every time there is an update of a software which is in a task or we have to fix something in it (like documentation) we have to bump this may results a large number of tasks. 

It's maybe find and I am not sure if it's bad or not  and could end up with some scaling issues.

I wish we could use git as versioning system instead of directory structure but I am not sure how we can achieve that.

so basically  my vote is neutral :)

jerop:
hoping the work in [TEP-0101](https://github.com/tektoncd/community/pull/663) will address this issue, so closing this pull request

PR #962, opened 2022-04-05
added -n to IMAGE_URL result to avoid adding a line feed

ggalloro:
Added -n to the echo command used to export the IMAGE_URL result in order to avoid to add a new line after the image url that caused gke-deploy in a pipeline to fail with a "could not parse reference" when IMAGE_URL result is used as an argument to the --image parameter in gke-deploy

### DIFF
--- a/task/kaniko/0.6/kaniko.yaml
+++ b/task/kaniko/0.6/kaniko.yaml
@@ -62,4 +62,4 @@ spec:
       script: |
         set -e
         image="$(params.IMAGE)"
-        echo "${image}" | tee "$(results.IMAGE_URL.path)"
+        echo -n "${image}" | tee "$(results.IMAGE_URL.path)"


vinamra28:
> 

@ggalloro I guess for this change we can edit the previous version 0.6 as well, I guess it will not break existing users.
@vdemeester @chmouel thoughts?

ggalloro:
@vinamra28 agree, was not sure if using a new version of not, but editing existing version seems more appropriate.

vdemeester:
I think it would be fine editing the existing task indeed.

vinamra28:
@ggalloro can you please squash the commits and then we can merge this?

ggalloro:
> @ggalloro can you please squash the commits and then we can merge this?

done

vdemeester (approved):

PR #966, opened 2022-04-11
add type to params (needed for openshift)

jdurzi:
# Changes

Added a type to the params to prevent the Task breaking in OpenShift Pipelines (this requires an explicit type)

# Submitter Checklist

review them:

          ```

          spec:
            description: >-

          ```

for more details._

---

### DIFF
--- a/task/argocd-task-sync-and-wait/0.2/argocd-task-sync-and-wait.yaml
+++ b/task/argocd-task-sync-and-wait/0.2/argocd-task-sync-and-wait.yaml
@@ -20,13 +20,17 @@ spec:
   params:
     - name: application-name
       description: name of the application to sync
+      type: string
     - name: revision
       description: the revision to sync to
       default: HEAD
+      type: string
     - name: flags
       default: --
+      type: string
     - name: argocd-version
       default: v2.2.2
+      type: string
   stepTemplate:
     envFrom:
       - configMapRef:


vinamra28 (commented):
@jdurzi we can avoid bumping the version since there are no breaking changes. Can you please update the previous one instead?

vdemeester:
@jdurzi 👋🏼 thanks for the PR,

> Added a type to the params to prevent the Task breaking in OpenShift Pipelines (this requires an explicit type)

What version of OpenShift Pipelines are you using ? And are the webhooks running in your instance ? 
Asking this because there is nothing specific to OpenShift Pipelines on `params` handling. The only reason this (no types) would fail (on OpenShift Pipelines or any tekton pipeline distribution) is if the pipeline webhook is not running (as it's the one which sets defaults).

Anyway, I think it's better to be explicit with the `params` types, so 👍🏼 on this. I agree with @vinamra28, I don't think we need to bump the version.

jdurzi:
> @jdurzi 👋🏼 thanks for the PR,
> 
> > Added a type to the params to prevent the Task breaking in OpenShift Pipelines (this requires an explicit type)
> 
> What version of OpenShift Pipelines are you using ? And are the webhooks running in your instance ? Asking this because there is nothing specific to OpenShift Pipelines on `params` handling. The only reason this (no types) would fail (on OpenShift Pipelines or any tekton pipeline distribution) is if the pipeline webhook is not running (as it's the one which sets defaults).
> 
> Anyway, I think it's better to be explicit with the `params` types, so 👍🏼 on this. I agree with @vinamra28, I don't think we need to bump the version.

Weird. We upgraded the version of the operator on Friday, and everything seems to be running fine, including the pipelines webhook. Now all of our Pipelines fail unless an explicit type is given, and this affects tkn hub tasks as well, of course. We'll raise a support ticket!

vinamra28 (approved):

vdemeester (approved):
/lgtm

PR #969, opened 2022-04-17
Update buildah to v1.23.3

jimmyjones2:
buildah v1.18.0 is pretty old - update to the latest v1.23.3. This fixes a few security advisories in buildah and also from its dependencies (I'm not sure if they affect buildah when used in Tekton).

### DIFF
--- a/task/buildah/0.3/README.md
+++ b/task/buildah/0.3/README.md
@@ -17,7 +17,7 @@ kubectl apply -f https://raw.githubusercontent.com/tektoncd/catalog/main/task/bu
 
 * **IMAGE**: The name (reference) of the image to build.
 * **BUILDER_IMAGE:**: The name of the image containing the Buildah tool. See
-  note below.  (_default:_ quay.io/buildah/stable:v1.18.0)
+  note below.  (_default:_ quay.io/buildah/stable:v1.23.3)
 * **DOCKERFILE**: The path to the `Dockerfile` to execute (_default:_
   `./Dockerfile`)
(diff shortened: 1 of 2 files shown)


vinamra28 (approved):
/cc @vdemeester

vdemeester:
I wonder if we should bump the task version though 🤔

vinamra28:
> I wonder if we should bump the task version though 🤔

@vdemeester I don't have any strong opinions. Are there breaking changes done in buildah 1.23.3 which were there in 1.18.x?

vdemeester:
> > I wonder if we should bump the task version though thinking
> 
> @vdemeester I don't have any strong opinions. Are there breaking changes done in buildah 1.23.3 which were there in 1.18.x?

There might be ? this is where it gets a little fuzzy 😅.

vinamra28:
> > > I wonder if we should bump the task version though thinking
> > 
> > 
> > @vdemeester I don't have any strong opinions. Are there breaking changes done in buildah 1.23.3 which were there in 1.18.x?
> 
> There might be ? this is where it gets a little fuzzy 😅.

in that case let's plan to bump. It's better to bump then fixing the breaks in future 😅

vinamra28:
@jimmyjones2 can you please bump the version of buildah 😅  ?

jimmyjones2:
@vinamra28 Sure. I'd also like to make a change to store `--digestfile` in an EmptyDir to allow the source workspace to be a ConfigMap. Should I do both in the same commit as both are a (slight) change in behavior?

jimmyjones2:
Replaced with #983

PR #1000, opened 2022-06-22
Adding pre_script parameter.

peterbosalliandercom:
# Changes

We need to install binaries before the go test can be run. In our case we need kubebuilder before we can run go tests in conjuction with operator-sdk which needs kubernetes client, etcd etc. 

# Submitter Checklist

review them:

          ```

          spec:
            description: >-

          ```

for more details._

---

### DIFF
--- a/task/golang-test/0.3/README.md
+++ b/task/golang-test/0.3/README.md
@@ -1,0 +1,58 @@
+# Golang Test
+
+This task is a Golang task to test Go projects.
+
+## Install the task
+
+```
+kubectl apply -f https://raw.githubusercontent.com/tektoncd/catalog/main/task/golang-test/0.2/golang-test.yaml
+```
+
+### Parameters
+
+* **package**: base package to build in
+* **packages**: packages to test (_default:_ ./cmd/...)
+* **context**: path to the directory to use as context (default: .)
+* **version**: golang version to use for builds (_default:_ latest)
+* **flags**: flags to use for `go test` command (_default:_ -race -cover -v)
+* **GOOS**: operating system target (_default:_ linux)
+* **GOARCH**: architecture target (_default:_ amd64)
+* **GO111MODULE**: value of module support (_default:_ auto)
+* **GOCACHE**: value for go caching path (_default:_ "")
+* **GOMODCACHE**: value for go module caching path (_default:_ "")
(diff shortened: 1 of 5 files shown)


chmouel:
usually in tekton pipelines world you would have another task before this golang task with a shared workspace between the two that does that and do "pre" stuff? this is not working for you ?

peterbosalliandercom:
> olang task

Before we run the golang task we need kubebuilder for that. With the pre-script way we can install and configure the kubebuilder. I expect that more users will need this pre-scripting in order to bootstrap the environment before they run the actual go test. It would be great if instead of the pre-script it would be possible to have a sort of boolean that includes the kubebuilder automatically but i think that is not really easy to do. Do you have an idea how to run these scripts using existing tekton taks? As an example we are using this pre-script right now. Do you have any other way to do this?
` 
 - name: PRE_SCRIPT
      description: Running pre_script before main script is executed
      type: string
      default: |
        export K8S_VERSION=1.21.2
        curl -sSLo envtest-bins.tar.gz "https://go.kubebuilder.io/test-tools/${K8S_VERSION}/$(go env GOOS)/$(go env GOARCH)"
        mkdir /usr/local/kubebuilder
        tar -C /usr/local/kubebuilder --strip-components=1 -zvxf envtest-bins.tar.gz`

chmouel:
@peterbosalliandercom you would have another task just before the golang task that does in its script the shell snippet you have in there and this would be add subsequently to the golang task (assuming you have added runAfter: this pre task)

peterbosalliandercom:
> @peterbosalliandercom you would have another task just before the golang task that does in its script the shell snippet you have in there and this would be add subsequently to the golang task (assuming you have added runAfter: this pre task)

Is there a task available in the catalog for that? If so that would be great.

chmouel:
> Is there a task available in the catalog for that? If so that would be great.

that would just be inline task using a shell image  with the script directive 

ie:

```yaml
      - name: pretask
        runAfter:
          - git-clone-task
        workspaces:
          - name: source
            workspace: source
        taskSpec:
          workspaces:
            - name: source
          steps:
            - name: noop-task
              image: registry.access.redhat.com/ubi9/ubi-micro #or alpine or others
              workingDir: $(workspaces.source.path)
              script: |
				# Your script that would do a kubebuilder
                exit 0
   - name: golang-test
     taskRef: golang
     etc.... 
```

i think that should work

PR #1010, opened 2022-07-04
Add fully flexible script to yq task

frerikandriessen:
# Changes

The yq task was missing a -P (pretty print) option, as well as a --no-color option. Instead of making new task versions (and params) for each of the myriad of options yq offers, I figured that what I want in a `yq` task is to be able to use yq, however I want it. A script is more suitable for that.

The only problem I have with this change is that it's quite different from the previous versions. I was thinking about splitting it up into a new task, but then the new task's name would not have been `yq`, or the old task would have to be renamed to `yq-replace`, and then we would still be stuck with a big change for every user.

In theory, if you update your task you would take into account you might have to update your pipelines to make them work with the new task version.
Let me know if you think this will be problematic.

# Submitter Checklist

review them:

          ```

          spec:
            description: >-

          ```

for more details._

---

### DIFF
--- a/task/yq/0.4/yq.yaml
+++ b/task/yq/0.4/yq.yaml
@@ -1,0 +1,57 @@
+apiVersion: tekton.dev/v1beta1
+kind: Task
+metadata:
+  name: yq
+  labels:
+    app.kubernetes.io/version: "0.4"
+  annotations:
+    tekton.dev/pipelines.minVersion: "0.12.1"
+    tekton.dev/categories: Developer Tools
+    tekton.dev/tags: yq
+    tekton.dev/displayName: "YQ"
+    tekton.dev/platforms: "linux/amd64"
+spec:
+  description: >-
+    This task can be used to replace fields in YAML files. For example for altering helm charts on GitOps repos.
+  workspaces:
+    - name: source
+      description: A workspace that contains the file which needs to be altered.
+  params:
+    - name: SCRIPT
+      type: string
+      description: The yq script to execute. Can be multiple lines for complex tasks.
+      default: ""
+    - name: image
+      type: string
+      description: The yq image to use.
+      default: docker.io/mikefarah/yq:4.27.5@sha256:2be3626ed633fbe1fc33ee9343a1256a6be53334412b2251b9a859f8c145bb53
+    - name: files
+      type: array
+      description: (deprecated, use SCRIPT instead) A list of files to execute the expression on. Needs to be relative to the source workspace.
+      default: []
+    - name: expression
+      type: string
+      description: (deprecated, use SCRIPT instead) The yq expression to apply. Can be used to replace yaml fields.
+      default: ""
+  results:
+    - name: yq
+      description: The result from your yq command. You can write to it using `$(results.yq.path)`
+  steps:
+    - name: yq-script
+      image: $(params.image)
+      workingDir: $(workspaces.source.path)
+      args: ["$(params.files[*])"]
+      script: |
+        /usr/bin/env sh
+        set -e
+
+        # For backwards compatibility with previous versions
+        if [ "$(params.SCRIPT)" = "" ]; then
+          for var in "$@"
+          do
+              /usr/bin/yq eval -i "$(params.expression)" "$var"
+          done
+          exit $?
+        fi
+
+        $(params.SCRIPT)
--- a/task/yq/0.4/README.md
+++ b/task/yq/0.4/README.md
@@ -1,0 +1,28 @@
+# YQ Task
+
+The following task is used to interface with the yq tool on any file in your workspace.
+
+## Installing the Task
+
+```bash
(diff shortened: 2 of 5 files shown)


task/yq/0.4/yq.yaml
```diff
@@ -0,0 +1,35 @@
...
+spec:
+  description: >-
+    This task can be used to replace fields in YAML files. For example for altering helm charts on GitOps repos.
+  workspaces:
+    - name: source
+      description: A workspace that contains the file which needs to be altered.
+  params:
+    - name: SCRIPT
+      type: string
+      description: The yq script to execute. Can be multiple lines for complex tasks.
```
vinamra28, line 22: @frerikandriessen I can see that a few existing params have been removed and in replace we have a new param `SCRIPT`. This can be a huge breaking change for existing users whenever they try to bump the version. How about we keep those params along with this new param `SCRIPT`? We can have a default value of `SCRIPT` param  to contain the script which was present earlier. This will not break existing user's experience. WDYT?

frerikandriessen: @vinamra28

vinamra28: @frerikandriessen tekton doesn't has this as of now. Probably in the param description we can say it as deprecated.

frerikandriessen: @vinamra28 because I needed to get the params in a backwards compatible way I've solved it like this, is that also what you intended? Let me know

vinamra28: @frerikandriessen yes. I hope this workaround should work 😅

task/yq/0.4/yq.yaml
```diff
@@ -0,0 +1,51 @@
...
+      description: The yq image to use.
+      default: docker.io/mikefarah/yq:4.16.2@sha256:0d4f6e27bdcac7316f635acd524ab0eecc4ad50834b54d10322268650c7712cb
+    - name: files
+      type: array
+      description: (deprecated, use SCRIPT instead) A list of files to execute the expression on. Needs to be relative to the source workspace.
+      default: []
+    - name: expression
+      type: string
+      description: (deprecated, use SCRIPT instead) The yq expression to apply. Can be used to replace yaml fields.
+      default: ""
```
vinamra28, line 34: @vdemeester @piyush-garg how does this idea of deprecating the params look like? Removing them directly will break the users. WDYT?

vinamra28 (commented):
/lgtm

frerikandriessen:
@vdemeester @piyush-garg could you take a look?

task/yq/0.4/yq.yaml
vdemeester: If it's a new version, we can definitely "break" the user. As commented elsewhere, we do not make any "backward compatible" guarantees from a version to another in the catalog. So from my point of view, removing it directly works for me.

frerikandriessen: Should I change that? Or is it also fine as is?

frerikandriessen:
@vinamra28 could you ping someone who is available to approve (or reject) this?

vinamra28:
> Catlin script lint Output
> ```
> ERROR: /usr/bin/shellcheck, [-s sh] failed:
> 
> In yq-yq-script line 6:
>   $(params.SCRIPT)
>   ^--------------^ SC2091: Remove surrounding $() to avoid executing output (or use eval if intentional).
> 
> For more information:
>   https://www.shellcheck.net/wiki/SC2091 -- Remove surrounding $() to avoid e...
> ```

@frerikandriessen, can you please try to resolve the above?

frerikandriessen:
> > Catlin script lint Output
> > ```
> > ERROR: /usr/bin/shellcheck, [-s sh] failed:
> > 
> > In yq-yq-script line 6:
> >   $(params.SCRIPT)
> >   ^--------------^ SC2091: Remove surrounding $() to avoid executing output (or use eval if intentional).
> > 
> > For more information:
> >   https://www.shellcheck.net/wiki/SC2091 -- Remove surrounding $() to avoid e...
> > ```
> 
> @frerikandriessen, can you please try to resolve the above? It seems that catlin now wants us to make it happy 😅

@vinamra28 wouldn't it be better to disable this false positive from the checker? Doing `eval '$(params.SCRIPT)'` feels double-wrong, because we do not want to execute/evaluate code, we just want to source whatever's stored in that variable there in the script. Making it 'explicit' that we want to execute/evaluate code gives (I think) the wrong idea.

frerikandriessen:
@vinamra28 I added 2 more things:
1. is a bugfix in which the task was unable to write to file due to permission errors, but the task still 'passed'
2. is a way for you to use yq to read out a specific value and store it in the task results

Let me know what you think

vinamra28 (commented):
/lgtm

chmouel:
/approve

thanks 🙇🏻

PR #1022, opened 2022-07-19
feat(script): add new task - version 0.1

gerrnot:
# Changes

feat(script): add new task - version 0.1

NOTE: This was intially called bash, now i generified it to allow any script (the branch name should be the only place where it is still called bash)

### DIFF
--- a/task/script/0.1/script.yaml
+++ b/task/script/0.1/script.yaml
@@ -1,0 +1,104 @@
+apiVersion: tekton.dev/v1beta1
+kind: Task
+metadata:
+  labels:
+    app.kubernetes.io/version: "0.1"
+  annotations:
+    tekton.dev/pipelines.minVersion: "0.17.0"
+    tekton.dev/categories: "CLI"
+    tekton.dev/displayName: "script"
+    tekton.dev/tags: "bash,shell,script,exec"
+    tekton.dev/platforms: "linux/amd64,linux/arm32v5,linux/arm32v7,linux/arm64v8,linux/i386,linux/mips64le,linux/ppc64le,linux/riscv64,linux/s390x"
+  name: script
+spec:
+  description: >-
+    Executes a script(e.g bash, python etc) and exposes stderr, stdout and exit_code as results.
+    In the default image bash version 5 is installed
+  params:
+    - name: "script_source"
+      description: >-
+        script source code (multiline)-string.
+        Will be copied to a file and then executed via '/var/myScript', so please specify a shebang line, e.g. for bash:
+        `#!/usr/bin/env bash`
+    - name: "image"
+      description: >-
+        Image that this task runs on.
+        Must contain bash since bash is used to start the provided script,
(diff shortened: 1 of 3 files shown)


task/script/0.1/tests/test_local.sh
```diff
@@ -0,0 +1,52 @@
+#!/usr/bin/env bash
+# only used for local quick testing, since this project has its own test run strategy described in CONTRIBUTING.md
+# USAGE: cd into directory of this script, then execute: ./run_local.sh
+
+set -ex
+
```
vinamra28, line 6: this seems to be good. Can you please move this outside of tests, maybe in a different directory named as `support`. Also you can mention about this script file in README.md so that if somebody wants to use this then they can use this. Btw to test tasks locally we already have a script in tests directory which you can trigger by running
```sh
bash test/run-test.sh <task-name> <version>
```

task/script/0.1/script.yaml
```diff
@@ -0,0 +1,108 @@
...
+        failure and then the result will not be accessible, because a failed tekton task has no results atm!
+  workspaces:
+    - name: "workspace-a"
+      optional: true
+    - name: "workspace-b"
+      optional: true
+    - name: "workspace-c"
+      optional: true
+    - name: "workspace-d"
+      optional: true
```
vinamra28, line 108: instead of 4 workspaces can we have only 3? Reason: affinity-assitant runs by default and it won't allow Pipeline to bind more than 2 PVC's simultaneously in a Task. Probably 3 workspaces can be such that one can be used to bind pvc, one for configmap and one for secrets and all 3 can be optional. WDYT?

vinamra28 (commented):
/ok-to-test

task/script/0.1/script.yaml
gerrnot: Ok, I reduced it to three in the follow-up push.

In the long run, it would be good if tekton supports more namespaces, but luckily three is actually what we use productively atm!

task/script/0.1/tests/test_local.sh
gerrnot: Thanks! I move the script and documented both scripts.

PS: In between I thought about merging them together, but ran into quite some troubles and gave up on it (actually I deleted a namespace that I did not intent to), so I guess it is least safer this way.

bendory:
Can you split this into separate PRs -- it would be easier to review if each PR focuses on one change or set of changes. I see here several issues that need not be reviewed together:

- a new task
- testing infrastructure, which is a new feature
- documentation changes -- some of which are for the testing infrastructure

gerrnot:
@bendory Ok, i refactored everything regarding the test utility out to a separate pr: https://github.com/tektoncd/catalog/pull/1041

bendory:
> @bendory Ok, i refactored everything regarding the test utility out to a separate pr: #1041

Thanks! That makes both PRs cleaner and clearer!

gerrnot:
unfortunately i have to withdraw this due to possible legal issue

PR #1030, opened 2022-07-28
v0.2 for orka-teardown task

jkomoda:
# Changes

Added version `0.2` for `orka-teardown` task

* Compatibility changes to make the pipelines also work properly when using Tekton v0.34.1
* Updated `orka-tekton-runner` docker image and updated versioning.

# Notes
* Requires `task/orka-init/0.2/orka-init.yaml` in `pre-apply-task-hook.sh` in order for tests to work.

# Submitter Checklist

review them:

          ```

          spec:
            description: >-

          ```

for more details._

---

### DIFF
--- a/task/orka-teardown/0.2/orka-teardown.yaml
+++ b/task/orka-teardown/0.2/orka-teardown.yaml
@@ -1,0 +1,73 @@
+---
+apiVersion: tekton.dev/v1beta1
+kind: Task
+metadata:
+  name: orka-teardown
+  labels:
+    app.kubernetes.io/version: "0.2"
+  annotations:
+    tekton.dev/categories: Cloud, Kubernetes
+    tekton.dev/pipelines.minVersion: "0.16.0"
+    tekton.dev/tags: "orka, macstadium, teardown, cleanup"
+    tekton.dev/displayName: "orka teardown"
+    tekton.dev/platforms: "linux/amd64"
+spec:
+  description: >-
+    With this set of Tasks, you can use your Orka environment
+    to run macOS builds and macOS-related testing from your Tekton pipelines.
+
+    This Task cleans up the Orka environment after your workload completes.
+    This Task deletes the VM instances deployed with `orka-deploy` and their related templates.
+  params:
+    - name: orka-token-secret
+      type: string
+      description: |
+        The name of the secret holding the authentication token used to access the Orka API.
+      default: orka-token
+    - name: orka-token-secret-key
+      type: string
+      description: |
+        The name of the key in the Orka token secret, which holds the authentication token.
+      default: token
+    - name: orka-vm-name-config
+      type: string
+      description: |
+        The name of the config map, which stores the name of the generated VM configuration.
+      default: orka-vm-name
+    - name: orka-vm-name-config-key
+      type: string
+      description: |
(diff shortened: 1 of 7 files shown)


task/orka-teardown/0.2/tests/resources.yaml
```diff
@@ -0,0 +1,49 @@
...
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: orka-svc
+  namespace: orka-teardown-0-2
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: orka-runner-teardown
```
vinamra28, line 28: this ClusterRole can probably clash with version 0.1's ClusterRole, hence we'll have to update the name here 😅 
```suggestion
  name: orka-runner-teardown-0-2
```

task/orka-teardown/0.2/tests/resources.yaml
```diff
@@ -0,0 +1,49 @@
...
+      - configmaps
+      - secrets
+    verbs:
+      - create
+      - delete
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: orka-runner-teardown
```
vinamra28, line 41: same as above

vinamra28 (changes requested):

task/orka-teardown/0.2/orka-teardown.yaml
```diff
@@ -0,0 +1,73 @@
...
+      type: string
+      description: |
+        The name of the config map, which stores the name of the generated VM configuration.
+      default: orka-vm-name
+    - name: orka-vm-name-config-key
+      type: string
+      description: |
+        The name of the key in the VM name config map, which stores the name of the generated VM configuration.
+      default: vm-name
+    - name: docker-image
```
vinamra28, line 42: docker-image seems to be a generic name, previous name `orka-image` seems to be better 😅

spikeburton: @vinamra28 `orka-image` is actually a poor name for the input. For folks familiar with Orka, an "image" typically refers to the macOS virtual machine base image and this can/will cause confusion.

I don't see what's wrong with `docker-image`, it refers to the container image used to run the task 🙂 but if you prefer, maybe `orka-tekton-runner-image` or something like that would be better.

As an aside, I am not sure why this parameter is even needed at all, IIRC the image used for task steps should not be parameterized. Can you remind us why it is needed?

vinamra28: @spikeburton agreed `orka-image` seems to be a poor name. `orka-tekton-runner-image` sounds good 👍. The reason why I prefer `orka` keyword in the parameter name is - within a `Pipeline` there can be multiple such parameters where user might have to pass their own image and there can be a time when they might get confused while passing the params. 

> As an aside, I am not sure why this parameter is even needed at all, IIRC the image used for task steps should not be parameterized. Can you remind us why it is needed?

There can be users who are running Tekton pipelines in air-gapped environment and doesn't have access to public image registry so in that case they might have to pass their own images. Hope this answers the query 🙂

spikeburton: Great, thanks for the clarification 🙂

PR #1089, opened 2022-10-10
double quote to prevent globbing and word splitting

dlbewley:
fix shellcheck error SC2086

# Changes

Fixes error from catalog build job by double quoting variable used in shell script.


vinamra28 (approved):
/lgtm

PR #1090, opened 2022-10-11
Switch pylint image to a more updated one

chmouel:
cytopia pylint image hasn't updated since 7 months ago and multiple
release.

pipeline-components looks a legit org
https://gitlab.com/pipeline-components/pylint

image is alpine based
https://gitlab.com/pipeline-components/pylint/-/blob/master/Dockerfile

### DIFF
--- a/task/pylint/0.2/pylint.yaml
+++ b/task/pylint/0.2/pylint.yaml
@@ -20,7 +20,7 @@ spec:
   params:
     - name: image
       description: The container image with pylint
-      default: docker.io/cytopia/pylint@sha256:6e5b49b6d54cbd845b93139eeb3f4f558b07bf6a87001ce254782463df1443d2
+      default: registry.gitlab.com/pipeline-components/pylint:edge
     - name: path
       description: The path to the module which should be analysed by pylint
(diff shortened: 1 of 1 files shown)


task/pylint/0.2/pylint.yaml
```diff
@@ -20,7 +20,7 @@ spec:
   params:
     - name: image
       description: The container image with pylint
-      default: docker.io/cytopia/pylint@sha256:6e5b49b6d54cbd845b93139eeb3f4f558b07bf6a87001ce254782463df1443d2
+      default: registry.gitlab.com/pipeline-components/pylint:edge
```
vinamra28, line 23: should we tag the image with digest as well?
/approve
/test .*

chmouel: Yeah but same as before ,how could we keep this updated ? We end up with very old version of tools in catalog

vdemeester: We need some *dependabot-like* automation here.. something that would watch images and submit a PR on new tag published.

PR #1094, opened 2022-10-21
Make directory safe only right before the git-init

chmouel:
We were doing the git safe add-directory first but when having a
WORKSPACE_BASIC_AUTH_DIRECTORY_BOUND we would erase the .gitconfig and
that safe variable would not be there anymore.

Do this only right before calling the git-init so we know the setting
would be applied.

cc @jhonis


vinamra28 (approved):

vdemeester (approved):
/lgtm

PR #1378, opened 2026-06-08
ci: add CI summary fan-out job

vdemeester:
# Changes

Add a single **CI summary** job that depends on all CI jobs (`linting` and the `e2e-tests` matrix) and surfaces an aggregate pass/fail status, matching the pattern used in `tektoncd/pipeline` and other tektoncd repos.

## Motivation

This provides a single required status check that fans out over the matrix legs. Branch protection (and Prow/tide) only needs to require **CI summary** rather than enumerating every matrix combination — which change over time as pipeline versions are bumped (e.g. `e2e tests (latest)` with `v1.9.0` → `v1.13.0`).

The job runs with `if: always()` and fails if any required job's result is not `success` or `skipped`.

/kind cleanup

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -181,3 +181,35 @@ jobs:
             echo '::endgroup::'
           done
         fi
+
+  ci-summary:
+    name: CI summary
+    needs: [linting, e2e-tests]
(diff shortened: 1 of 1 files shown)


vinamra28 (approved):
/lgtm

